PR #44, opened 2026-04-24
TAKO-2600: tool registry convention + auto-gen registry/server.json codegen

jfassio:
Phase 1.5 of the [Tako MCP](https://linear.app/trytako/project/tako-mcp-635ac07fae22) Cloudflare Workers port. Closes [TAKO-2600](https://linear.app/trytako/issue/TAKO-2600).

## Summary

- Every tool now ships as one file under `workers/src/tools/` that default-exports a `ToolModule`
- A codegen walks that directory and emits **both** `registry/server.json` (external MCP registry card) and `workers/src/tools/_registry.ts` (static barrel the Worker auto-registers from). Same scan → no possible drift between what the registry advertises and what the Worker actually exposes
- CI fails on drift: `npm run registry:check` regenerates in-memory and exits non-zero on any diff
- Closes the Phase 1 TODO in `mcp.ts:69` — `/mcp` is now gated by `extractBearer` before the SDK sees the request; missing / malformed / empty Authorization headers return a uniform 401 with the Phase 1 JSON-RPC bearer error shape (code `-32001`, `data.kind`)

## Files

- **`workers/src/tools/types.ts`** — `ToolModule<InputSchema, Output>` (typed, used with `satisfies` at tool files) + `AnyToolModule` (type-erased, used at the registry boundary) + `ToolContext { token, env }`
- **`workers/src/tools/_example.ts`** — stub tool proving the contract end-to-end. The first Phase 2 port ([TAKO-2602](https://linear.app/trytako/issue/TAKO-2602)) should delete this file
- **`workers/src/tools/_registry.ts`** — codegen-emitted static barrel (regenerated on every `registry:gen`)
- **`workers/scripts/gen-registry.ts`** — codegen with `write` and `--check` modes. Uses zod 4's native `z.toJSONSchema`, flattens to the registry's per-parameter `{type, description, required, default}` shape, deterministic output (preserved metadata key order, 2-space indent, LF, trailing newline)
- **`registry/metadata.json`** — static fields (authors, categories, links, installation, transport, authentication) that aren't derivable from tool code. Preserved verbatim from the current `registry/server.json`
- **`registry/server.json`** — now generated. See "Expected diffs" below
- **`workers/src/mcp.ts`** — `createMcpServer(ctx)` loops `TOOL_REGISTRY` and calls `server.registerTool(...)` per tool. Handler wrapper emits `structuredContent` alongside a JSON-stringified `content[0].text` fallback when the tool declares an `outputSchema`. Bearer auth gate added at the top of `handleMcpRequest`
- **`workers/package.json`** — adds `registry:gen`, `registry:check`, `predeploy:staging`, `predeploy:production` scripts; adds `zod` runtime dep; adds `tsx` devDep
- **`.github/workflows/workers-ci.yml`** — adds `Registry check` step; extends path filter to include `registry/**`

## Expected registry/server.json diffs (from pre-existing hand-written file)

1. **`api_token` disappears from every tool's `parameters`.** TS port uses connection-level Bearer auth (header extracted once, forwarded as `X-API-Key` by `django.ts`); tools no longer take a per-call token. This is correct behavior for the TS Worker.
2. **The 8 Python tools are gone from the registry** until they're ported in Phase 2. The file currently lists only `_example`. The Python pip distribution still carries the old tool metadata via its own independent `_SERVER_CARD` constant in `src/tako_mcp/server.py` — this PR does not touch that file.

Per our [pre-investigation discussion](https://linear.app/trytako/issue/TAKO-2600), this diff is intentional and does not have live consumers to worry about (Python server isn't currently operational).

## Test plan

- [x] `npm test` — **39/39 passing** (was 38/38 — +1 tools/call end-to-end test that proves the `_example` handler pipeline, including Zod input validation and `structuredContent` shaping, works via a real MCP `tools/call` roundtrip)
- [x] `npm run registry:check` — passes with clean working tree
- [x] `npm run registry:check` — fails with exit 1 on injected drift, prints the offending file path and `Run \`npm run registry:gen\`` hint
- [x] `npm run registry:gen` — regenerates both files; running again is a no-op (deterministic)
- [x] `npx wrangler deploy --dry-run --env staging` — validates
- [x] `npx wrangler deploy --dry-run --env production` — validates
- [ ] CI runs end-to-end on this PR

## What's next

Unblocks the 8 Phase 2 tickets ([TAKO-2602](https://linear.app/trytako/issue/TAKO-2602)..[TAKO-2609](https://linear.app/trytako/issue/TAKO-2609)). Each one: write one tool file under `workers/src/tools/`, run `npm run registry:gen`, commit both the tool file and the regenerated `registry/server.json` + `_registry.ts`. No manual registry editing.


registry/metadata.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "docker": {
+      "image": "ghcr.io/takodata/tako-mcp:latest",
+      "ports": ["8001:8001"],
+      "environment": {
+        "PUBLIC_BASE_URL": "https://tako.com"
+      }
+    }
+  },
+  "transport": {
```
robertabbott, line 51: `runtime` is `"python"` and `package.type` is `"pip"`, but `registry/server.json` is now generated from the TS Worker's tool set. Since the Python server carries its own independent `_SERVER_CARD`, the external registry would be advertising pip installation for a server whose tools come from the Cloudflare Worker. Should these fields be updated to reflect the TS Worker now, or is the intent to update them when the Python server is fully decommissioned?

workers/package.json
```diff
@@ -11,17 +11,23 @@
     "dev": "wrangler dev",
     "dev:staging": "wrangler dev --env staging",
     "deploy:staging": "wrangler deploy --env staging",
+    "predeploy:staging": "npm run registry:gen",
```
robertabbott: `predeploy:staging` runs `registry:gen` (write mode). A local `wrangler deploy --env staging` would silently regenerate the registry files and deploy whatever the current tool source produces — potentially different from what was committed and reviewed in the PR. Would `registry:check` be a safer choice here, so a manual deploy fails loudly if the committed files are stale?

workers/src/mcp.ts
```diff
@@ -41,6 +50,71 @@ export function createMcpServer(): McpServer {
...
+      .shape;
+    if (outputShape !== undefined) {
+      config.outputSchema = outputShape;
+    }
+  }
+
+  server.registerTool(
+    tool.name,
+    config as Parameters<McpServer["registerTool"]>[1],
+    async (input) => {
```
robertabbott, line 107: `tool.handler(input, ctx)` is awaited without a try/catch. When Phase 2 tools call Django and encounter `DjangoUnauthorizedError` or `DjangoTimeoutError`, the MCP SDK catches the raw throw and surfaces `err.message` as a generic tool error. Is there a plan for a catch at the `registerTool` boundary that maps `DjangoError` subtypes to structured `isError: true` results (e.g., distinguishing "your token was rejected by Django" from "the upstream timed out"), or will each tool be expected to handle its own errors?

robertabbott (commented):
**`workers/tsconfig.json` — `scripts/` excluded from type-checking**

`include` covers `src/**/*.ts` but not `scripts/`. `npm run typecheck` won't catch type errors in `gen-registry.ts` — a type-level regression in the codegen would only surface at `tsx` runtime, not in CI's typecheck step. Should `"scripts/**/*.ts"` be added to `include`?

workers/package.json
jfassio: Good catch — flipped in 1195c51. `predeploy:staging` and `predeploy:production` now run `registry:check` instead of `registry:gen`, so a local `npm run deploy:*` fails loudly on committed-registry drift instead of silently regenerating. CI's `registry:check` step is unchanged.

registry/metadata.json
jfassio: Intentional — deferring. The pip package is still published and carries its own independent `_SERVER_CARD`, so the external registry advertising pip installation is still an accurate published install path today even though `registry/server.json` is now generated from the TS Worker's tool set (which is the thing that's actually changing). The `runtime` / `package` / `installation.pip` / `transport` fields all flip together when Python is formally decommissioned — doing it piecemeal would leave the registry describing a server that doesn't exist on either path. I'll open a followup to retire these in one pass once the Worker is fully live.

jfassio:
Re: `workers/tsconfig.json` — `scripts/` excluded from type-checking

Fixed in 1195c51. Added `workers/scripts/tsconfig.json` that extends the base and adds Node types via `@types/node`, plus threaded a second `tsc --noEmit -p scripts/tsconfig.json` into the `typecheck` script. Kept this as a separate tsconfig rather than merging into the base so worker source stays isolated from Node globals — a stray `process` / `fs` reference leaking into `src/` still fails the base typecheck.

Verified manually: injected `const __probe: number = "s"` into `gen-registry.ts`, `npm run typecheck` now exits 2 with `scripts/gen-registry.ts(2,7): error TS2322`.

robertabbott (approved):
LGTM. All three actionable comments addressed:

- `predeploy:*` flipped to `registry:check` — deploys now fail loud on drift
- Django error mapping lands via `djangoErrorToToolResult` with discriminated `kind`, plus unit coverage for each subtype
- `scripts/tsconfig.json` gets its own typecheck pass with Node types, base stays clean of Node globals

Metadata runtime/package deferral is reasonable — the pip package is still a live install path and flipping those fields piecemeal would leave the registry describing a server that doesn't exist on either path.

PR #45, opened 2026-04-24
feat(workers): port Phase 2 tools + deferred candidates

jfassio:
## Summary

Ports the Python MCP tool surface from `src/tako_mcp/server.py` to TypeScript `ToolModule`s following the TAKO-2600 contract, and adds 4 deferred candidates the TAKO-2599 audit ([PR #43](https://github.com/TakoData/tako-mcp/pull/43)) surfaced. **9 tools total** after dropping three deprecated endpoints (see _Endpoint cleanup_ below).

**Stacked on** [PR #44 (TAKO-2600)](https://github.com/TakoData/tako-mcp/pull/44). Base is `jfassio/tako-2600-tool-registry-convention-codegen` so the diff only shows the Phase 2 delta. Rebase onto `main` once #44 lands.

## Tools

**Phase 2 ports (TAKO-2602..2609):**
| Tool | Method | Path |
|---|---|---|
| `knowledge_search` | POST | `/api/v1/knowledge_search` |
| `get_chart_image` | — | (URL constructor, no fetch) |
| `explore_knowledge_graph` | POST | `/api/v1/explore/` |
| `create_chart` | POST | `/api/v1/thin_viz/create/` |
| `open_chart_ui` | — | (embed URL + iframe HTML builder) |

**Deferred candidates (audit-proposed, now in scope):**
| Tool | Method | Path |
|---|---|---|
| `list_reports` | GET | `/api/v1/internal/reports/` |
| `create_report` | POST | `/api/v1/internal/reports/` |
| `get_report` | GET | `/api/v1/internal/reports/{id}/` |
| `get_credit_balance` | GET | `/api/v1/credit_balance/` |

## Endpoint cleanup (post-review, per Bobby's feedback)

- **Removed `get_card_insights`** — `/api/v1/internal/chart-configs/{id}/chart-insights/` is no longer supported.
- **Removed `list_chart_schemas` + `get_chart_schema`** — schema discovery flow is gone; clients now supply component configurations directly.
- **Rewrote `create_chart`** to hit `POST /api/v1/thin_viz/create/` (the [Thin-Viz: Create Card](https://docs.tako.com/api-reference/thinviz-direct-create) endpoint) instead of the deprecated `POST /api/v1/thin_viz/default_schema/{name}/create/`. Input schema drops `schema_name` and accepts `components[]` with optional `title`, `description`, `source`.

## Deviations from the Python reference

- **`get_chart_image` skips the pre-ping.** URL is deterministic from `pub_id`; pinging doubles latency. Clients discover 404 on fetch.
- **`open_chart_ui` returns structured data instead of a raw MCP-UI resource.** The current `registerTool` adapter (`mcp.ts:98-115`) only emits text content, so raw `resource` content blocks would require a contract extension. Tool returns `{pub_id, embed_url, iframe_html, dark_mode, width, height}` — clients can render the iframe HTML inline or fall back to the embed URL. Re-adding full MCP-UI is a follow-up.
- **`api_token` is not a tool-call input.** It's lifted from the Bearer header by TAKO-2598's auth extractor and injected via `ToolContext.token`, so token values never appear in tool metadata.

## Verification

- `npm run typecheck` — clean.
- `npm run registry:check` — passes, no drift between `registry/server.json` and the runtime barrel (9 tools).
- `npm test` — 41/41 pass. `index.test.ts` integration tests updated to cover the 9-tool set and exercise the SDK round-trip via `open_chart_ui` (no fetch mocking needed).

## Test plan

- [x] Merge [#44](https://github.com/TakoData/tako-mcp/pull/44) (TAKO-2600) first, then rebase this onto `main`.
- [x] Review each tool module against the Python source in `src/tako_mcp/server.py` to catch any missed behavior.
- [x] Deploy to staging and smoke-test each tool from an MCP client against `mcp.staging.tako.com` with a real Tako Bearer token — in particular, confirm `create_chart` round-trips against `/api/v1/thin_viz/create/` with real component configs.
- [ ] Per-tool unit tests (each tool has just the integration coverage from `index.test.ts` today) — file follow-up ticket if per-tool coverage is wanted before GA.

Refs: TAKO-2599 (audit), TAKO-2600 (registry contract), TAKO-2602..2609 (per-tool tickets).


jfassio:
## Self-Review Loop Summary (3 iterations)

- **Iter 1** — Reviewer found **0 Critical, 3 Important, 5 Minor** issues.
  - Fixed: `DjangoBadRequestError.body` is now spliced into the thrown message by `create_chart` + `create_report` (so LLMs can read DRF validation guidance and retry); `get_report` surfaces `webpage_url` as a dedicated top-level output field; `open_chart_ui` header now documents the MCP-UI interactivity regression vs the Python `UIResource` behavior. `env.ts` gained an optional `PUBLIC_BASE_URL`; `get_chart_image` + `open_chart_ui` prefer it when set. `explore_knowledge_graph` caps `sample_members` at 10. `get_credit_balance` schema tightened to `.loose()` with a `credit_balance` hint.
  - Not fixed: `list_reports` status enum left closed (drift risk low).
- **Iter 2** — Reviewer found **0 Critical, 3 Important, 5 Minor** issues (evolved, not repeats).
  - Fixed: `get_credit_balance.credit_balance` is now `number | string` (DRF serializes `DecimalField` as a string by default — iter-1's `z.number()`-only schema would have broken on first call). Added a `resolvePublicBase(env)` helper in `env.ts` that enforces the same invariants `buildUrl` enforces for `DJANGO_BASE_URL` (non-empty, http/https scheme, no trailing slash) — `get_chart_image` and `open_chart_ui` call it instead of inlining fallback logic. New `workers/src/tools/create_report.test.ts` covers the `DjangoBadRequestError`-splice path (handler throws plain `Error`, message contains tool + status + body; non-400 errors propagate unchanged). Dropped the `?? ""` id fallback in `create_report` (throws) and `list_reports` (skips + warns). `embed_url` / `image_url` output validation tightened from `z.string().url()` to a regex requiring http(s). Renamed `EXPORT_URL_KEY_EXCLUSIONS` → `TOP_LEVEL_URL_KEYS` in `get_report` and dropped the dead `_urls` plural branch.
- **Iter 3** — Reviewer found **0 Critical, 0 Important, 3 Minor**. Assessment: **Ready to merge.**
  - Fixed anyway (small polish): `get_report` description now explicitly names `failed` as a terminal polling state; `get_credit_balance` uses `.safeParse` and throws a human-readable sentence on backend shape drift instead of a raw Zod dump; `open_chart_ui` header notes the Python `postMessage('tako::resize')` listener as a second deliberate omission.

### Verification
- `npm run typecheck`: clean
- `npm run registry:check`: passes (12 tools, no drift)
- `npm test`: **41/41 vitest tests pass** (added `create_report.test.ts` covering the error-forwarding contract)

### Known deferred items
- Per-tool Django-mocking unit tests for the remaining 11 tools (only `create_report` has a dedicated test file; the rest are covered by the SDK-wiring integration test via `open_chart_ui`). Filing as a follow-up ticket.
- Full MCP-UI interactive `UIResource` return in `open_chart_ui` — requires extending the `ToolModule` contract with a content-block emitter. Parked alongside `postMessage('tako::resize')` re-introduction.
- `list_report_types` MCP tool — blocked on a small backend addition to expose `get_all_report_types()` registry.

**Reviewer assessment: Ready to merge.**

workers/src/index.test.ts
```diff
@@ -118,14 +118,26 @@ describe("worker routing", () => {
...
     };
-    const names = body.result.tools.map((t) => t.name);
-    // Exact set: when Phase 2 ports land, `_example` will be gone and
-    // real tool names take its place. The assertion then updates in the
-    // same PR that drops the stub.
-    expect(names).toEqual(["_example"]);
+    const names = body.result.tools.map((t) => t.name).sort();
+    // The exact set of tools registered by the codegen barrel. When a tool
+    // is added or removed, update this list alongside the tool-module file.
+    expect(names).toEqual([
```
robertabbott, line 124: The PR body table lists 12 tools (8 Phase 2 ports + 4 deferred candidates), but the registry, barrel, and this test assertion all list 9. `get_card_insights`, `list_chart_schemas`, and `get_chart_schema` appear in the description table but have no implementation files, registry entries, or test coverage. The `create_chart` header comment says the schema-name flow is deprecated (explaining the schema tools), but what about `get_card_insights`? Were all three intentionally deferred, or should the PR description be updated to reflect the actual 9-tool scope?

workers/src/tools/get_chart_image.ts
```diff
@@ -0,0 +1,70 @@
...
+  async handler(input, ctx) {
+    const pubId = encodeURIComponent(input.pub_id);
+    const darkMode = input.dark_mode ? "true" : "false";
+    // `resolvePublicBase` prefers `PUBLIC_BASE_URL` when set (this is the
+    // URL the user's browser will load), falls back to `DJANGO_BASE_URL`,
+    // and enforces the same invariants `buildUrl` does for the Django
+    // origin (non-empty, http/https scheme, no trailing slash). Throws
+    // loud on config drift — this URL flows to user browsers, so it's a
+    // security boundary, not a soft fallback.
+    const base = resolvePublicBase(ctx.env);
```
robertabbott: The Python reference distinguishes `PUBLIC_API_URL` (for `/api/v1/image/` — defaults to `TAKO_API_URL`) from `PUBLIC_BASE_URL` (for `/embed/` — defaults to `https://tako.com`). The TS code uses `resolvePublicBase` for both `get_chart_image` and `open_chart_ui`, collapsing that distinction. If the production API origin and web origin differ (as the Python defaults suggest: `api.tako.com` vs `tako.com`), would `get_chart_image` produce an image URL on the wrong domain once `PUBLIC_BASE_URL` is wired, or does the deployment topology assume API and web share an origin?

workers/src/tools/create_report.ts
```diff
@@ -0,0 +1,130 @@
...
+        body,
+        { timeoutMs: 30_000 },
+      );
+    } catch (err) {
+      // The tool's input-schema description promises LLMs can retry after a
+      // 400 by reading the list of valid `report_type` values from the
+      // response body. `DjangoBadRequestError.message` is deliberately
+      // body-free (log-injection guard in `django.ts`), so we re-throw with
+      // the body spliced in so the SDK surfaces it as tool text content.
+      if (err instanceof DjangoBadRequestError) {
```
robertabbott: `create_chart` and `create_report` both catch `DjangoBadRequestError` and re-throw as plain `Error` with the body spliced in, so the validation detail reaches the LLM via `content[0].text`. The other 7 tools let the error propagate to `djangoErrorToToolResult`, which puts the body in `structuredContent.body` but keeps `content[0].text` body-free. Would it simplify things to include `.body` in the text content inside `djangoErrorToToolResult` itself (maybe gated on status 400), so individual tools don't need the catch-and-rethrow pattern? Or is the bifurcation intentional — e.g., only write tools should surface validation detail to the LLM?

workers/src/tools/create_chart.ts
```diff
@@ -0,0 +1,135 @@
...
+    const cardId = data.card_id ?? null;
+    const base = {
+      card_id: cardId,
+      title: data.title ?? null,
+      description: data.description ?? null,
+      webpage_url: data.webpage_url ?? null,
+      embed_url: data.embed_url ?? null,
+      image_url: data.image_url ?? null,
+    };
+    if (cardId !== null && cardId !== undefined && cardId !== "") {
```
robertabbott: `cardId` is assigned via `const cardId = data.card_id ?? null` (line 115), so its type is `string | null` — never `undefined`. The `cardId !== undefined` arm of the condition is unreachable. Same pattern at `knowledge_search.ts:113`. Should the condition simplify to `cardId !== null && cardId !== ""`?

workers/src/tools/open_chart_ui.ts
```diff
@@ -0,0 +1,156 @@
...
+    ></iframe>
+  </body>
+</html>`;
+
+    return {
+      pub_id: input.pub_id,
+      embed_url: embedUrl,
+      iframe_html: iframeHtml,
+      dark_mode: input.dark_mode,
+      width: input.width,
```
robertabbott, line 160: The output includes `width: input.width` (default 900), but the iframe HTML uses `width="100%"` unconditionally. A consumer receiving `{width: 900, height: 600}` might expect the iframe to be 900px wide. Should the HTML respect `input.width` (matching how `input.height` is applied), or should the output schema note that `width` is advisory for the consumer's own container?

workers/src/index.test.ts
jfassio: PR description updated

robertabbott (approved):

PR #48, opened 2026-04-24
fix(workers): create_report generation + template resolve, knowledge_search fast-first

jfassio:
## Summary

Two user-visible MCP bugs observed while generating reports and searching the knowledge base from staging:

- **`create_report` never started generation.** A plain `POST /api/v1/internal/reports/` only creates a Draft; Celery dispatch happens on a separate `POST .../{id}/analyze/`. The tool now always calls `analyze` and surfaces `celery_task_id`.
- **`create_report` 400s on free-form prompts.** `agent_report` (the only type today) requires `input_template` + `output_template` XML in `config` that the LLM can't construct. The tool now mirrors the web UI flow: list templates → find the seeded `is_default: true` for the requested `report_type` → GET its detail (which carries assembled XML in `config`) → fill the input XML with `research_objective` → POST with config + `template_source`. Explicit `config` bypasses the lookup.
- **`knowledge_search` default silently returned `[]`.** `search_effort="deep"` hits the Orca async pipeline via a redirect, returns `202 {task_id, status: "pending"}`, and we don't poll yet — so every default call looked empty. Changed default: `search_effort` is optional; omitted → run `fast` first, escalate to `deep` only if `fast` returns zero cards. Explicit effort (any value) is a single call.
- Broadened `knowledge_search` description to mention prediction markets (Polymarket) so the LLM routes betting/odds queries through it.

## Test plan

- [x] `npm --prefix workers test` → 57/57 pass (includes 7 new create_report tests + 5 new knowledge_search tests)
- [x] `npm --prefix workers run typecheck` → clean
- [x] `npm --prefix workers run registry:check` → ok (9 tools, server.json regenerated)
- [ ] Manual: create a report from a running Worker pointed at staging — confirm it leaves Draft and lands on `running`.
- [ ] Manual: `knowledge_search` with no `search_effort` on a query `fast` can answer — confirm one call, cards returned.
- [ ] Manual: same with a query that stumps `fast` — confirm fallback to `deep`.

### DIFF
--- a/workers/src/tools/create_report.test.ts
+++ b/workers/src/tools/create_report.test.ts
@@ -27,6 +27,12 @@ import {
 import type { Env } from "../env.js";
 import type { ToolContext } from "./types.js";
 import create_report from "./create_report.js";
+import {
+  jsonResponse,
+  mockFetchOnce,
+  mockFetchSequence,
+  requestFrom,
+} from "./__test_helpers.js";
 
 const ENV: Env = { DJANGO_BASE_URL: "https://trytako.com" };
 const CTX: ToolContext = { token: "sk-test", env: ENV };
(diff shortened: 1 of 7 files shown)


workers/src/tools/create_report.ts
```diff
@@ -103,18 +283,35 @@ const create_report = {
...
       throw new Error(
         "Tako create_report response missing both `report_id` and `id`",
       );
     }
+
+    // Kick off generation. The create call alone leaves the report as a
+    // Draft; `analyze` is what transitions it to RUNNING and dispatches
+    // the Celery task. Empty body is fine for non-brief report types —
+    // backend assembles config from what create already persisted.
+    const analyzed = await djangoPost<AnalyzeResponse>(
```
robertabbott: **Analyze failure leaves an orphaned Draft with no recovery path**

`djangoPost` for `/analyze/` runs after the report has already been created at line 274. If analyze throws (502, timeout, Celery queue down), the thrown `DjangoHttpError` doesn't carry `reportId` — the LLM has no way to surface the draft to the user or retry generation. Since there's no `analyze_report` MCP tool, the draft is effectively orphaned from the MCP perspective. Would it make sense to catch analyze failures and return a partial success (`status: "created_but_not_started"`, `report_id` included, plus a message explaining what happened) so the LLM can tell the user their report was created and they can trigger generation from the web UI?

workers/src/tools/create_report.test.ts
```diff
@@ -88,6 +119,298 @@ describe("create_report error forwarding", () => {
     expect(bad.body).toContain("company_deep_dive");
   });
 
+  it("kicks off generation by POSTing to /analyze/ after create", async () => {
```
robertabbott, line 87: **No test covering the create-succeeds-then-analyze-fails path**

The suite covers the happy path (create 201 → analyze 202) and the create-failure path (400 propagation), but there's no test for "create succeeds (201) then analyze fails (500 or timeout)." This is the scenario where the report is created as a Draft but the error doesn't include `report_id` — the current behavior isn't locked by a test. If someone later wraps the analyze call in a try/catch for partial-success handling, the suite wouldn't catch a regression in the error shape. Worth adding a test that mocks create as 201 then analyze as 500 and asserts the expected error type and whether `report_id` is recoverable?

workers/src/tools/knowledge_search.test.ts
```diff
@@ -0,0 +1,176 @@
...
+});
+
+function jsonResponse(status: number, body: unknown): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { "content-type": "application/json" },
+  });
+}
+
+function mockFetchSequence(
```
robertabbott: **Duplicated test helpers**

`mockFetchSequence`, `jsonResponse`, and `requestFrom` are copied verbatim from `create_report.test.ts`. With two tool test files using them today and more tools likely coming, would it be worth extracting these into a shared `workers/src/tools/__test_helpers.ts` before the third copy appears?

robertabbott (approved):

PR #49, opened 2026-04-27
feat(workers): keep workers.dev enabled + post-deploy smoke (TAKO-2611)

jfassio:
## Summary

Two related changes bundled — the second wouldn't have been verifiable end-to-end without the first.

**1. Re-enable `workers_dev` for both envs.** The previous staging deploy disabled the `*.workers.dev` URL by default (wrangler v4 behavior change), leaving the Worker unreachable. Adds explicit `workers_dev: true` per env + `preview_urls: true` for staging.

**2. Adds the TAKO-2611 post-deploy smoke test.**

- `workers/scripts/smoke.ts` — MCP SDK client running 9 checks: `/health`, `initialize`, `tools/list`, and 6 per-tool canaries.
- `.github/workflows/workers-smoke.yml` — `workflow_run` auto-trigger after each successful staging deploy + `workflow_dispatch` for manual runs against any base URL.

Tools exercised in smoke:

| Tool | Canary | Type |
|---|---|---|
| `knowledge_search` | `"US GDP"` query → save first `card_id` for chaining | read |
| `get_credit_balance` | no args → expect `details.credit_balance` | read |
| `list_reports` | `{limit:1}` → expect `reports[]` array | read |
| `get_chart_image` | `{pub_id}` from knowledge_search → `image_url` http(s) | read, chained |
| `open_chart_ui` | `{pub_id}` from knowledge_search → `structuredContent` | read, chained |
| `create_chart` | `{components:[]}` → expect `isError: true` | **negative** — validates input rejection without creating a chart |

Excluded by design: `create_report` / `get_report` (long-running), `explore_knowledge_graph` (slated for removal in #47).

## Out of scope (deferred, blocked on Bobby)

- **TAKO-2610 custom-domain bind** for \`mcp.staging.tako.com\` and \`mcp.tako.com\`. Requires Cloudflare Zone:Edit access on the \`tako.com\` zone, which I don't currently have. DNS today still resolves both to the legacy AWS-ECS Python deploy with broken TLS; replacing those records needs Zone perms. Workers remain reachable via \`*.workers.dev\` URLs in the meantime.
- **\`workflow_run\` auto-trigger validation in CI.** Requires \`CLOUDFLARE_API_TOKEN\` + \`CLOUDFLARE_ACCOUNT_ID\` GitHub secrets so \`workers-deploy.yml\` actually runs in CI; same access dependency. Manual \`workflow_dispatch\` of \`workers-smoke\` works today as long as \`TAKO_SMOKE_API_TOKEN\` is set.

A follow-up Linear issue captures the full unblock ask.

## Test plan

- [x] \`npm run typecheck\` clean (src + scripts)
- [x] \`npm test\` — 60/60 passing
- [x] \`npm run registry:check\` — no drift, 9 tools
- [x] \`npm run smoke\` against staging — all 9 canaries green (real \`knowledge_search\` results, real credit balance, real chained \`pub_id\`, \`create_chart\` negative test confirmed)
- [x] \`gh secret set TAKO_SMOKE_API_TOKEN\` on the repo so CI can run smoke
- [x] Manually dispatch \`workers-smoke\` once via GitHub UI to confirm CI path

### DIFF
--- a/.github/workflows/workers-smoke.yml
+++ b/.github/workflows/workers-smoke.yml
@@ -1,15 +1,17 @@
 name: workers-smoke
 
 # Triggers (TAKO-2611):
-#   - `workflow_run`: fire automatically after every successful `workers-deploy`
-#     run on `main`. The smoke target is the staging Worker by default; we
-#     could later read the deployed env from the upstream run if/when a prod
-#     auto-smoke is wanted.
-#   - `workflow_dispatch`: manual trigger so we can smoke prod post-promotion,
-#     or smoke any other base URL without redeploying.
+#   - `workflow_run`: fire automatically after a successful `workers-deploy`
+#     run that was triggered by a `push` to `main` (which always deploys
+#     staging). We DO NOT auto-smoke after `workflow_dispatch` deploys —
+#     those can target production and we don't have a prod-aware smoke yet,
+#     so silently smoking staging while prod was deployed would produce a
+#     misleading green badge. Manually dispatch this workflow to smoke any
+#     post-promotion prod deploy.
+#   - `workflow_dispatch`: manual trigger to smoke any base URL.
 #
 # `workflow_run` only fires for upstream runs that completed (any conclusion);
-# the `if:` on the job below filters to successful upstream runs.
+# the `if:` on the job below filters to successful + push-triggered upstreams.
 on:
   workflow_run:
     workflows: [workers-deploy]
@@ -17,12 +19,19 @@ on:
   workflow_dispatch:
     inputs:
       base_url:
-        description: 'Base URL of the Worker to smoke-test (no trailing slash)'
+        description: 'Base URL of the Worker to smoke-test (no trailing slash). Leave empty to use the staging default.'
         type: string
-        # Default points at the staging *.workers.dev URL while TAKO-2610
-        # (custom-domain bind for `mcp.staging.tako.com`) is pending. Flip
-        # to `https://mcp.staging.tako.com` once that ships.
+        # Default duplicated from the workflow-level `STAGING_BASE_URL` env
+        # below — GitHub Actions doesn't allow `${{ env.X }}` substitution
+        # in `workflow_dispatch` input defaults. If the staging URL ever
+        # changes, update both the env var AND this default in lockstep.
         default: 'https://tako-mcp-staging.bobby-118.workers.dev'
+
+# Single source of truth for the staging URL. Referenced from the smoke step
+# below; the `workflow_dispatch` input default has to repeat the value (see
+# note on the input above).
+env:
+  STAGING_BASE_URL: 'https://tako-mcp-staging.bobby-118.workers.dev'
 
 # Latest staging deploy is what we want to validate; cancel an in-flight
 # smoke if a newer deploy lands while it's running.
@@ -38,9 +47,17 @@ jobs:
     name: smoke
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    # workflow_run fires regardless of conclusion — gate to success only so we
-    # don't smoke a deploy that already failed.
-    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success'
+    # Run when:
+    #   - manually dispatched (any base URL the user picked), OR
+    #   - the upstream `workers-deploy` succeeded AND was triggered by a
+    #     `push` (which always deploys staging — see workers-deploy.yml). We
+    #     skip `workflow_dispatch`-triggered upstream runs because those may
+    #     have targeted production, and our smoke target is hardcoded to
+    #     staging here.
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.workflow_run.conclusion == 'success' &&
+       github.event.workflow_run.event == 'push')
     defaults:
       run:
         working-directory: workers
@@ -58,11 +75,18 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      # Smoke target: manual dispatch → user-supplied input; auto-trigger →
-      # always staging. Token comes from a GitHub Actions secret and is masked
-      # by Actions in logs; the smoke script also takes care never to print it.
+      # Smoke target precedence:
+      #   1. `workflow_dispatch` input (if set) — lets ops smoke any URL
+      #   2. `STAGING_BASE_URL` env (workflow-level constant)
+      #
+      # The smoke script (`scripts/smoke.ts`) requires `SMOKE_BASE_URL` to be
+      # set explicitly — there is no in-script default — so any source URL
+      # change happens in this YAML, not in TypeScript.
+      #
+      # `TAKO_SMOKE_API_TOKEN` is a GitHub Actions secret, automatically
+      # masked in logs; the smoke script also never prints it.
       - name: Smoke
         env:
-          SMOKE_BASE_URL: ${{ github.event_name == 'workflow_dispatch' && inputs.base_url || 'https://tako-mcp-staging.bobby-118.workers.dev' }}
+          SMOKE_BASE_URL: ${{ inputs.base_url || env.STAGING_BASE_URL }}
           TAKO_SMOKE_API_TOKEN: ${{ secrets.TAKO_SMOKE_API_TOKEN }}
         run: npm run smoke
--- a/.github/workflows/workers-smoke.yml
+++ b/.github/workflows/workers-smoke.yml
@@ -1,0 +1,68 @@
+name: workers-smoke
+
+# Triggers (TAKO-2611):
+#   - `workflow_run`: fire automatically after every successful `workers-deploy`
+#     run on `main`. The smoke target is the staging Worker by default; we
+#     could later read the deployed env from the upstream run if/when a prod
+#     auto-smoke is wanted.
+#   - `workflow_dispatch`: manual trigger so we can smoke prod post-promotion,
+#     or smoke any other base URL without redeploying.
+#
+# `workflow_run` only fires for upstream runs that completed (any conclusion);
+# the `if:` on the job below filters to successful upstream runs.
+on:
+  workflow_run:
+    workflows: [workers-deploy]
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      base_url:
+        description: 'Base URL of the Worker to smoke-test (no trailing slash)'
+        type: string
+        # Default points at the staging *.workers.dev URL while TAKO-2610
+        # (custom-domain bind for `mcp.staging.tako.com`) is pending. Flip
+        # to `https://mcp.staging.tako.com` once that ships.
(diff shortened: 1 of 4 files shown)


.github/workflows/workers-smoke.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Install dependencies
+        run: npm ci
+
+      # Smoke target: manual dispatch → user-supplied input; auto-trigger →
+      # always staging. Token comes from a GitHub Actions secret and is masked
+      # by Actions in logs; the smoke script also takes care never to print it.
+      - name: Smoke
+        env:
+          SMOKE_BASE_URL: ${{ github.event_name == 'workflow_dispatch' && inputs.base_url || 'https://tako-mcp-staging.bobby-118.workers.dev' }}
```
robertabbott: The `SMOKE_BASE_URL` for the auto-trigger path is hardcoded to staging, but `workers-deploy.yml` accepts a `workflow_dispatch` input that can target production. If someone manually dispatches a production deploy, the `workflow_run` that fires next will still smoke staging and report green — even if prod is broken. The header comment acknowledges reading the env from the upstream run is "for later", but right now a green smoke badge on a prod deploy could mislead. Would it be safer for the auto-trigger to skip when the upstream run targeted production (e.g. read `github.event.workflow_run.event` / inputs and bail) until the prod-aware path lands?

workers/wrangler.jsonc
```diff
@@ -29,13 +31,17 @@
...
+      "workers_dev": true,
+      "preview_urls": true
     },
     "production": {
       "name": "tako-mcp-production",
       "vars": {
         "DJANGO_BASE_URL": "https://trytako.com"
-      }
+      },
+      "workers_dev": true,
```
robertabbott, line 54: `workers_dev: true` on the production env keeps `tako-mcp-production.<account>.workers.dev` reachable as a production endpoint forever, not just while TAKO-2610's custom-domain bind is pending. Once `mcp.tako.com` resolves to the Worker, that secondary URL becomes an alternate prod entrypoint that bypasses anything attached to the custom domain (Zone-level rules, future WAF/rate limits, analytics tags). Is the long-term intent to keep the `workers.dev` URL live on prod, or flip it to `false` after the custom domain ships?

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,276 @@
...
+ * Secrets handling: the token is read from env, attached to the transport's
+ * `requestInit.headers`, and never logged. We print only structured result
+ * summaries (counts, tool names, sanitized URLs) — never request headers or
+ * raw response bodies.
+ */
+
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
+
+const DEFAULT_BASE_URL = "https://tako-mcp-staging.bobby-118.workers.dev";
```
robertabbott: `bobby-118.workers.dev` is the Cloudflare account's subdomain — the same literal appears here, in `workers-smoke.yml:25`, and `workers-smoke.yml:66`. If the account moves (e.g. personal → team account) all three need updating in lockstep, and a partial update silently smokes a stale deployment. Would it work to require an explicit `SMOKE_BASE_URL` (no default) and have the auto-trigger set it from a single workflow-level env var, so there's one source of truth?

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  // ----- f) create_chart NEGATIVE test ----------------------------------
+  // Smoke must not have side effects, so we verify the tool's input
+  // validation rejects an empty `components` array — this exercises the
+  // tool surface (registration, schema, error-mapping) without producing
+  // a real card.
+  const ccResult = await client.callTool({
+    name: "create_chart",
+    arguments: { components: [] },
```
robertabbott: `{components: []}` is rejected by `inputSchema`'s `.min(1)` (line 33 of `create_chart.ts`), so the request fails MCP input validation before the handler ever calls `djangoPost`. That means the only write tool's actual network path — auth, Django round-trip, error mapping for write 4xx — isn't exercised by smoke. The header comment frames this as testing "registration, schema, error-mapping" which is fair, but if `djangoPost` or write-path auth regressed, every read tool would catch it and `create_chart` would not. Would a schema-valid-but-server-rejected payload (e.g. one component with a bogus `component_type`) give better coverage while still being side-effect free?

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,276 @@
...
+      | { image_url?: string; pub_id?: string }
+      | undefined;
+    assert(giStructured, "get_chart_image missing structuredContent");
+    assert(
+      typeof giStructured.image_url === "string" &&
+        HTTP_URL_REGEX.test(giStructured.image_url),
+      `get_chart_image.image_url is not http(s): ${JSON.stringify(giStructured.image_url)}`,
+    );
+    ok(`get_chart_image {pub_id:${chainPubId}} → image_url present`);
+  } else {
```
robertabbott: If `knowledge_search "US GDP"` returns non-empty results but every `card_id` is null/empty, both `get_chart_image` and `open_chart_ui` log `↷ skipped` and the smoke still passes. That's itself a regression mode worth catching — `knowledge_search` returning no chartable cards for `"US GDP"` is exactly the kind of pipeline regression smoke should flag. Should the absence of a usable `card_id` fail the smoke instead of skipping?

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,276 @@
...
+    .find((id): id is string => typeof id === "string" && id.length > 0);
+
+  // ----- b) get_credit_balance ------------------------------------------
+  const cbResult = await callOk(client, "get_credit_balance", {});
+  const cbStructured = cbResult.structuredContent as
+    | { details?: Record<string, unknown> }
+    | undefined;
+  assert(cbStructured, "get_credit_balance missing structuredContent");
+  assert(
+    cbStructured.details && typeof cbStructured.details === "object",
```
robertabbott, line 225: The smoke asserts `details` is an object and reads `details["credit_balance"]`, but never validates the value's type or presence — `${balance ?? "<unset>"}` happily prints `<unset>` and the canary still passes. If the field is renamed (`balance`, `credits`) or moved into a sub-object, this regression goes unnoticed. Worth tightening to `typeof balance === "number"`, or is the looseness intentional because the API's exact shape isn't pinned yet?

robertabbott (commented):
Design pass — questions on the smoke wiring and the wrangler env config. Code-level pass to follow.

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,276 @@
...
+
+try {
+  // `as never` papers over an SDK-vs-strict-TS tension: the SDK types
+  // `StreamableHTTPClientTransport#sessionId` as `string | undefined`, which
+  // doesn't satisfy `Transport` under `exactOptionalPropertyTypes: true`.
+  await client.connect(transport as never);
+  const serverInfo = client.getServerVersion();
+  ok(
+    `initialize → ${serverInfo?.name ?? "<unknown>"} ` +
+      `${serverInfo?.version ?? ""}`.trim(),
```
robertabbott:
```ts
ok(
  `initialize → ${serverInfo?.name ?? "<unknown>"} ` +
    `${serverInfo?.version ?? ""}`.trim(),
);
```
The `.trim()` is applied to the second segment only, so when `serverInfo?.version` is empty, the inner expression collapses to `""` but the outer string still carries the trailing space from `${serverInfo?.name ?? "<unknown>"} ` — the log reads `initialize → tako-mcp ` (trailing space). Was the intent to trim the whole concatenation, i.e. `` (`initialize → ${name} ${version}`).trim() ``?

workers/wrangler.jsonc
```diff
@@ -29,13 +31,17 @@
...
+      "preview_urls": false
     }
   }
-}
+}
```
robertabbott: The diff shows `\ No newline at end of file` on the closing brace. Most editors and the POSIX text-file definition want a trailing `\n`; the previous version of the file had one. Worth restoring to keep diffs clean for the next edit?

robertabbott (commented):
Code-level pass — two small things.

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,356 @@
...
+  assert(cbStructured, "get_credit_balance missing structuredContent");
+  assert(
+    cbStructured.details && typeof cbStructured.details === "object",
+    "get_credit_balance.details is not an object",
+  );
+  const balance = cbStructured.details["credit_balance"];
+  const balanceNumeric =
+    typeof balance === "number"
+      ? balance
+      : typeof balance === "string" && Number.isFinite(Number(balance))
```
robertabbott: The new numeric coercion accepts an empty string (or all-whitespace string) as a valid `credit_balance` of `0` — `Number("") === 0` and `Number.isFinite(0) === true`. DRF's `DecimalField` serialization won't actually produce that in practice, so this is pedantic, but if you wanted to close the hole, an explicit `balance !== ""` (or `balance.trim() !== ""`) before the `Number(...)` call would do it. Optional.

workers/scripts/smoke.ts
```diff
@@ -0,0 +1,356 @@
...
+    }
+  } catch (err) {
+    if (err instanceof McpError) {
+      createChartFailedAsExpected = true;
+      // The SDK's InvalidParams (-32602) message includes "tool's output
+      // schema" when the failure is response-validation; surface that
+      // distinct case in the log so it's clear the handler ran.
+      createChartFailureMode = err.message.includes("output schema")
+        ? `McpError -32602 (server returned non-schema-conformant error response — handler ran, Django rejected)`
+        : `McpError ${err.code}: ${err.message.slice(0, 200)}`;
```
robertabbott, line 346: The fallback accepts `McpError -32602` as evidence the write path ran, with the comment noting the handler "doesn't synthesize a schema-conformant error payload". That's a real handler/SDK gap and the smoke is right not to block on it, but is there a Linear issue tracking the cleanup? If not, worth filing — once that's fixed, the `McpError` branch becomes dead code and should fail the smoke instead.

robertabbott (approved):
Re-reviewed `dfcd3fe`. All 8 prior comments addressed cleanly:

1. ✅ Auto-smoke gated on `workflow_run.event == 'push'` — manual prod deploys no longer trigger a misleading staging smoke.
2. ✅ Production `workers_dev: true` documented as temporary with `TODO(TAKO-2610)`.
3. ✅ `SMOKE_BASE_URL` now canonical via workflow-level env; in-script default removed.
4. ✅ `create_chart` negative now sends a schema-valid bogus `component_type` and reaches Django.
5. ✅ `assert(chainPubId, ...)` — missing card_id fails the smoke instead of silently skipping.
6. ✅ `credit_balance` validated as numeric (number or numeric string).
7. ✅ `.trim()` consolidated into a single template literal.
8. ✅ Trailing newline restored at EOF.

Two small non-blocking follow-up notes inline. LGTM.

PR #52, opened 2026-04-27
fix(workers): poll async deep knowledge_search to completion (TAKO-2686)

jfassio:
## Summary

Slack reports of \`knowledge_search\` with \`search_effort: \"deep\"\` returning silent empty results trace to an unimplemented gap: Tako's Django backend goes async on deep mode (\`202 {task_id, status: \"pending\"}\` + status polling) and the Worker handler wasn't polling — it just read \`outputs.knowledge_cards\` off the 202 body, found nothing, returned \`count: 0\`. The legacy Python MCP had the same gap.

This PR implements client-side polling.

## What it does

- Detects the async-task response shape on the initial POST and switches into polling mode.
- Polls \`/api/v1/knowledge_search/async/status/\` every **10s** for up to **290s** total, exiting on the first terminal state.
- Returns \`result.outputs.knowledge_cards[]\` on \`COMPLETED\`.
- Throws with the Django error on \`FAILED\` / \`INTERRUPTED\`.
- On budget exhaustion, throws with **accumulated progress info** so the LLM gets actionable feedback (\`\"did not complete within 290s (32 progress events; last: agent_status)\"\`) instead of silent empty.
- Accumulates the \`events[]\` array across polls via the \`since_index\` query param to avoid refetching duplicates.

## Sizing math

Three constraints intersect:

| Constraint | Value | How we sit relative to it |
|---|---|---|
| Tako's deep orchestrator timeout (\`orchestrator_deep_config.py\`) | 300s | We poll up to 290s — captures essentially all of it |
| Cloudflare Workers subrequest cap (default tier) | 50 per invocation | 1 POST + 29 GETs = **30 subrequests**, ~20 of headroom |
| MCP client's per-tool timeout (MCP TS SDK \`DEFAULT_REQUEST_TIMEOUT_MSEC\`) | 60_000ms | Practical ceiling for clients on defaults — our budget is the right shape for clients that pass \`options.timeout\`. Server-sent progress notifications would lift the 60s lid for default clients but that's a separate ticket. |

## Tests

- 7 new vitest cases: async-completion on first poll, PENDING→IN_PROGRESS→COMPLETED, FAILED with error, INTERRUPTED, budget exhaustion, status-endpoint 404, and fast→deep auto-escalation through the polling path.
- Mock counts scale from \`__test_only__\` exports so future budget tweaks don't silently break the budget-exhaustion test.

## End-to-end verification against \`mcp.staging.tako.com\`

| Call | Pre-fix | Post-fix |
|---|---|---|
| \`knowledge_search {query: \"US GDP\"}\` (omitted effort) | 5 results, ~2s (fast path) | Unchanged — same fast path |
| \`knowledge_search {query: \"CASAFARI Portuguese proptech Lisbon real estate\", search_effort: \"deep\"}\` | **silent \`count: 0\`** in <2s | **\`count: 3\` real cards** in **196.8s** via polling |

Reports: \`CASAFARI Corporate Events\`, \`CASAFARI Entities\`, \`Lisbon Population Trend\` — each with usable \`pub_id\` chains for \`open_chart_ui\`.

## What's still on the architectural roadmap

For deep queries that genuinely take >60s, default-timeout MCP clients (e.g. Claude Code on out-of-the-box settings) will still cancel before we return cards — same wall as before, just hit a few seconds later. The proper fix is to either:

1. Emit MCP progress notifications during polling (resets client timeout if client opts in to \`resetTimeoutOnProgress\`), or
2. Split into multi-tool architecture (\`knowledge_search_deep_start\` + \`knowledge_search_deep_status\`) so the LLM orchestrates polling across multiple tool calls

Both are follow-up work outside this PR's scope. Today's fix is "stop returning silent empty + give the LLM information to act on" which is itself a real win.

## Test plan

- [x] \`npm run typecheck\` clean
- [x] \`npm test\` — 67/67 passing (12 in \`knowledge_search.test.ts\`)
- [x] \`npm run registry:check\` — no drift, 8 tools
- [x] \`npm run smoke\` against staging — all 9 canaries green (omitted-effort path unchanged)
- [x] Explicit deep call against staging returns real cards in budget (CASAFARI run above)
- [ ] Reviewer: optional — re-run \`SMOKE_BASE_URL=https://mcp.staging.tako.com QUERY=... npm run smoke\` against another deep query to spot-check

## Tickets

- Closes TAKO-2686
- Related: future ticket for MCP progress notifications / multi-tool deep architecture (mentioned in this PR's body, not yet filed)


workers/src/tools/knowledge_search.ts
```diff
@@ -8,18 +8,59 @@
...
+// (Orca pipeline routinely runs 60-300s); a 10s detection delta on
+// completion is irrelevant against that wait.
+const POLL_INTERVAL_MS = 10_000;
+const POLL_BUDGET_MS = 290_000;
+const STATUS_REQUEST_TIMEOUT_MS = 10_000;
+
+// Django's task status discriminator. `PENDING` and `IN_PROGRESS` are the
+// non-terminal states we keep polling through; the rest exit the loop.
+// Source: `tako/app/backend/monolith/models.py:2719` (AsyncTaskStatusChoices).
+const TERMINAL_STATES = new Set(["COMPLETED", "FAILED", "INTERRUPTED"]);
```
robertabbott: `TERMINAL_STATES` is defined alongside a comment that names its source-of-truth (`AsyncTaskStatusChoices`) and is re-exported via `__test_only__` at line 322 with the comment "exposing it keeps the test in lockstep with any future state the backend introduces." But the loop hard-codes the literals (`"COMPLETED"` at line 223, `"FAILED" || "INTERRUPTED"` at line 226), and the test file imports `__test_only__` but only reads `POLL_BUDGET_MS` / `POLL_INTERVAL_MS`. The set is never read by any code path. Should the loop and/or the tests actually use `TERMINAL_STATES.has(status.status)` so the constant earns its keep, or should it be removed?

workers/src/tools/knowledge_search.ts
```diff
@@ -66,18 +107,143 @@ const outputSchema = z.object({
...
+    if (status.events && status.events.length > 0) {
+      latestEvents = latestEvents.concat(status.events);
+      const maxId = status.events.reduce(
+        (acc, e) => (typeof e.id === "number" && e.id > acc ? e.id : acc),
+        nextSinceIndex - 1,
+      );
+      nextSinceIndex = maxId + 1;
+    }
+
+    if (status.status === "COMPLETED") {
```
robertabbott: The `isAsyncTaskInitiation` guard at line 159 is documented as casing-tolerant ("the guard tolerant of `status` casing differences (`pending` vs `PENDING`) that we've seen between the initial 202 and subsequent polls"), and the PR body confirms the initial 202 uses lowercase `"pending"`. But the terminal checks at lines 223 and 226 compare strictly against uppercase `"COMPLETED"` / `"FAILED"` / `"INTERRUPTED"`. If the status endpoint ever returns lowercase (older deploy, different code path, future change), the loop falls through to "keep polling" and burns the full 290s budget before throwing "did not complete." Should the comparisons normalize via `.toUpperCase()` to match the casing tolerance the type guard already accepts?

workers/src/tools/knowledge_search.ts
```diff
@@ -66,18 +107,143 @@ const outputSchema = z.object({
...
+  // avoid re-fetching the same events every poll — the response only
+  // includes events at index >= since_index.
+  let nextSinceIndex = 0;
+  let latestEvents: AsyncTaskEvent[] = [];
+
+  // Poll-first ordering: a fast-completing deep task can return cards on
+  // the very first GET without paying the interval delay. Slow tasks then
+  // pay the delay between subsequent polls.
+  while (true) {
+    const status = await djangoGet<AsyncTaskStatus>(
```
robertabbott: The `djangoGet` at line 201 has no surrounding try/catch, so a single transient `502`/`503` from the status endpoint (Django restart, LB blip mid-deploy, etc.) throws out of the entire tool call, even though the backend task is still running and would have completed on the next poll. The user sees a tool error and has no signal that retrying within a few seconds would just hit the same task. Was leaving this un-retried a deliberate "fail loud, let the LLM retry" choice, or worth at least one in-loop retry on transport-level errors before giving up?

workers/src/tools/knowledge_search.ts
```diff
@@ -66,18 +107,143 @@ const outputSchema = z.object({
...
+  const list = events ?? [];
+  const lastTyped = [...list]
+    .reverse()
+    .find((e) => typeof e.data?.event_type === "string");
+  if (list.length === 0) return "no progress events emitted";
+  const lastEventType = lastTyped?.data?.event_type ?? "untyped";
+  return `${list.length} progress event${list.length === 1 ? "" : "s"}; last: ${lastEventType}`;
+}
+
+async function pollAsyncKnowledgeSearch(
```
robertabbott, line 237: `pollAsyncKnowledgeSearch` is named for `knowledge_search` but the shape (`POST` → `202 {task_id}` → poll status endpoint → terminal state) looks generic to any Tako async task. The PR body hints at follow-up tickets (progress notifications, multi-tool deep architecture), and `create_report.ts` already deals with async-task semantics via `celery_task_id`. Is there a reason to keep this inside `knowledge_search.ts` rather than colocating it with `djangoGet`/`djangoPost` once a second consumer arrives? Fine to defer; flagging the seam.

workers/src/tools/knowledge_search.test.ts
```diff
@@ -143,3 +158,285 @@ describe("knowledge_search fast-first-deep-fallback", () => {
...
+      ).rejects.toThrow(/failed.*Orca pipeline blew up/i);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("throws when the budget is exhausted before the task terminates", async () => {
+    vi.useFakeTimers();
+    try {
+      // POST + enough PENDING responses to outlast the budget (50s / 2s
```
robertabbott: The inline comment says "50s / 2s interval = 25 max polls; 40 is comfortable headroom," but `POLL_BUDGET_MS` is 290000 and `POLL_INTERVAL_MS` is 10000, so the real math is 290/10 = 29 polls. `MAX_PENDING_POLLS` is computed correctly from the constants, but the comment is from an earlier draft of the budget. Worth updating to match (or just deleting since `MAX_PENDING_POLLS` is self-documenting)?

workers/src/tools/knowledge_search.ts
```diff
@@ -66,18 +107,143 @@ const outputSchema = z.object({
...
+): Promise<KnowledgeCard[]> {
+  const startedAt = Date.now();
+  // Tracked across polls so the timeout/failure message can describe how
+  // far the pipeline got. We pass `since_index` to the status endpoint to
+  // avoid re-fetching the same events every poll — the response only
+  // includes events at index >= since_index.
+  let nextSinceIndex = 0;
+  let latestEvents: AsyncTaskEvent[] = [];
+
+  // Poll-first ordering: a fast-completing deep task can return cards on
```
robertabbott: The "Poll-first ordering" comment claims a fast-completing deep task can return cards on the first GET without paying the interval delay. But the same file at line 54 notes "Orca pipeline routinely runs 60-300s" — the chance of `COMPLETED` on the first GET (≈1ms after POST returns) is essentially zero in practice. The ordering is fine, but the rationale could be honest: it just keeps the loop body in a single shape rather than poll-after-sleep. Worth softening?

robertabbott (approved):

PR #53, opened 2026-04-28
feat(workers): MCP Apps widget for open_chart_ui (TAKO-2679)

jfassio:
## Summary

- Drops the JSON-blob-with-inline-iframe-HTML payload from `open_chart_ui`. Output is now `pub_id` / `embed_url` / `image_url` / `dark_mode` / `width` / `height` — no more raw `<iframe>` markup leaking into chat as text.
- Adds a generic `appUiResource` hook on `ToolModule`. When a tool declares it, `mcp.ts` registers the bundle as an MCP resource (`mimeType: text/html;profile=mcp-app`, CSP `frameDomains` pinned to the public embed origin) and threads the URI into `_meta` under all three of `ui.resourceUri` (open MCP Apps spec), `ui/resourceUri` (legacy flat), and `openai/outputTemplate` (OpenAI Apps SDK namespace).
- `open_chart_ui` ships a thin widget bundle: a single `<iframe>` whose `src` updates from `structuredContent.embed_url` arriving over either the `window.openai` global (ChatGPT Apps SDK — multi-key probing because the documented `toolOutput` is null for dev-mode custom connectors) or `ui/notifications/tool-result` postMessage ( / VS Code / Goose).
- Adds `extraContentBlocks` hook for an inline base64 PNG fallback. **Skipped when an `appUiResource` is also set** — in combination, the response crossed ChatGPT's internal size guard (~150K tokens) and silently disabled widget data flow. The image is also redundant when the widget renders the chart interactively.

## Test plan

- [x] `npm test` — 83/83 passing (added 10 unit tests for `open_chart_ui` handler / `extraContentBlocks` hook / `appUiResource` factory; 4 integration tests for `_meta.ui` on `tools/list`, `resources/list`, and `resources/read`)
- [x] `npm run registry:check` — ok
- [x] Verified live in ChatGPT (custom connector + developer mode) — interactive Tako chart renders inline
- [ ] Other MCP clients (Claude desktop, VS Code Insiders, Goose) — fall back to `structuredContent` + markdown link; widget will activate automatically when those clients support MCP Apps

## Notes

- This PR ships **only the chart fix**. OAuth follow-up (separate PR — replaces the static-Bearer auth path with a real OAuth flow for ChatGPT custom connectors).
- Smoke updated (`workers/scripts/smoke.ts`) to assert `_meta.ui.resourceUri` on the tool listing and the registered resource on `resources/list`. Soft-warns rather than fails so the smoke is still useful even if only the widget piece broke in a deploy.

### DIFF
--- a/workers/src/tools/open_chart_ui.ts
+++ b/workers/src/tools/open_chart_ui.ts
@@ -266,8 +266,8 @@ const WIDGET_HTML = `<!doctype html>
   // Subscribe to host updates. Multiple event-name candidates because
   // OpenAI's emitted name has drifted across SDK releases. Bind on both
   // \`window\` and \`document\` because some hosts dispatch on one and not
-  // the other. Also handles re-runs when the tool is invoked again in
-  // the same widget.
+  // the other. Prevents duplicate renders from redundant events within a
+  // single tool call (\`render()\` is one-shot via the \`rendered\` flag).
   var EVENT_NAMES = [
     "openai:set_globals",
     "openai:tool_result",
@@ -394,6 +394,10 @@ const open_chart_ui = {
       // that doesn't look like an image.
       if (!contentType.startsWith("image/")) return [];
       const buffer = await response.arrayBuffer();
+      // 0-byte 200 is plausible if a renderer returned early; emitting a
+      // `{ data: "", mimeType: "image/png" }` block would have clients try
+      // to render an invalid image. Mirror the oversize bail.
+      if (buffer.byteLength === 0) return [];
       if (buffer.byteLength > MAX_INLINE_PNG_BYTES) return [];
       return [
         {
--- a/workers/src/tools/open_chart_ui.ts
+++ b/workers/src/tools/open_chart_ui.ts
@@ -1,43 +1,44 @@
 /**
- * `open_chart_ui` — return everything a client needs to render an interactive
- * Tako chart iframe.
+ * `open_chart_ui` — return an inlined chart image plus URLs for the static
+ * PNG and the interactive embed.
  *
- * Ports `open_chart_ui` from `src/tako_mcp/server.py:691`.
+ * Originally ported from `open_chart_ui` in `src/tako_mcp/server.py:691`,
+ * which used `mcp-ui`'s `create_ui_resource()` to emit an MCP `resource`
+ * content block containing a live iframe. The TS port previously stuffed a
+ * ready-to-render `<iframe>` HTML string into a JSON output field; the LLM
+ * helpfully echoed it as text and clients displayed `<iframe …></iframe>`
+ * verbatim. Iteration two switched to `image_url` + `embed_url` and asked
+ * the LLM to render `![chart](image_url)` — better, but claude.ai web gates
+ * external markdown images behind a "Show Image" click-to-load.
  *
- * KNOWN REGRESSION vs the Python reference: the Python tool returns a
- * `list[UIResource]` via the `mcp-ui` library, which MCP-UI-aware clients
- * (e.g. Claude Desktop with the MCP-UI extension) auto-render as a live
- * interactive iframe. The TS registry contract (`mcp.ts:95-117`) currently
- * only emits `type: "text"` content blocks, so we cannot hand back a raw
- * MCP-UI `resource` content block without a `ToolModule` contract
- * extension (e.g. an optional `toContentBlocks(output)` hook). That
- * extension is out of scope for Phase 2 port. Clients that previously saw
- * an auto-rendered chart here will now see a JSON payload containing the
- * iframe HTML — a real UX regression for those clients.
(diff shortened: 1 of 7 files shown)


workers/src/tools/types.ts
```diff
@@ -77,6 +124,26 @@ export interface ToolModule<
...
+  /**
+   * Optional hook to append extra MCP content blocks (image, audio, resource)
+   * after the default JSON-stringified text block. Called once per
+   * `tools/call`, after `handler` resolves. Tools should treat this as
+   * best-effort presentation: if the hook throws or returns `[]`, the text +
+   * `structuredContent` pair already provides a working response.
+   *
+   * Example: `open_chart_ui` uses this to inline a base64 PNG so MCP clients
+   * (claude.ai etc.) render the chart without a click-to-load gate.
+   */
```
robertabbott, line 142: Nit: `extraContentBlocks` is silently bypassed in `mcp.ts` when `appUiResource` is also declared on the same tool (to avoid the ChatGPT ~150K token size guard). Since the two hooks are defined as independent optionals here, would it be worth noting the interaction in this JSDoc — something like "Skipped when `appUiResource` is also set; see `mcp.ts` for details" — so the next person declaring both hooks discovers the coupling at the interface rather than in the wiring?

workers/src/tools/open_chart_ui.ts
```diff
@@ -108,59 +330,84 @@ const open_chart_ui = {
...
+      if (!response.ok) return [];
+      const contentType = response.headers.get("content-type") ?? "";
+      // Defensive: the PNG endpoint should always return image/png, but
+      // an upstream redirect to an HTML error page would otherwise let us
+      // base64 the HTML and ship it as `mimeType: "image/png"` — a
+      // garbage block the client would try to render. Reject anything
+      // that doesn't look like an image.
+      if (!contentType.startsWith("image/")) return [];
+      const buffer = await response.arrayBuffer();
+      if (buffer.byteLength > MAX_INLINE_PNG_BYTES) return [];
```
robertabbott, line 401: Nit: `buffer.byteLength > MAX_INLINE_PNG_BYTES` guards the upper bound, but a 0-byte `arrayBuffer()` (plausible on a 200 from a renderer that returned early) would produce `{ type: "image", data: "", mimeType: "image/png" }` — an invalid image block that clients would try to render. Should there be a `buffer.byteLength === 0` return-`[]` guard here, matching the oversize path?

workers/src/tools/open_chart_ui.ts
```diff
@@ -54,51 +55,272 @@ const inputSchema = z.object({
...
+  function notifyHeight(h) {
+    try {
+      if (window.openai && typeof window.openai.notifyIntrinsicHeight === "function") {
+        window.openai.notifyIntrinsicHeight(h);
+      }
+    } catch (e) { /* ignore */ }
+  }
+
+  function render(structuredContent) {
+    if (rendered) return true;
```
robertabbott, line 201: Nit: `render()` bails with `if (rendered) return true;` here, making the widget one-shot. But the comment on the event listener registration (~line 269 in the template) says "Also handles re-runs when the tool is invoked again in the same widget." If MCP Apps hosts always create a fresh widget per tool call, the flag is just double-render prevention and the comment is slightly misleading — worth tightening to "prevents duplicate renders from redundant events within a single tool call"?

robertabbott (approved):
Clean PR — good test coverage, solid security boundaries (URL scheme validation at both server and widget layers, content-type checks, size caps), and the multi-client compatibility work is thoroughly documented. Three minor nits inline, none blocking.

workers/src/tools/types.ts
jfassio: Good call — added a paragraph to the JSDoc on `extraContentBlocks` calling out that it's bypassed when `appUiResource` is also set, the ChatGPT ~150K-token rationale, and a pointer to `mcp.ts`. (2eba1c8)

workers/src/tools/open_chart_ui.ts
jfassio: Good catch — added `if (buffer.byteLength === 0) return [];` alongside the oversize guard, with a comment noting the renderer-returned-early case. New unit test covers it. (2eba1c8)

workers/src/tools/open_chart_ui.ts
jfassio: Agreed — replaced with "Prevents duplicate renders from redundant events within a single tool call (`render()` is one-shot via the `rendered` flag)." Matches what the code actually does. (2eba1c8)

jfassio:
Addressed 3 review nits in 2eba1c8: `extraContentBlocks`/`appUiResource` interaction noted in JSDoc, 0-byte buffer guard added (with test), widget event-listener comment tightened. Typecheck + 84/84 tests + registry:check all green. @claude-address

PR #54, opened 2026-04-28
feat(workers): multi-user OAuth via Stytch + JWT-stateless tokens (TAKO-2679)

jfassio:
## Summary

Adds OAuth 2.1 + DCR + PKCE on the MCP Worker so consumer hosts (Claude.ai, ChatGPT) can connect to Tako with **per-user identity**. Each user authenticates via Tako's existing Stytch project on a Worker-hosted login page; the Worker then calls `/api/v1/api_token/` with the resulting Stytch session JWT to retrieve the user's Tako API token, encrypts it with AES-GCM, and embeds it in signed access tokens. Future MCP calls decrypt the token from the access JWT and forward to Django as `X-API-Key`, so each user's calls hit Tako as themselves.

Closes the loop on [TAKO-2679](https://linear.app/trytako/issue/TAKO-2679/oauth-shim-for-consumer-mcp-hosts-claudeai-chatgpt). Validated end-to-end against ChatGPT in staging — see screenshots in the Linear ticket.

## How it works

```
Claude.ai / ChatGPT
    │
    ▼
GET /authorize  ── no session cookie? ─▶  set state cookie, 302 to /login
    │
    ▼
GET /login  ── renders Stytch SDK ─▶  user picks Google or magic link
    │
    ▼
Stytch  ── redirect ─▶  GET /oauth/stytch_callback
    │
    ▼
/oauth/stytch_callback:
  1. exchange ?token for Stytch session JWT + user info
  2. call Tako /api/v1/api_token/ with the JWT as Cookie header
  3. encrypt the Tako token with OAUTH_ENC_KEY (AES-GCM 256)
  4. mint our own session JWT, set tako_oauth_session cookie
  5. 302 back to /authorize
    │
    ▼
GET /authorize  ── with session ─▶  render consent page (Allow / Deny)
    │
    ▼
POST /authorize  ── Allow ─▶  mint auth code, 302 back to client
    │
    ▼
POST /token  ── swap auth code for access + refresh JWTs
    │
    ▼
POST /mcp  with `Authorization: Bearer <access JWT>`
  ── verify signature, decrypt enc_tako_token from claims,
     forward to Django with X-API-Key: <user's Tako token>
```

The existing static-Bearer Claude Code path on `/mcp` is preserved unchanged: a non-JWT bearer falls through to raw-token forwarding.

## Architecture decisions

**JWT-stateless, not KV-backed** 
- Per-request hot path: HMAC verify + AES-GCM decrypt, no I/O.
- Tradeoff: no instant per-grant revocation (delayed up to TTL — 15min for access, 14d for refresh).
- Mitigation: rotating the user's Tako API token at trytako.com instantly invalidates every JWT we ever issued for them (Django returns 401 on the wrapped X-API-Key).
- Comparison doc: `/tmp/oauth-real-impl-comparison.md`.

**Two crypto keys, separate purposes:**
- `OAUTH_SIGN_KEY` (HS256) — signs every JWT.
- `OAUTH_ENC_KEY` (AES-256-GCM) — encrypts the Tako API token claim only.
- Held separately so a future signing-key rotation doesn't require coordinated re-encryption of in-flight tokens.

**Worker-only, no Django changes.**
- Stytch JWTs are signed by Stytch and validated by signature in Django's `StytchAuthBackend` — works whether the JWT arrives via a real browser cookie or via the Worker's server-side `Cookie:` header.
- Login page hosted on `mcp.tako.com` (not trytako.com) — slightly different brand surface, but skips ~2 weeks of cross-repo coordination.

## File layout

```
src/oauth/
├── jwt.ts          (HS256 + AES-GCM primitives, Web Crypto only)
├── types.ts        (shared claim shapes)
├── cookies.ts      (HttpOnly + Secure + SameSite=Lax cookie helpers)
├── stytch.ts       (Stytch REST API client — direct fetch, no SDK)
├── identity.ts     (Stytch JWT → Tako API token via /api/v1/api_token/)
├── handlers.ts     (all HTTP handlers — discovery, DCR, /authorize, /token, /login, /oauth/stytch_callback)
├── access.ts       (tryResolveOAuthAccessToken for /mcp integration)
└── *.test.ts       (focused unit tests per module)
```

Plus edits to `env.ts` (6 new optional secrets), `index.ts` (route the new endpoints), `mcp.ts` (try OAuth bearer first, fall through to raw token).

## Self-review

Two iterations of unbiased code review (no conversation context carried over).

| | Iteration 1 | Iteration 2 |
|---|---|---|
| Critical | 2 | **0** |
| Important | 6 | 6 |
| Minor | 9 | 6 |

Both critical issues from iteration 1 fixed in commit `181dc29`:
- Login page `escapeHtml`-then-`JSON.stringify` was double-escaping Stytch tokens.
- Single-use auth codes — design constraint of JWT-stateless, documented as TODO.

Iteration 2 produced **zero new critical issues**. The 6 important items were addressed in commit `78ca32c`:
- Clickjacking (`X-Frame-Options: DENY` + CSP `frame-ancestors 'none'` on all HTML).
- `Cache-Control: no-store` on auth pages (don't cache consent's PII).
- `client_id` JWTs now expire after 1 year (no forever-valid registrations).
- `scope` validated against `SUPPORTED_SCOPES` — only `mcp` today.
- Discovery endpoints return 404 when OAuth unconfigured.
- Refresh-token rotation gap — same JWT-stateless constraint as auth-code reuse, parallel TODO comment.

Other security hardening from review:
- `redirect_uri` scheme validation in `/register` (rejects `javascript:`, `data:`, non-loopback `http:`).
- 4KB body cap + control-char rejection on `/register`.
- `client_id` cross-check at `/token` (belt-and-suspenders vs PKCE).
- Stytch JWT shape validation before injecting into `Cookie:` header (header-injection guard).
- Form-action on consent rebuilt from validated params instead of echoing raw `url.search`.
- 30s clock-skew tolerance on JWT `exp`.
- `temporarily_unavailable` (RFC 6749) instead of ad-hoc `service_unavailable`.

## Known gaps deferred to a follow-up ticket

All flagged in code with `KNOWN GAP (TAKO-2679 follow-up)` comments:

1. **Single-use auth codes** — 60s TTL + PKCE bound replay risk; KV-backed `jti` set would close it.
2. **Refresh-token rotation with replay detection** — same constraint, same fix.
3. **State-cookie nonce via Stytch `state` param** — RFC 9700 belt-and-suspenders; cookie alone is adequate under the threat model (no XSS on `mcp.tako.com` + HTTPS-only).

## Tests

- **138 tests pass** (71 existing + 67 new across the OAuth modules).
- Coverage spans: HS256 round-trip, RFC 7636 PKCE vector, AES-GCM tamper detection, key rotation behavior, cookie helpers, DCR happy path + scheme rejection (javascript:/data:/non-loopback http:) + control-char rejection + oversized body, /authorize redirect-to-login flow, consent rendering with HTML-injection safety, /token PKCE mismatch + redirect_uri mismatch + client_id mismatch + refresh grant, scope validation, expired client_id rejection, discovery 404 when disabled, security headers on HTML responses, OAuth-disabled fall-through.
- `npm run typecheck` ✅
- `npm run registry:check` ✅

## Deploy checklist

Worker secrets (per env):
- `OAUTH_SIGN_KEY` — `openssl rand -base64 48`
- `OAUTH_ENC_KEY` — `openssl rand -base64 32` (must decode to exactly 32 bytes for AES-256)
- `STYTCH_PROJECT_ID` — from Stytch dashboard
- `STYTCH_SECRET` — from Stytch dashboard
- `STYTCH_PUBLIC_TOKEN` — from Stytch dashboard
- `STYTCH_BASE_URL` — `https://test.stytch.com` or `https://api.stytch.com`

Stytch dashboard:
- Add `https://mcp.tako.com/oauth/stytch_callback` and `https://mcp.staging.tako.com/oauth/stytch_callback` to Redirect URLs, with both **Login** and **Signup** enabled.

Without these set, the OAuth endpoints return 503 and the Worker still serves the static-Bearer Claude Code path on `/mcp` unchanged.

## Test plan

- [ ] Set staging secrets per the checklist above
- [ ] `npx wrangler deploy --env staging`
- [ ] `curl https://tako-mcp-staging.bobby-118.workers.dev/.well-known/oauth-authorization-server | jq` → expect full metadata
- [ ] In ChatGPT (Pro/Business + Developer Mode): add custom connector `https://tako-mcp-staging.bobby-118.workers.dev/mcp`
- [ ] Walk through Sign-in-with-Google → consent page should show your Tako email → click Allow
- [ ] In a new chat: ask for a Tako tool call (`knowledge_search`, `get_credit_balance`); verify it succeeds
- [ ] Verify Claude Code with a raw Tako API token still works (backwards-compat smoke test)
- [ ] Repeat in production after staging is green

### DIFF
--- a/workers/src/oauth/handlers.ts
+++ b/workers/src/oauth/handlers.ts
@@ -1,0 +1,866 @@
+/**
+ * HTTP handlers for the OAuth 2.1 + DCR + PKCE surface, plus the
+ * Stytch-mediated login dance that backs it. Every handler is a pure
+ * `(request, env) => Response` function so they're easy to wire into
+ * `index.ts` and easy to unit-test in isolation.
+ *
+ * Flow overview:
+ *
+ *   Claude.ai / ChatGPT
+ *      │
+ *      ▼
+ *   GET /authorize ──no session?──▶ set state cookie, 302 to /login
+ *      │
+ *      ▼
+ *   GET /login ──renders Stytch UI──▶ user picks Google / magic-link
+ *      │
+ *      ▼
+ *   Stytch (Google / email) ──redirect──▶ GET /oauth/stytch_callback
+ *      │
+ *      ▼
+ *   /oauth/stytch_callback:
+ *     1. exchange ?token for Stytch session JWT + user info
+ *     2. call Tako /api/v1/api_token/ with the JWT as a Cookie header
+ *     3. encrypt the Tako token, mint our own session JWT
+ *     4. set `tako_oauth_session` cookie, 302 back to /authorize
+ *      │
+ *      ▼
+ *   GET /authorize ──with session──▶ render consent page (Allow / Deny)
+ *      │
+ *      ▼
+ *   POST /authorize ──Allow──▶ mint auth code, 302 back to client
+ *      │
+ *      ▼
+ *   POST /token ──swap auth code for access + refresh JWTs──▶ client done
+ *
+ * State the Worker holds:
+ *   • `tako_oauth_state` cookie  — short JWT, OAuth params across Stytch
+ *   • `tako_oauth_session` cookie — short JWT, user identity + enc Tako token
+ *   • Auth codes / access tokens / refresh tokens — all signed JWTs, no DB
+ */
+
+import type { Env } from "../env.js";
+import {
+  encryptAesGcm,
+  sha256B64Url,
+  signJwt,
+  verifyJwt,
+} from "./jwt.js";
+import {
+  authenticateStytchToken,
+  primaryEmail,
+  StytchError,
+  type StytchTokenKind,
+} from "./stytch.js";
+import { fetchTakoApiToken, IdentityError } from "./identity.js";
+import {
+  buildClearCookie,
+  buildSetCookie,
+  readCookie,
+  SESSION_COOKIE,
+  SESSION_COOKIE_MAX_AGE_S,
+  STATE_COOKIE,
+  STATE_COOKIE_MAX_AGE_S,
+} from "./cookies.js";
+import type {
+  AccessTokenClaims,
+  AuthCodeClaims,
+  ClientIdClaims,
+  RefreshTokenClaims,
+  SessionCookieClaims,
+  StateCookieClaims,
+} from "./types.js";
+
+/* --------------------------- TTLs --------------------------- */
+
+const ACCESS_TOKEN_TTL_S = 15 * 60;
+const REFRESH_TOKEN_TTL_S = 14 * 24 * 60 * 60;
+const AUTH_CODE_TTL_S = 60;
+
+/* --------------------------- Config helpers --------------------------- */
+
+interface OAuthConfig {
+  signKey: string;
+  encKey: string;
+  stytch: {
+    projectId: string;
+    secret: string;
+    publicToken: string;
+    baseUrl: string;
+  };
+}
+
+/**
+ * Pull the OAuth config bundle out of `Env`, returning `null` if any
+ * required field is missing. Every handler that touches OAuth state
+ * gates on this so the Worker still serves the static-Bearer Claude
+ * Code path when OAuth is intentionally disabled in an env.
+ */
+function readConfig(env: Env): OAuthConfig | null {
+  const signKey = env.OAUTH_SIGN_KEY;
+  const encKey = env.OAUTH_ENC_KEY;
+  const projectId = env.STYTCH_PROJECT_ID;
+  const secret = env.STYTCH_SECRET;
+  const publicToken = env.STYTCH_PUBLIC_TOKEN;
+  const baseUrl = env.STYTCH_BASE_URL;
+  if (
+    typeof signKey !== "string" ||
+    signKey.length === 0 ||
+    typeof encKey !== "string" ||
+    encKey.length === 0 ||
+    typeof projectId !== "string" ||
+    projectId.length === 0 ||
+    typeof secret !== "string" ||
+    secret.length === 0 ||
+    typeof publicToken !== "string" ||
+    publicToken.length === 0 ||
+    typeof baseUrl !== "string" ||
+    baseUrl.length === 0
+  ) {
+    return null;
+  }
+  return {
+    signKey,
+    encKey,
+    stytch: { projectId, secret, publicToken, baseUrl },
+  };
+}
+
+function oauthDisabledResponse(): Response {
+  return jsonError(
+    "service_unavailable",
+    "OAuth is not configured on this Worker (missing OAUTH_*/STYTCH_* secrets)",
+    503,
+  );
+}
+
+function jsonError(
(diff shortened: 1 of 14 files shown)


workers/src/oauth/handlers.ts
```diff
@@ -0,0 +1,1087 @@
...
+  if (code_challenge_method !== "S256") {
+    return "code_challenge_method must be `S256`";
+  }
+  // Validate `scope`: only values in SUPPORTED_SCOPES are accepted.
+  // Empty / null defaults to "mcp" downstream. This fails-closed instead
+  // of silently echoing unknown scope strings into issued tokens — which
+  // matters once any downstream system starts gating behavior on scope.
+  const scope = p.get("scope");
+  if (scope !== null && scope.length > 0) {
+    const requested = scope.split(/\s+/).filter((s) => s.length > 0);
```
robertabbott, line 440: Nit: `p.get("scope")` returns `""` for `?scope=` (present but empty). The `scope !== null && scope.length > 0` guard skips validation, so `""` flows through to line 570 where `parsed.scope ?? "mcp"` keeps it as `""` — nullish coalescing doesn't trigger on empty string. Should this normalize empty to null (`p.get("scope") || null`) so the `?? "mcp"` default kicks in consistently?

workers/src/oauth/handlers.ts
```diff
@@ -0,0 +1,1087 @@
...
+  if (grant_type === "authorization_code") {
+    return handleAuthorizationCodeGrant(params, cfg);
+  }
+  if (grant_type === "refresh_token") {
+    return handleRefreshGrant(params, cfg);
+  }
+  return jsonError(
+    "unsupported_grant_type",
+    `grant_type \`${grant_type}\` is not supported`,
+    400,
```
robertabbott, line 769: Nit: `grant_type` is `null` when the param is absent, so this interpolates as `` grant_type `null` is not supported ``. Would a separate check for the missing case (`"grant_type is required"`) give clients a more actionable message?

robertabbott (approved):
Clean implementation. Two minor nits below — neither is blocking.

workers/src/oauth/handlers.ts
jfassio: Fixed in 6092433 — normalized to `p.get("scope") || null` so empty string collapses to null and the `?? "mcp"` default applies. Added a test that hits `/authorize` with `?scope=` and asserts the issued token's scope is `mcp`.

workers/src/oauth/handlers.ts
jfassio: Fixed in 6092433 — added a separate null check that returns `invalid_request` with `grant_type is required` (per RFC 6749), so the unsupported-value message is reserved for actual unsupported values. Added a test for the missing-param path.

jfassio:
Addressed 2 review nits in 6092433. @claude-address

PR #55, opened 2026-04-29
MCP Apps widget: handshake, CSP-safe data URI, height fits, connector identity

jfassio:
## Summary

A bundle of small improvements to the worker MCP — primarily fixing the chart-widget rendering pipeline on **claude.ai web** (which had several distinct issues) and tightening the connector's surface in both Claude and ChatGPT clients.

### Connector identity & polish
- **Connector card identity** (`feat(workers): declare connector identity ... serverInfo`): adds `title: "Tako"`, `websiteUrl`, `description`, and an `icons[]` array (SVG light/dark + 180×180 PNG) to the server's `serverInfo`. Honored by clients that support the MCP Apps `Implementation` schema. ChatGPT's dev-mode UI ignores this (uses user-typed connector name), so the title/icon pic only kicks in once the connector is published to the directory.
- **Brand-icon proxy** (`fix(workers): proxy connector icons through /icons/* for stable URLs`): icons live behind a hashed CloudFront archive that rotates per Tako frontend deploy, so direct URLs would rot. New `/icons/*` route on the worker proxies to the upstream archive with a 24h edge cache; the `serverInfo.icons[]` URLs point at our own origin and are stable across deploys. Update one constant in `src/icons.ts` if Tako's archive hash changes.
- **`knowledge_search` description nudge** (`fix(workers): nudge LLM to chain knowledge_search → open_chart_ui`): adds an imperative directive to the tool description so ChatGPT chains into `open_chart_ui` automatically when the user asks to *see* a chart. Without it, ChatGPT was returning text + URL and the user had to ask twice.

### Widget — chart-render pipeline (the long battle)
- **Height handshake listener** (`fix(workers): tighten open_chart_ui default height + add embed resize handshake`): widget now listens for `tako-embed-height` postMessages from the inner Tako embed page so it can self-correct iframe height per chart once the web-app side ships the broadcast (separate PR in the `tako/` repo, not yet merged).
- **MCP Apps `ui/initialize` handshake** (`fix(workers): implement MCP Apps ui/initialize handshake in widget`): the widget never sent the spec-required handshake (`ui/initialize` request → `ui/notifications/initialized` notification). Per the MCP Apps spec (2026-01-26): *\"The Host MUST NOT send any request or notification to the View before it receives an `initialized` notification.\"* claude.ai correctly withheld `tool-result` until the handshake completed; we just never told it we were ready. Subsequently tightened the fallback timeout (2 s → 200 ms) to keep claude.ai's first tool-result delivery from racing the handshake.
- **Cross-origin iframe → static image fallback** (`fix(workers): widget falls back to  when host blocks cross-origin iframe`): claude.ai's outer-document CSP is hardcoded to `frame-src 'self' blob: data:; img-src 'self' blob: data:` and ignores our `_meta.ui.csp.frameDomains` declaration (filed publicly: anthropics/claude-ai-mcp#40, #54). Widget detects via `typeof window.openai` (defined on ChatGPT, undefined on claude.ai) and:
  - **ChatGPT**: keeps the interactive iframe path against `staging.trytako.com/embed/...`.
  - **claude.ai**: switches to a static `` plus a `target=_blank` anchor at `embed_url` so a click opens the fully interactive chart in a new tab.
- **Inline chart PNG as `data:` URI** (`fix(workers): inline chart PNG as data: URI + clickable image fallback`): even the static `` was blocked by `img-src`. Worker fetches the PNG and embeds it as a `data:image/png;base64,...` URI — the only scheme that passes claude.ai's `img-src`. Capped at 250 KB raw to keep the wire envelope small.
- **`_meta` routing for the data URI** (`fix(workers): route inline chart data URI through _meta to dodge LLM context cap`): putting the ~250 KB data URI in `structuredContent` triggered claude.ai's *\"Tool result too large for context, stored at /mnt/user-data/...json\"* guard, which offloads the result to a file and skips widget delivery. New `extraMeta` hook on `ToolModule` attaches the data URI to `result._meta` instead — per the MCP spec, `_meta` is metadata-only and not included in the LLM's context window. Widget reads from `params._meta` in the `ui/notifications/tool-result` postMessage.
- **Pre-size from PNG dims** (`fix(workers): pre-size widget height from PNG dims before image load`): claude.ai reads `documentElement.offsetHeight` once on widget mount and doesn't re-poll (anthropics/claude-ai-mcp#69), so post-load `notifyHeight` calls were ignored and the iframe stayed wrong-sized with internal scrollbar. Worker now parses the PNG's IHDR chunk for natural width/height, ships them via `_meta`, and the widget pre-computes `(clientWidth / naturalWidth) * naturalHeight` and sets the document height *before* the image loads. Post-load adjustment kept as a refinement.

## What didn't work (and is documented in commit messages)
- Restoring the `image` content block in `content[]` next to the widget metadata (commit `1997681`, reverted in `bba0efd`): silently disabled ChatGPT's widget data flow at large response sizes. The `extraMeta` route in `_meta` is the right path because `_meta` isn't tokenized into ChatGPT's context the same way.
- Per-host content-block branching: there's no reliable way to detect host at handler time, so the right fix is the spec-blessed `_meta` channel.

## Out of scope / followups

- **`anthropics/claude-ai-mcp#40` / `#54`** — claude.ai not honoring `_meta.ui.csp.frameDomains` for custom connectors. This PR worked around it on our side; the proper fix is on Anthropic's side.
- **`anthropics/claude-ai-mcp#211`** — claude.ai's *\"Tool result could not be submitted\"* transport error on streamable-HTTP MCP transports. Hits us intermittently; no server-side fix available.
- **Tako web-app embed-page handshake** — once `/embed/{pub_id}` starts broadcasting `tako-embed-height` via postMessage, the widget self-corrects iframe height per chart with no further worker changes. Listener already wired in this PR.
- **Connector publishing** — branding (real Tako logo + name) only kicks in once the connector ships to ChatGPT's app directory and Claude's connector listing.

## Test plan

- [x] `npm run typecheck` passes.
- [x] `npm test` — all 157 tests pass.
- [x] Deploy to staging (`npx wrangler deploy --env staging`).
- [x] Verify connector card icon + title in claude.ai (after disconnect/reconnect).
- [x] Run a single `open_chart_ui` call in claude.ai web; check DevTools console for the full sequence: `ui/initialize sent` → `listener attached` → `ui/notifications/initialized sent` → `rendered { mode: \"img-data\" }` → `img pre-sized from PNG dims`.
- [x] Verify chart renders inline with no \"too large for context\" error and no internal scrollbar.
- [x] Click the chart image → opens fully interactive embed in a new tab.
- [x] Same flow in ChatGPT — confirm the interactive iframe still renders (no regression).
- [x] Smoke check `/icons/favicon.svg` on staging returns 200 with correct MIME.


workers/src/tools/open_chart_ui.ts
```diff
@@ -197,29 +246,136 @@ const WIDGET_HTML = `<!doctype html>
...
+      frame.style.height = h + "px";
+      frame.style.minHeight = h + "px";
+      frame.setAttribute("height", String(h));
+      frame.classList.remove("hidden");
+    } else if (validImage) {
+      // Static fallback. We don't pin a height on \`<img>\` — the image
+      // has its own intrinsic aspect ratio and stretching to a fixed
+      // height would distort the chart. Width-only sizing yields a
+      // slightly different height than the iframe path; that's fine.
+      image.src = imageSrc;
```
robertabbott: **Broken image dead-end on claude.ai when data URI isn't available**

When `fetchImageDataUrlAndDims` returns `undefined` (PNG > 250 KB, fetch timeout, or non-PNG content type), the widget falls through to the `validHttpImage` path and sets `image.src` to the cross-origin `image_url`. On claude.ai, whose CSP enforces `img-src 'self' blob: data:`, this image is blocked — the user sees a broken `` (or the alt text "Tako chart") with no way to reach the interactive chart.

The `else if (validEmbed)` branch (iframe fallback) is unreachable in this case because `validImage` is `true` — the HTTP URL passes validation even though CSP will block it at render time.

Would adding an `image.addEventListener("error", ...)` handler that shows the placeholder with the existing click-through `<a href=embed_url>` link close this gap? That would make the degradation chain: data URI → CSP-blocked image → clickable "Open chart" link, rather than a dead-end broken image.

workers/src/tools/open_chart_ui.ts
```diff
@@ -197,29 +246,136 @@ const WIDGET_HTML = `<!doctype html>
...
+      frame.setAttribute("height", String(h));
+      frame.classList.remove("hidden");
+    } else {
+      // Nothing usable; leave the placeholder visible.
+      return false;
+    }
+
     placeholder.classList.add("hidden");
     rendered = true;
     notifyHeight(h);
```
robertabbott, line 545: **`notifyHeight(h)` overwrites pre-sized height; log reports wrong height**

After the pre-sizing block (lines 327–337) computes `renderedH` from the PNG's natural dimensions and calls `notifyHeight(renderedH)`, the unconditional `notifyHeight(h)` here overwrites it with the default structuredContent height (720). Today `notifyHeight` is a no-op on claude.ai (no `window.openai`), so this is harmless in production — but if a future MCP host exposes `notifyIntrinsicHeight` while also restricting iframe rendering, the overwrite becomes a real sizing bug.

More immediately: the log at lines 370–378 always reports `height: h` (720) rather than the actual rendered height for the image path, which would mislead debugging when investigating height mismatches.

Should the effective height be tracked across branches (e.g., `var effectiveH = h;` at the top, updated to `renderedH` in the pre-sizing block) and used for both the final `notifyHeight` and the log?

workers/src/tools/open_chart_ui.ts
jfassio: Fixed in b605eb1. Added an `image.addEventListener("error", …)` handler in the `validImage` branch that:

- Hides the broken `` and the (would-have-shown) `imageLink` wrapper.
- Repurposes the `#tako-placeholder` element: replaces its "Loading chart…" text with a styled `<a target="_blank" href={embed_url}>Open interactive chart →</a>` (when `validEmbed` is true).
- Falls back to "Couldn't load chart." when there's no embed URL either (defensive — current schema guarantees one, but matches the surrounding pattern).
- Logs `img errored, showing click-through fallback` so this is grep-able in DevTools when investigating future "no chart visible" reports.

Degradation chain is now: data URI → CSP-blocked HTTP image → click-through "Open interactive chart" link, as you suggested. No more dead-end broken image on claude.ai when the PNG is over the 250 KB inline cap.

workers/src/tools/open_chart_ui.ts
jfassio: Good catch on the underlying concern, but the specific bug you flagged actually doesn't reproduce in the current file — checked carefully. Two things changed since the commit you reviewed against:

1. The pre-sizing-from-PNG-dims block (the prior `lines 327–337` you reference) was removed in a follow-up commit, replaced with the `image.load`-atomic measurement at lines 456–480 (`void imgNaturalW; void imgNaturalH;` is the breadcrumb). Comment on lines 443–446 explains why: pre-sizing from PNG natural dims caused claude.ai to lock the outer iframe at the wrong height. So `notifyHeight(renderedH)` is no longer called from a pre-size block.

2. The `validImage` branch now returns early at line 493 (`return true;`) immediately after assigning `image.src = imageSrc;`. The unconditional `notifyHeight(h)` at the bottom is only reached by the iframe paths, where `h` matches the height we just pinned on `frame.style.height`. So no overwrite, and the bottom log's `height: h` is accurate for the paths that actually reach it.

That said — your observation about the log having branches that referenced the image path was correct, those were unreachable. Cleaned up in b605eb1: dropped `validImage`/`validDataImage` from the bottom log's `mode`/`src` ternaries (they were dead since the early return) and added a comment noting that the tail block only runs for the iframe paths. Image path logging stays in the `image.load` and (new) `image.error` listeners.

jfassio:
Addressed 2 comments from this review round in b605eb1. @claude-address

robertabbott (approved):

PR #59, opened 2026-04-30
feat(workers): knowledge_search auto-renders top chart inline

jfassio:
## Summary

- `knowledge_search` now ships the chart widget for the top result as part of its own `tools/call` response. ChatGPT renders the chart inline on the first call instead of forcing the user to ask twice ("give me the chart" → link → "give me the chart" → card). claude.ai falls through to its existing inline-PNG path via the `widgetSuppressed` rule.
- Same `ui://tako/embed/chart` widget URI as `open_chart_ui` — both tools share one bundle. The widget HTML, PNG fetch helpers, URL builders, and the `AppUiResource` factory are extracted to `workers/src/tools/_chart_widget.ts` so the two tools don't drift the next time a host bug needs a fix.
- `mcp.ts` dedupes `registerResource` / `ResourceTemplate` registrations across tools sharing a widget URI (the SDK throws on duplicates), and threads the handler's `output` into `dynamic.resolveUriFromInput` so `knowledge_search` can derive the per-call URI from `results[0].card_id`.
- Descriptions rewritten: `knowledge_search` drops the "MUST chain into `open_chart_ui`" paragraph and explains the auto-render plus multi-card narration. `open_chart_ui` reframed as the explicit re-render tool for `create_chart` chains, non-top results, and pub_id references from prior turns.

## Behavior matrix

| Host | Before | After |
| --- | --- | --- |
| ChatGPT | `knowledge_search` returned data → link → user asked again → `open_chart_ui` rendered widget | `knowledge_search` returns data + widget in one call |
| claude.ai | Same two-turn pattern | `knowledge_search` returns data + link to interactive chart |
| Other MCP-Apps hosts | Two-turn pattern | Same widget bundle delivered on first call |
| Plain MCP clients | Text-only fallback | Text + structured content with `embed_url` for the model to surface as a link |

## Test plan

- [x] `npx vitest run` — 180/180 pass (12 new tests added in `knowledge_search.test.ts` covering the auto-chain output shape, `extraMeta` data URL + dimensions, `extraContentBlocks` image block, and the dynamic URI resolver reading from output)
- [x] `npm run typecheck` — clean (workers + scripts)
- [x] `npm run registry:check` — in sync (9 tools)
- [x] Manual smoke on staging connector against ChatGPT: ask "what's AAPL stock price today, give me a tako chart" — expect chart card inline on first turn, no second tool call
- [x] Manual smoke on staging connector against claude.ai: same query — expect inline PNG content block on first turn
- [ ] Verify follow-up "show me [non-top result title]" still routes through `open_chart_ui` correctly

Screenshot 2026-04-30 at 15 47 48


workers/src/mcp.ts
```diff
@@ -440,7 +485,15 @@ function registerTool(
       // image-baked dynamic variant).
```
robertabbott, line 496: **`extraMeta` runs unconditionally even on widget-suppressed hosts.**

The `extraContentBlocks` call (around line 442) is gated on `ui === undefined` (skip when the widget is registered). But the `extraMeta` call just above this hunk (around line 467) has no such gate — it fires for every tool call. Both `knowledge_search` and `open_chart_ui` use `extraMeta` exclusively to ship `image_data_url` for the widget to read via `params._meta`. On claude.ai (widget suppressed → `ui === undefined`), `extraMeta` still fetches the PNG and base64-encodes it, but no widget will consume it. Net effect on a chart-bearing tool call to claude.ai: TWO PNG fetches for the same image (one in `extraMeta`, one in `extraContentBlocks`) and a ~330KB unused data URL inflating the JSON-RPC response. Should `extraMeta` be gated on `ui !== undefined` so it mirrors the inverse rule for `extraContentBlocks`?

workers/src/tools/open_chart_ui.ts
```diff
@@ -1,43 +1,42 @@
...
+ * a fresh `card_id`, or when the user asks to chart a non-top result
+ * from a previous `knowledge_search` call. For first-time chart
+ * requests, prefer `knowledge_search`: it auto-renders the top result
+ * inline as part of the same tool call, so the model doesn't need to
+ * chain.
  *
  * Subrequest cost: one extra outbound `fetch` to the PNG endpoint per
- * tool call. Within Workers' 50/1000 cap with plenty of headroom.
+ * tool call (via `extraMeta`'s `fetchImageDataUrlAndDims` and, on
+ * widget-suppressed hosts, `extraContentBlocks`'s `fetchPngContentBlock`
```
robertabbott: **Comment claims `extraMeta` and `extraContentBlocks` are mutually exclusive — they aren't.**

This header comment says: "one extra outbound `fetch` to the PNG endpoint per tool call (via `extraMeta`'s `fetchImageDataUrlAndDims` and, on widget-suppressed hosts, `extraContentBlocks`'s `fetchPngContentBlock` — they're mutually exclusive)". Looking at `mcp.ts`, `extraContentBlocks` is gated on `ui === undefined` but `extraMeta` runs unconditionally. So on widget-suppressed hosts both fire, doubling the PNG fetch. Either the comment should say "they overlap on widget-suppressed hosts" or the wiring in `mcp.ts` should match the comment (gate `extraMeta` on `ui !== undefined`). Which is the intended invariant?

workers/src/tools/knowledge_search.ts
```diff
@@ -389,8 +425,86 @@ const knowledge_search = {
...
+    // Inline the top chart's PNG as a `data:` URI on `_meta` (kept
+    // off the LLM's context window per the MCP Apps spec) plus the
+    // source PNG's natural pixel dimensions. The widget reads these
+    // for the image-baked render path on hosts whose CSP rejects
+    // cross-origin imgs (claude.ai). When the handler didn't populate
+    // `image_url` (no top card / no card_id), we have nothing to
+    // fetch and skip silently.
+    void _ctx;
+    if (output.image_url === undefined) return undefined;
+    const fetched = await fetchImageDataUrlAndDims(output.image_url);
```
robertabbott, line 469: **Subrequest math comment is now stale.**

The sizing comment at lines 74-82 (above this hunk) says `2 + 29 + 2 = 33 subrequests — under the cap with ~17 slots of headroom`. With this PR the handler now also auto-chains a chart, so the new `extraMeta` hook here adds one more outbound `fetch` per call on widget-enabled hosts; on widget-suppressed hosts there are two more (`extraMeta` + `extraContentBlocks`, per the duplicate-fetch comment on `mcp.ts`). For an explicit deep search on claude.ai that's 33 + 2 = 35. Worth bumping the comment so a future tuner doesn't trip the 50-cap, or is the headroom assertion still considered safe enough to leave as-is?

workers/src/tools/knowledge_search.ts
```diff
@@ -389,8 +425,86 @@ const knowledge_search = {
...
+      // `knowledge_search`'s pub_id is derived from the handler's
+      // output (`output.pub_id`, set when the top card has a
+      // card_id), NOT from input — its input is a search query.
+      // Pre-handler-result calls (and tool calls that produced no
+      // top card) fall back to the static URI so a registered
+      // resource always answers; the widget itself handles the
+      // no-chart case.
+      void _input;
+      const pubId =
+        typeof (output as { pub_id?: unknown } | undefined)?.pub_id === "string"
```
robertabbott, line 501: **`output` triple-cast — could a typed resolver remove the cast noise?**

This resolver does `typeof (output as { pub_id?: unknown } | undefined)?.pub_id === "string" ? (output as { pub_id: string }).pub_id : ""`. The reason is the type-erasure at `AnyToolModule` — `resolveUriFromInput`'s parameters are `unknown` at the registry boundary. But the resolver lives in the same file as the typed handler whose Output shape it knows. Would making `dynamic.resolveUriFromInput` generic over the tool's Output type (with the erasure happening only at the registry) make this safer to evolve, or is duplicating the cast across two tools tolerable since each ships only one resolver?

registry/server.json
```diff
@@ -205,10 +205,10 @@
         }
       },
       "annotations": {
-        "title": "Tako: Search Charts",
+        "title": "Tako: Live Data & Charts",
         "readOnlyHint": true,
         "destructiveHint": false,
-        "openWorldHint": false
+        "openWorldHint": true
```
robertabbott: **`openWorldHint` flipped from `false` to `true` for `knowledge_search`.**

The annotation flipped in this PR. The hint definition is roughly "tool's effects are observable outside Tako (e.g. creates a public URL)". `knowledge_search` doesn't create new public URLs — it returns existing chart references. The auto-chain widget surfaces those existing URLs inline, but they were already public and previously surfaced via a follow-up `open_chart_ui` (which is `true`). Was the flip intentional, and if so, is there a host that gates behavior on this flag?

workers/src/tools/_chart_widget.ts
```diff
@@ -0,0 +1,994 @@
...
+    // Reject anything that's not an image — an upstream redirect to an
+    // HTML error page would otherwise let us base64 HTML and ship it
+    // as a `data:image/...` URI the client can't render.
+    if (!contentType.startsWith("image/")) return undefined;
+    const buffer = await response.arrayBuffer();
+    if (buffer.byteLength === 0) return undefined;
+    if (buffer.byteLength > MAX_INLINE_DATA_URL_BYTES) return undefined;
+    const dims = parsePngDimensions(buffer);
+    if (dims === undefined) return undefined;
+    const mime = contentType.split(";")[0]!.trim();
```
robertabbott: **`mime` derivation looks like dead code after the PNG-only filter.**

`fetchImageDataUrlAndDims` returns `undefined` whenever `parsePngDimensions` fails (line 836), which guarantees the buffer is a PNG by the time line 837 runs. So `const mime = contentType.split(";")[0]!.trim()` always equals `"image/png"`. The MIME derivation looks like residue from a path that supported other image types. Is this intentional headroom for a future format, or can it just be `mimeType: "image/png"` to drop a few lines?

robertabbott (commented):
Six observations from a code-level read — all minor, none blockers. The biggest one is #1 (duplicate PNG fetch on widget-suppressed hosts).

workers/src/tools/knowledge_search.test.ts
```diff
@@ -577,3 +578,334 @@ describe("knowledge_search async-task polling (TAKO-2686)", () => {
...
+      "https://staging.trytako.com/api/v1/image/aapl-price/?dark_mode=true",
+    );
+    // Defaults match `open_chart_ui` so a follow-up explicit render
+    // produces a visually identical chart.
+    expect(out.dark_mode).toBe(true);
+    expect(out.width).toBe(900);
+    expect(out.height).toBe(720);
+  });
+
+  it("uses the FIRST card with a card_id when multiple are returned (top-only auto-render)", async () => {
```
robertabbott: **Test title says "first card WITH a card_id" — but the handler picks the first card unconditionally.**

This test is titled `uses the FIRST card with a card_id when multiple are returned (top-only auto-render)`, and the fixture has `card_id` set on `results[0]`. But the handler at `knowledge_search.ts:436` does `const topCardId = results[0]?.card_id` — it picks the first result period, then bails out of auto-chain if its `card_id` is null. So if a real response had `results[0].card_id = null` and `results[1].card_id = "abc"`, no auto-chain would happen even though there's a chart-bearing card available. Is the test title wrong, or is the intent to skip null `card_id` results and pick the first chart-bearing one? If the latter, this test isn't actually exercising that path — both fixture cards have card_ids.

workers/src/mcp.ts
jfassio: Fixed in 7d95acd — gated `extraMeta` on `ui !== undefined`, mirroring the inverse rule for `extraContentBlocks`. On widget-suppressed hosts (claude.ai) the worker now does exactly one PNG fetch (via `extraContentBlocks` → `fetchPngContentBlock`) and the ~330 KB unused `image_data_url` no longer ships in the JSON-RPC envelope. Also added a regression test in `index.test.ts` that calls `open_chart_ui` with a `claude` User-Agent and asserts `_meta.image_data_url` is absent.

workers/src/tools/open_chart_ui.ts
jfassio: Wiring fixed (the comment is now accurate). With `extraMeta` gated on `ui !== undefined` and `extraContentBlocks` gated on `ui === undefined`, each call hits exactly one of the two PNG-fetch hooks. Tightened the wording in this header to spell out the gate explicitly so the invariant is harder to drift away from.

workers/src/tools/knowledge_search.ts
jfassio: Bumped the sizing comment to `2 + 29 + 2 + 1 = 34` subrequests with the auto-chain PNG fetch counted explicitly. After the gating fix, only one of `extraMeta` / `extraContentBlocks` fires per call, so the +1 is constant across hosts (not +2 on suppressed). Headroom drops from ~17 to ~16 slots — still safely under the 50 cap.

workers/src/tools/knowledge_search.ts
jfassio: Leaving as-is for this round. The cast lives in only two tools (`open_chart_ui` and `knowledge_search`), each carries it next to the typed handler that knows the Output shape, and the `void _input` / `void _ctx` style there already documents which arg the resolver actually reads. Making `dynamic.resolveUriFromInput` generic over the tool Output would cascade through `AppUiResource`, `ToolModule`, and `AnyToolModule` (with the erasure pushed to the registry boundary) — happy to do that as its own PR if a third widget-bearing tool lands and the duplication starts to bite, but at N=2 the local cast feels under the threshold.

registry/server.json
jfassio: Reverted to `false` in 7d95acd. The flip was unintentional drift in the auto-chain commit — `knowledge_search` returns existing chart references rather than creating new public URLs, so `false` is the right value. `registry/server.json` regenerated to match. (`open_chart_ui` keeps `true` since it is the explicit "render this public URL" tool.)

workers/src/tools/_chart_widget.ts
jfassio: Fixed in 7d95acd — replaced the `mime` derivation with a hardcoded `"image/png"` data-URI prefix, since `parsePngDimensions` upstream already validates the PNG signature (89 50 4E 47…) and bails to `undefined` on anything else. Left the parallel split in `fetchPngContentBlock` alone — that path only checks `image/*` content-type (no IHDR gate), so the upstream MIME is the right source of truth there.

workers/src/tools/knowledge_search.test.ts
jfassio: Renamed the test in 7d95acd to `"uses results[0] unconditionally when multiple cards are returned (top-only auto-render)"` — that matches the actual handler behavior. The skip-null-card_id case you described is intentionally covered by the next test ("omits chart fields when no card has a card_id"), which exercises the `results[0].card_id === null` path and asserts no chart fields land in the output. The design is "top-only": if `results[0]` has no card_id, no auto-chain happens and the user can ask for a non-top result via `open_chart_ui`.

jfassio:
Addressed all 7 review comments in 7d95acd — gated `extraMeta` on `ui !== undefined` (the headline fix), reverted `openWorldHint` to `false`, dropped the dead `mime` derivation, renamed the misleading test title, and updated the stale subrequest math + mutually-exclusive comments. New regression test in `index.test.ts` locks in the gating. 181/181 tests pass; typecheck + registry:check clean. @claude-address

robertabbott (approved):

PR #61, opened 2026-04-30
feat(workers): export_report tool for MD/JSON/PDF/PPTX exports

jfassio:
## Summary

- New `export_report` MCP tool that proxies `GET /api/v1/internal/reports/{id}/export/{slug}/` (the same endpoint the Tako web Export menu hits) and returns the file inline: text content for `markdown` / `json`, base64 for `pdf` / `powerpoint`.
- Format → URL slug map: `markdown→markdown`, `json→json`, `pdf→pdf`, `powerpoint→pptx`. PDF and JSON slugs confirmed from staging web UI's network panel; markdown/pptx are conventional and surface a clear "format slug may not be supported" error if they ever 404.
- 4 MiB hard cap on response size (oversize → "download from web UI" error pointing at `webpage_url`); 60 s fetch timeout. 404 responses surface a `wait_for_report` hint so the LLM does not retry on a still-running report.

## Test plan

- [x] `npx vitest run` — 191/191 pass (10 new tests in `export_report.test.ts`)
- [x] `npx tsx scripts/gen-registry.ts --check` — no drift
- [x] Deployed to staging (`mcp.staging.tako.com`, version `c90754cc-b181-4396-a941-0917722e6880`)
- [x] Verify staging: PDF export end-to-end via Tako stg MCP client
- [x] Verify staging: JSON / Markdown / PowerPoint exports each return the expected content type and shape
- [x] Confirm the `pptx` slug for PowerPoint is correct (will surface a clear error if not — flip to `powerpoint` if needed)
Screenshot 2026-05-01 at 14 37 01


registry/server.json
```diff
@@ -125,6 +125,28 @@
...
+      "name": "export_report",
+      "description": "Use this when the user explicitly asks to export, download, or save a completed Tako report in a specific file format (Markdown, JSON, PDF, or PowerPoint). Returns a short-lived `download_url` (default: 5 minutes) — surface it to the user as a markdown link, e.g. `[Download as PDF](download_url)`, and tell them how long they have to click before it expires (`expires_in_seconds`). The report MUST be `status === \"completed\"`; if the click 404s with \"report not ready,\" call `wait_for_report` first and re-mint a fresh URL. Do NOT call this proactively after `create_report` or `get_report` — only when the user actually asks for an export. If the user just wants to read the report in their browser, share the report's `webpage_url` instead — no export needed.",
+      "parameters": {
+        "report_id": {
+          "type": "string",
+          "description": "Report ID returned from create_report or list_reports.",
+          "required": true
+        },
+        "format": {
+          "type": "string",
```
robertabbott, line 138: `format` parameter missing enum constraint — The Zod inputSchema in `export_report.ts` restricts this to `["markdown", "json", "pdf", "powerpoint"]`, but the registry entry is just `"type": "string"`. LLMs that parse the registry schema (rather than the Zod schema) won't know the valid values, increasing the chance of invalid calls (e.g., passing `"pptx"` instead of `"powerpoint"`). Should this add `"enum": ["markdown", "json", "pdf", "powerpoint"]` to match the other enum-constrained parameters in the registry?

workers/src/exports.ts
```diff
@@ -0,0 +1,352 @@
...
+      upstream.status,
+      `upstream returned ${upstream.status}: ${text || "(empty)"}`,
+    );
+  }
+
+  // Mirror the upstream content-type when present so the browser
+  // picks the right preview/download path. Filename uses the format
+  // extension; the report id is included for "which report did I
+  // download" disambiguation when the user has many.
+  const filename = `tako-report-${payload.rid}.${FORMAT_EXTENSION[payload.fmt]}`;
```
robertabbott: `payload.rid` flows into the `Content-Disposition` filename unescaped — The filename is built as `tako-report-${payload.rid}.${ext}`. The `rid` comes from the encrypted token (so an attacker can't inject arbitrary values without the key), but a legitimate report ID containing `"`, `\`, or newlines would break the `filename="..."` quoting or — in some user-agents — open a response-splitting vector. `encodeURIComponent` on line 259 protects the URL path but not the header. Report IDs are currently UUID-shaped so this is safe today, but would a defensive sanitize (strip non-alphanumeric-dash-underscore) before interpolating into the header be worth the insurance?

workers/src/exports.ts
```diff
@@ -0,0 +1,352 @@
...
+    return "";
+  }
+}
+
+/** Strict base64 (no URL-safe alphabet) — used for the AES key in env. */
+function base64Decode(s: string): Uint8Array {
+  return new Uint8Array(Buffer.from(s, "base64"));
+}
+
+function base64UrlEncode(data: Uint8Array): string {
```
robertabbott: Duplicated base64url + AES-GCM primitives — `oauth/jwt.ts` already exports `b64url()`, `b64urlDecode()`, `importAesKey()`, `encryptAesGcm()`, and `decryptAesGcm()` with the same AES-256-GCM + base64url pattern reimplemented here (`base64UrlEncode`, `base64UrlDecode`, `loadKey`, and the inline encrypt/decrypt blocks). The `jwt.ts` versions use `btoa`/`atob` (Workers-native), while these use `Buffer` (Node compat layer). Would it make sense to extract the shared primitives into a common crypto module, or is there a deliberate reason to keep the two implementations separate (e.g., wanting export tokens independent of the OAuth module's evolution)?

workers/src/exports.ts
```diff
@@ -0,0 +1,352 @@
...
+    typeof payload.rid !== "string" ||
+    payload.rid === "" ||
+    typeof payload.key !== "string" ||
+    payload.key === "" ||
+    typeof payload.exp !== "number" ||
+    !EXPORT_FORMATS.includes(payload.fmt as ExportFormat)
+  ) {
+    throw new Error("Token payload is malformed");
+  }
+  if (payload.exp <= nowSeconds) {
```
robertabbott: `exp <= nowSeconds` rejects tokens checked at exactly the expiry second — `mintExportToken` sets `exp = floor(now) + ttlSeconds`, but `verifyExportToken` rejects when `exp <= nowSeconds`, so a token verified at exactly `exp` fails. The effective TTL is `ttlSeconds - 1`. The tests use `expiresAt + 1` to assert expiry, which passes, but a check at exactly `expiresAt` would also reject. Is the off-by-one intentional (erring conservative), or should this be `exp < nowSeconds` to honor the full 300s?

workers/src/tools/export_report.ts
```diff
@@ -0,0 +1,110 @@
...
+    title: "Tako: Export Report",
+    readOnlyHint: true,
+    destructiveHint: false,
+    // The download URL is publicly resolvable for its TTL — set
+    // `openWorldHint: true` to flag that a side-effect-free token is
+    // entering circulation.
+    openWorldHint: true,
+  },
+  async handler(input, ctx) {
+    if (ctx.env.EXPORT_TOKEN_KEY === undefined || ctx.env.EXPORT_TOKEN_KEY === "") {
```
robertabbott: The handler pre-checks `EXPORT_TOKEN_KEY` presence and throws a nice operator-facing error mentioning `wrangler secret put`. But if the key is present with the wrong byte-length, `loadKey` (called inside `mintExportToken`) throws a generic `"must decode to 32 bytes"` error that doesn't include the operator-friendly fix hint. Since the handler is already pre-validating, should it also validate the key length up front so the actionable message covers that failure mode too?

workers/src/exports.ts
```diff
@@ -0,0 +1,352 @@
...
+      `upstream fetch failed: ${err instanceof Error ? err.message : "unknown"}`,
+    );
+  }
+
+  if (!upstream.ok) {
+    // Forward the upstream status so the user (and Worker logs) can
+    // tell "report not done" (404) from "auth bad" (401/403) from
+    // "render failed" (5xx).
+    const text = await safeReadText(upstream);
+    return textResponse(
```
robertabbott: Non-2xx upstream responses are returned as raw plaintext to the browser — For a 404 (report not ready), the user clicking the download link sees `upstream returned 404: {"error": "Report not found"}` in a bare browser tab with no chat context. The tool description tells the LLM to advise `wait_for_report`, but the person clicking has no LLM in the loop. Would it be worth returning a minimal HTML page with a human-readable message (e.g., "This report isn't ready yet — go back to your chat and ask the assistant to try again") instead of a raw error string?

workers/src/exports.ts
jfassio: Fixed in c11c12a — flipped `<=` to `<` so a token verified at exactly `exp` is still valid (the full advertised 300s is honored). New test `accepts a token verified at exactly its expiry second` locks the boundary so a future refactor can't silently re-introduce the off-by-one. The existing `expiresAt + 1` test still rejects since `expiresAt < expiresAt + 1`.

registry/server.json
jfassio: Done in c11c12a — updated `flattenParameters` in `gen-registry.ts` to forward `enum` from the Zod-emitted JSON Schema, and re-ran the codegen.

One thing worth flagging: your premise was slightly off — there were no other enum-constrained params in the registry. The codegen was actively stripping them (the docstring even said so: _"enums… is dropped because the external registry's parameter format does not carry them"_). Fixing the codegen restored enums on three params at once: `export_report.format`, `knowledge_search.search_effort` (`fast`/`medium`/`deep`/`auto`), and `list_reports.status` (`pending`/`running`/`completed`/`failed`). Those last two were silently dropping their constraints too.

workers/src/exports.ts
jfassio: Done in c11c12a — added a `sanitizeFilenameComponent` helper that restricts to `[A-Za-z0-9_-]`, replacing anything outside that set with `_`. UUID-shaped report IDs pass through unchanged, but a future schema change that produces a non-UUID `rid` (or a yet-to-exist mint bug that lets a tampered value through) can't inject `"`/`\`/CRLF into the `Content-Disposition` header. New test `sanitizes the report id in the Content-Disposition filename` exercises a crafted rid containing quote, newline, and equals to lock the contract.

workers/src/exports.ts
jfassio: Partial dedup in c11c12a — replaced `exports.ts`'s inline `base64UrlEncode`/`base64UrlDecode`/`base64Decode` (which used the Node `Buffer` compat layer) with imports of `b64url`/`b64urlDecode` from `oauth/jwt.ts` (Workers-native `btoa`/`atob`). The two modules can no longer drift on encoding rules.

Kept the AES-GCM mint/verify logic local instead of using `encryptAesGcm`/`decryptAesGcm` from jwt.ts: the OAuth helpers return `null` on failure (the "treat as missing claim" semantic the OAuth flow needs), whereas the route handler here needs to throw to distinguish tampered/expired/malformed in `console.warn` for operator triage. Wrapping `decryptAesGcm` to re-throw worked fine but didn't obviously beat the inline `try { decrypt } catch { throw new Error("Token authentication failed") }` pattern, so I left the AES path alone.

Happy to extract everything into a shared `workers/src/crypto.ts` if you'd rather — just say the word and I'll do it as a follow-up.

workers/src/tools/export_report.ts
jfassio: Done in c11c12a — extracted `assertExportTokenKeyConfigured(env)` in `exports.ts` that pre-validates presence + base64 decodability + 32-byte length, throwing one operator-friendly message that always mentions `EXPORT_TOKEN_KEY` and `wrangler secret put` regardless of which failure mode was hit. The handler calls the helper instead of inlining the presence check.

`loadKey` keeps its terser internal-only message — it's only reachable from inside `mintExportToken`/`verifyExportToken`, which the handler already pre-validates before calling. New test `throws a clear operator-targeted error when EXPORT_TOKEN_KEY is the wrong byte length` covers the previously-uncaught case (asserts the message includes both `EXPORT_TOKEN_KEY` and `wrangler secret put` plus the actual byte count).

workers/src/exports.ts
jfassio: Done in c11c12a — added a small inline-styled `htmlErrorResponse(status, title, message)` helper. Used it for:

- 401 invalid/expired token → "Download link expired"
- 404 upstream not-ready → "Report not ready yet — Go back to your chat and ask the assistant to try the export again in a moment."
- Other upstream non-2xx (5xx, 401/403 from Django, etc.) → "Couldn't generate the export"
- 502 fetch-threw → "Couldn't reach the export service"
- 500 misconfig → "Service unavailable" with a "contact support" hint

Plain-text 400 stays for the empty-token path (only reachable via hand-edited URL, no real user gets there). Upstream raw bodies still go to `console.warn` with `rid` + `fmt` for operator triage; they just don't leak into the user-facing page anymore.

Three new tests pin the contract: `returns a friendly HTML page (401) for an invalid token`, `(404) when upstream reports the report isn't ready` (also asserts the upstream raw error doesn't bleed through), and `for upstream 5xx`.

jfassio:
Addressed all 6 review comments from @robertabbott in c11c12a — replies posted on each thread. Highlights: off-by-one in token expiry, filename sanitization, front-loaded `EXPORT_TOKEN_KEY` validation, friendly HTML error pages for browser-clicked URLs, base64url helper dedup against `oauth/jwt.ts`, and registry codegen now forwards `enum` (which incidentally restored constraints on `knowledge_search.search_effort` and `list_reports.status` too). 206/206 tests pass.

@claude-address

workers/src/exports.ts
```diff
@@ -0,0 +1,491 @@
...
+  try {
+    payload = await verifyExportToken(token, env);
+  } catch {
+    // Single generic message — distinguishing expired / tampered /
+    // malformed would help an attacker fingerprint the validator.
+    return htmlErrorResponse(
+      401,
+      "Download link expired",
+      "This download link is no longer valid — it may have expired or already been used.",
+    );
```
robertabbott, line 324: The catch swallows all `verifyExportToken` failures — correct for the user-facing response (don't help an attacker fingerprint the validator), but there's no server-side log entry. If `EXPORT_TOKEN_KEY` is accidentally rotated or removed from a running deployment, `loadKey` throws a config error and the route silently returns "Download link expired" with no operator signal. Would a `console.warn` with the caught error message before the generic 401 close that observability gap? The response stays generic; only the operator log gets the real reason — and the volume is low since download URLs are only minted on explicit export requests.

workers/src/tools/export_report.ts
```diff
@@ -0,0 +1,112 @@
...
+      ctx.env,
+    );
+
+    return {
+      report_id: input.report_id,
+      format: input.format,
+      // Token already URL-safe (base64url), so no extra encoding here.
+      download_url: `${base}/exports/${token}`,
+      expires_at: expiresAt,
+      expires_in_seconds: DEFAULT_TOKEN_TTL_SECONDS,
```
robertabbott: Minor: `expires_in_seconds` returns the constant `DEFAULT_TOKEN_TTL_SECONDS` (300) rather than deriving from the actual mint result (`expiresAt - Math.floor(Date.now() / 1000)`). Today these are equivalent since the tool never passes a custom TTL, but `expires_at` is the authoritative value from `mintExportToken` while this is a fixed constant that happens to match. If the TTL is ever customized per-format (e.g. larger files get a longer click window), the LLM would tell the user "you have 5 minutes" when the actual window differs. Nit-level — fine to leave as-is if per-format TTLs aren't on the horizon.

robertabbott (approved):
Clean implementation — the security model (AEAD tokens, tight TTL, streaming proxy) is sound, and the fix commit addressed all prior feedback well. Two minor observations below, neither blocking.

workers/src/exports.ts
jfassio: Good catch — added a `console.warn` in the catch with the underlying error message before the generic 401 in 72f35b1. User-facing page stays identical (no fingerprinting), but a rotated/removed `EXPORT_TOKEN_KEY` will now show up in operator logs as `[exports] verifyExportToken failed: …`.

jfassio: Good catch — added a `console.warn` in the catch with the underlying error message before the generic 401 in 72f35b1. User-facing page stays identical (no fingerprinting), but a rotated/removed `EXPORT_TOKEN_KEY` will now show up in operator logs as `[exports] verifyExportToken failed: …`.

jfassio: Good catch — added a `console.warn` in the catch with the underlying error message before the generic 401 in 72f35b1. User-facing page stays identical (no fingerprinting), but a rotated/removed `EXPORT_TOKEN_KEY` will now show up in operator logs as `[exports] verifyExportToken failed: …`.

jfassio: Good catch — added a `console.warn` in the catch with the underlying error message before the generic 401 in 72f35b1. User-facing page stays identical (no fingerprinting), but a rotated/removed `EXPORT_TOKEN_KEY` will now show up in operator logs as `[exports] verifyExportToken failed: …`.

workers/src/tools/export_report.ts
jfassio: Fixed in 72f35b1 — `expires_in_seconds` now derives from `expiresAt - Math.floor(Date.now() / 1000)` so it reflects the actual mint result. Same value today, but won't lie if per-format TTLs ever land. Dropped the now-unused `DEFAULT_TOKEN_TTL_SECONDS` import.

jfassio:
Addressed both follow-up observations from the approval review in 72f35b1 — replies posted on each thread. `exports.ts` now logs verifyExportToken failures via `console.warn` (user response stays generic), and `export_report.ts` derives `expires_in_seconds` from the actual mint result instead of the constant. 206/206 tests pass. @claude-address

PR #66, opened 2026-05-04
feat(oauth): single-use auth-code + refresh-token enforcement (TAKO-2701)

jfassio:
## Summary
- Closes the OAuth 2.1 §4.1.2 single-use gap on `/token`: auth codes and refresh tokens now carry a `jti` claim, and the redemption handler records the `jti` in Workers Cache on first exchange so any subsequent presentation short-circuits with `invalid_grant`.
- Same fix applied symmetrically to refresh-token rotation (§4.3.1) per the ticket. Two new behavioral tests assert that the second redemption of either grant returns 400.
- Linear: [TAKO-2701](https://linear.app/trytako/issue/TAKO-2701/single-use-auth-code-enforcement-workers-cache-backed-jti). Follow-up to TAKO-2679; replaces the two `KNOWN GAP` TODOs in `handlers.ts`.

## Caveats (deliberate, documented at the helper)
- **Per-colo, not global.** Workers Cache is per-edge. Cross-colo replay would bypass enforcement, but consumer MCP hosts (Claude.ai, ChatGPT) are sticky to one colo per session, so this catches the realistic threat. Cross-colo replay would itself be a strong signal warranting KV.
- **Best-effort LRU.** The 60s auth-code window is hard to evict; the 14-day refresh window is plausibly evictable under memory pressure. Refresh-replay protection is correspondingly weaker than auth-code protection — KV-backed enforcement remains the path forward if/when refresh-replay needs to be hard.

## Files
| File | Change |
|---|---|
| `workers/src/oauth/types.ts` | `jti: string` added to `AuthCodeClaims` and `RefreshTokenClaims` |
| `workers/src/oauth/handlers.ts` | `crypto.randomUUID()` at both mint sites; new `checkAndMarkRedeemed()` helper; wired into both redemption paths; old TODO comments removed |
| `workers/src/oauth/handlers.test.ts` | Two new tests: replay of redeemed auth code → 400; replay of redeemed refresh token → 400 |

Net: +161 / −14 across 3 files.

## Test plan
- [x] `npm test` green (183/183 — `handlers.test.ts` now 46 tests)
- [x] `npm run registry:check` ok
- [x] CI green
- [ ] Post-deploy: redeem an auth code twice in staging, second attempt should return `{"error":"invalid_grant","error_description":"authorization code already redeemed"}`

### DIFF
--- a/workers/src/oauth/handlers.ts
+++ b/workers/src/oauth/handlers.ts
@@ -791,12 +791,26 @@ export async function handleToken(req: Request, env: Env): Promise<Response> {
  *   plausible under memory pressure — refresh-replay protection is
  *   correspondingly weaker than auth-code protection. KV-backed enforcement
  *   is the answer if/when refresh replay becomes a hard requirement.
+ *
+ * Rolling cutover (`jti` undefined): tokens minted before this code
+ * shipped have no `jti` claim. `verifyJwt` validates signature + `exp`
+ * only — the runtime cast to `RefreshTokenClaims` does not enforce
+ * shape — so `claims.jti` is `undefined` for legacy tokens. If we keyed
+ * the cache on that, every legacy token would collide on a single
+ * `…/undefined` slot and the first post-deploy refresh by any user
+ * would lock out every other still-active session. We instead bypass
+ * enforcement for tokens without `jti`; they remain redeemable for the
+ * remainder of their natural TTL. New tokens (post-deploy) carry `jti`
+ * and get full enforcement. The bypass becomes dead code once all
+ * legacy refresh tokens age out (≤14 days) and can be removed in a
+ * follow-up.
  */
 async function checkAndMarkRedeemed(
   kind: "auth-code" | "refresh-token",
-  jti: string,
+  jti: string | undefined,
   ttlSeconds: number,
 ): Promise<Response | null> {
+  if (!jti) return null;
   const cache = caches.default;
   const cacheKey = new Request(`https://cache.local/oauth-${kind}/${jti}`);
   if (await cache.match(cacheKey)) {
--- a/workers/src/oauth/handlers.ts
+++ b/workers/src/oauth/handlers.ts
@@ -640,6 +640,7 @@ export async function handleAuthorize(
     user_email: session!.user_email,
     enc_tako_token,
     exp: now + AUTH_CODE_TTL_S,
+    jti: crypto.randomUUID(),
   };
   const code = await signJwt(codeClaims, cfg.signKey);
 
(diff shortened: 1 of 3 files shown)


robertabbott (approved):

workers/src/oauth/handlers.ts
```diff
@@ -770,6 +771,64 @@ export async function handleToken(req: Request, env: Env): Promise<Response> {
...
+ */
+async function checkAndMarkRedeemed(
+  kind: "auth-code" | "refresh-token",
+  jti: string | undefined,
+  ttlSeconds: number,
+): Promise<Response | null> {
+  if (!jti) return null;
+  const cache = caches.default;
+  const cacheKey = new Request(`https://cache.local/oauth-${kind}/${jti}`);
+  if (await cache.match(cacheKey)) {
```
robertabbott, line 823: `cache.match(cacheKey)` and `cache.put(cacheKey, ...)` are two separate awaits, and each yield lets another inflight request in the same isolate interleave. Two concurrent redemptions for the same `jti` could both see a cache miss before either writes. The existing caveats cover per-colo and LRU gaps — would it be worth adding a one-liner noting the non-atomic check-then-put as a third acknowledged limitation, since this is the security-hardening PR?

workers/src/oauth/types.ts
```diff
@@ -78,6 +82,10 @@ export interface RefreshTokenClaims extends BaseClaims {
   user_id: string;
   user_email: string;
   enc_tako_token: string;
+  /** Unique JWT ID for single-use enforcement (OAuth 2.1 §4.3.1
+   *  refresh-token rotation). Recorded in Workers Cache on successful
+   *  exchange so re-presenting the same refresh token is rejected. */
+  jti: string;
```
robertabbott: `RefreshTokenClaims.jti` is typed as required `string`, but legacy refresh tokens (minted before this deploy) will deserialize with `jti === undefined` via the `as T` cast in `verifyJwt`. `checkAndMarkRedeemed` handles this correctly (`string | undefined` param), but downstream code that trusts the type (e.g., `claims.jti.toUpperCase()` in a future log line) would type-check and crash. Would making `jti` optional (`jti?: string`) for the transition period — then tightening it in the follow-up cleanup when the bypass is removed — be worth the churn, or is the 14-day window short enough to leave it?

workers/src/oauth/handlers.ts
jfassio: Good catch. Added a third bullet to the `Caveats` block on `checkAndMarkRedeemed` acknowledging the TOCTOU window — two concurrent redemptions of the same `jti` in one isolate can both observe a cache miss before either `put` lands. Sequential replay (the realistic threat) is still serialized; concurrent replay is best-effort. Note that KV's atomic CAS is the path forward if/when this matters — same answer as the other two caveats.

workers/src/oauth/types.ts
jfassio: Worth the churn — narrowed `RefreshTokenClaims.jti` to `jti?: string` to match runtime reality, and added a JSDoc note pointing back to the bypass so we tighten both back to required `string` together when the 14-day legacy window passes. Left `AuthCodeClaims.jti` as required since the 60s TTL means no pre-deploy auth code can survive this deploy. Typecheck still clean.

jfassio:
Addressed 2 comments from this review round (893fae1). @claude-address

PR #68, opened 2026-05-05
feat(workers): per-client knowledge_search deep flow + report cleanup

jfassio:
## Summary

- **Knowledge-search deep path is now per-client.** Claude.ai keeps the existing single-tool `knowledge_search` flow with auto-escalation + MCP `notifications/progress` (resets the per-call timeout on each emission). ChatGPT — whose Apps SDK doesn't honor progress for timeout extension — gets a separate `start_deep_knowledge_search` / `wait_for_knowledge_search` pair that's only registered when `client === "chatgpt"`. Same agent UX in both: one call → chart inline on fast hits; deep just routes through different machinery underneath.
- **`wait_for_report` removed.** Its 12-call back-to-back polling pattern was crashing both ChatGPT and Claude. `create_report` now returns immediately with `webpage_url`, the agent shares the link + tells the user about the email notification, and `get_report` covers the on-demand "is it ready?" path with a single shot — no chained polls.
- **`create_report` URL hallucination guard.** ChatGPT was triggering `create_report` correctly but then sharing a non-Tako URL (apparently from a parallel web search). Description now has an explicit LINK RULE that names the `webpage_url` field, shows the format, and bans any kind of URL substitution.

## Why the back-and-forth in the commit history

The path to the final design wasn't straight: a kickoff/wait split (`92a7310`-era), then a single-call flow with progress notifications (`fe84a8c`), then the realization that ChatGPT doesn't include `progressToken` while Claude does (`5dd2667` log line confirmed it), then the per-client split landing in `9e7f17d`. A bug where my runtime gate used `progressToken` presence instead of UA-based client detection broke deep on Claude (`8c34b16` fixed it). Happy to squash before merge if you'd prefer a cleaner history — let me know.

## What's new on the branch

- `start_deep_knowledge_search` (ChatGPT-only) — POSTs deep, returns `task_id` in <1s, no widget.
- `wait_for_knowledge_search` (ChatGPT-only) — polls status up to 50s, returns COMPLETED-with-cards or `timed_out: true`. No widget. Mirrors the old `wait_for_report` shape.
- `_async_search_shape.ts` — shared types/helpers between `knowledge_search` (Claude path) and the kickoff/wait pair (ChatGPT path).
- `ToolContext.client: McpClientKind` and `ToolContext.sendProgress(progress, opts)` plumbed through `mcp.ts`. Tools that need to vary behavior across known host quirks read `client`; tools that want to keep long-running calls alive on Claude/etc. call `sendProgress`.
- `[mcp]` log line per tool call shows the detected client + whether the request carried a `progressToken`. Useful for diagnosing future client behavior changes.

## Removed

- `wait_for_report` (tool + tests) — see above.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test` — 206 passing
- [x] `npm run registry:check` — 11 tools, no drift
- [x] Staging deploy + manual exercise:
  - [x] Claude: fast query → chart inline; deep query → single-call flow completes with chart (verified pre-PR)
  - [x] ChatGPT: fast query → chart inline; deep query → kickoff message → wait loop → COMPLETED → `open_chart_ui` renders chart inline (verified pre-PR)
  - [x] Both clients: `create_report` returns the actual Tako URL, no web-search hallucination
  - [x] Both clients: `get_report` works as a one-shot status check; no chain-polling


workers/src/tools/wait_for_knowledge_search.ts
```diff
@@ -0,0 +1,223 @@
...
+    .string()
+    .optional()
+    .describe(
+      'Short progress summary (e.g. "2 progress events; last: PLANNING"). Useful while timed_out.',
+    ),
+});
+
+type Output = z.infer<typeof outputSchema>;
+
+function isTransientStatusError(err: unknown): boolean {
```
robertabbott: `isTransientStatusError` appears identically in `knowledge_search.ts:127` and here. Since `_async_search_shape.ts` already centralizes the other shared primitives (`COMPLETED_STATE`, `FAILURE_STATES`, `summarizeProgress`), would `isTransientStatusError` fit there too?

workers/src/tools/knowledge_search.ts
```diff
@@ -273,23 +136,139 @@ function isTransientStatusError(err: unknown): boolean {
...
   taskId: string,
 ): Promise<KnowledgeCard[]> {
   const startedAt = Date.now();
-  // Tracked across polls so the timeout/failure message can describe how
-  // far the pipeline got. We always pass `since_index = maxId + 1` so the
-  // backend filter (inclusive `>=`) returns only events strictly newer
-  // than what we've already seen — no duplicates, no need to dedupe here.
   let nextSinceIndex = 0;
   let latestEvents: AsyncTaskEvent[] = [];
   let transientRetriesLeft = MAX_TRANSIENT_RETRIES;
```
robertabbott, line 250: The old test suite covered three `pollDeep` scenarios that were removed: "transient failures exceed the retry budget" (3 consecutive 503s), "does not retry the status GET on a 404", and "propagates 404 as a typed Django error". The code paths they exercised (lines 284-295 here) are still present. `wait_for_knowledge_search.test.ts` covers equivalent scenarios, but for its own polling implementation (which differs: backoff vs fixed interval, returns `timed_out` vs throws). Would it be worth keeping a minimal retry-budget-exhaustion and 404-no-retry test against `pollDeep` directly, to lock its behavior independently?

workers/src/tools/knowledge_search.ts
```diff
@@ -332,184 +304,38 @@ async function pollAsyncKnowledgeSearch(
...
-    // Normalize casing once: the 202 POST returns lowercase ("pending")
-    // while subsequent GETs return uppercase ("PENDING", "COMPLETED", …).
-    // Comparing against the upper-cased value insulates us from any
-    // future drift on either side.
+    // Emit a progress notification to keep the client's tool-call
+    // timeout alive. The message carries a short "current pipeline
+    // step" hint pulled from the most recent event, when available.
+    // Best-effort — `sendProgress` swallows errors internally so a
+    // notification failure can't break the polling loop.
+    void ctx.sendProgress(pollCount, {
```
robertabbott: `void ctx.sendProgress(...)` fires the notification without awaiting. On non-terminal iterations the `setTimeout` sleep follows so the notification completes, but on the COMPLETED iteration (line 319) `pollDeep` returns immediately — the last progress notification races with the handler's return value. In practice the response goes through structuredContent/meta assembly in mcp.ts before being sent, so the notification likely completes first. Is the fire-and-forget intentional here, or would awaiting on the terminal iteration be worth it for ordering guarantees?

workers/src/tools/wait_for_knowledge_search.ts
```diff
@@ -0,0 +1,223 @@
+/**
+ * `wait_for_knowledge_search` — poll the async knowledge-search status
+ * endpoint and return when the task terminates OR when the budget
+ * elapses, whichever comes first. Companion to
+ * `start_deep_knowledge_search`.
+ *
+ * Registered ONLY on clients that don't honor MCP
+ * `notifications/progress` for tool-call timeout extension (currently:
+ * ChatGPT). See `mcp.ts`'s `CHATGPT_ONLY_TOOLS` set and
```
robertabbott: Minor: this comment says `CHATGPT_ONLY_TOOLS` but the actual const in `mcp.ts:178` is `CHATGPT_ONLY_TOOL_NAMES`. Same mismatch in `start_deep_knowledge_search.ts:17`.

robertabbott (approved):

workers/src/tools/wait_for_knowledge_search.ts
jfassio: Moved `isTransientStatusError` into `_async_search_shape.ts` alongside the other shared primitives. Both `knowledge_search.ts` and `wait_for_knowledge_search.ts` now import it; classification rule stays in lockstep across the two polling loops. (1b51b5d)

workers/src/tools/knowledge_search.ts
jfassio: Good catch — the prior tests covered `pollDeep`'s retry behavior, and `wait_for_knowledge_search`'s tests cover its own loop. Added two minimal tests in `knowledge_search.test.ts` to lock `pollDeep` independently:
- `surfaces the underlying DjangoHttpError after exceeding the transient retry budget (3 consecutive 503s)`
- `does not retry the status GET on a 404 (terminal task-not-found)`

Both run against the live `knowledge_search.handler` with `search_effort: "deep"` so the polling-loop branches at lines 284-295 stay covered. (1b51b5d)

workers/src/tools/knowledge_search.ts
jfassio: Good question. The fire-and-forget on intermediate iterations is intentional (the trailing sleep gives the notification well over a network round-trip to flush, and awaiting per-iteration would block the loop on a transport that's strictly less important than the next status GET). On the terminal iteration the race is real though — `pollDeep` returns/throws and `mcp.ts` immediately starts assembling the response.

Switched to await-on-terminal: the call captures the promise unconditionally, awaits it on COMPLETED / FAILED / INTERRUPTED, and `void`s it on non-terminal iterations. Behavior on the common path is unchanged; the terminal path now has an explicit ordering guarantee. (1b51b5d)

workers/src/tools/wait_for_knowledge_search.ts
jfassio: Fixed — both file-header comments now reference `CHATGPT_ONLY_TOOL_NAMES` (the actual const). (1b51b5d)

jfassio:
Addressed all 4 comments from this review round (1b51b5d). @claude-address

PR #70, opened 2026-05-05
fix(workers): clean Claude UX + reliable ChatGPT deep escalation

jfassio:
## Summary

Five composing fixes that land on the same surface (per-host chart UX), grouped because they interact:

1. **Claude.ai widget suppression restored + PNG fallback suppressed on the same gate.** Claude users get text + structured content + the model-pasted `[Open in Tako](embed_url)` link. No constrained iframe (chart was clipped vertically with a tiny scrollbar — bad UX), no random PNG.
2. **`knowledge_search` description: always include the link.** Was "do NOT paste the embed URL"; now "Always include `[Open in Tako](embed_url)` once at the end of your reply." Uniform contract across hosts. ChatGPT users get a sharable link alongside the inline widget; Claude users get the link as their path to the chart.
3. **`start_deep_knowledge_search` trigger relaxed.** Was gated on "the user wants a thorough/deep research answer," which made ChatGPT skip escalation on natural-phrasing queries. Now: "Use this whenever `knowledge_search` returns 0 cards OR errors out — auto-escalate by default, do not wait for the user to ask for 'deep' first."
4. **Tako 404 (no-results) → empty `SyncSearchResponse` translation.** Tako's `/api/v1/knowledge_search` returns HTTP 404 with `RelevantResultsNotFoundError` when fast finds 0 cards (REST-style "no resource matched"). Without this translation, the thrown `DjangoNotFoundError` bypasses both the server-side auto-escalation (Claude) and the LLM-side directive (ChatGPT), leaving the user with a `not_found` error envelope and a training-data answer.
5. **ChatGPT-specific `knowledge_search` widget suppression + empty-result throw.** Two ChatGPT quirks compose into the same fix:
   - ChatGPT pins widget container height once and ignores later shrink notifications, leaving a persistent empty-widget gap on `count: 0` calls.
   - Developer-mode ChatGPT routinely disregards description text and treats `count: 0` as a valid "no results" answer, falling back to training knowledge.

   Suppress the widget on `knowledge_search` for ChatGPT (forcing `open_chart_ui` as the single chart-rendering path — same pattern `start_deep_knowledge_search` and `wait_for_knowledge_search` already use) AND throw an actionable tool-call error on empty fast, with explicit kickoff/wait instructions in the error message. Belt-and-suspenders against model non-compliance.

## Resulting per-host behavior

**Claude.ai:** text + URL link. No widget rendered (the constrained-iframe rendering is bad UX, link is strictly better), no PNG fallback. Server-side fast→deep auto-escalation unchanged for empty results.

**ChatGPT:**
- Success path: `knowledge_search` returns cards → model chains into `open_chart_ui` to render → chart renders inline. One extra tool call vs. before, but a clean chat surface.
- Empty path: `knowledge_search` throws → model chains into `start_deep_knowledge_search` → `wait_for_knowledge_search` (loops) → `open_chart_ui` to render. Gap-free.

Both paths end at a rendered chart with the LLM-pasted link below.

## Test plan

- [x] `npm run test` — 211/211 (added 4 new tests for 404→empty translation and the ChatGPT empty throw; updated `index.test.ts` with the new "knowledge_search has no widget metadata on ChatGPT" invariant alongside the existing "claude.ai gets no widget metadata" invariant)
- [x] End-to-end on staging via local Worker + cloudflared tunnel:
  - Claude.ai: queries that hit cached cards return text + clickable Tako link, no constrained iframe.
  - ChatGPT success path (e.g. Pfizer): `knowledge_search` → `open_chart_ui` → chart renders inline.
  - ChatGPT empty path (e.g. "Daily caloric supply per capita India vs China vs US"): `knowledge_search` throws → model calls `start_deep_knowledge_search` → polls `wait_for_knowledge_search` → `open_chart_ui` → chart renders, no gap.

## Rollout note

Connectors cache `tools/list` per connection lifecycle. After deploy, anyone with an active claude.ai or ChatGPT connector needs to disconnect/reconnect to pick up the new descriptions and the `knowledge_search` widget-suppression. Worth flagging in #eng or wherever rollout coordination happens.

## Follow-ups (not in this PR)

- The dormant `extraContentBlocks` implementations on `open_chart_ui` and `knowledge_search` are now gated on `ui === undefined && !widgetSuppressed` and never fire for those tools. Worth a cleanup pass.
- Tako-side: convert `RelevantResultsNotFoundError` from HTTP 404 to `200 { results: [] }` on `/api/v1/knowledge_search` so this defensive translation can be removed and any future MCP-style consumer doesn't hit the same trap.


workers/src/tools/knowledge_search.ts
```diff
@@ -203,6 +225,31 @@ const knowledge_search = {
...
+    // which the model treats as a hard signal it must act on
+    // rather than a soft suggestion it can ignore.
+    //
+    // Doesn't fire when:
+    //   - cards.length > 0 (real results to return),
+    //   - input.search_effort === "deep" (the explicit-deep guard
+    //     above already redirected this path; we never reach here),
+    //   - ctx.client !== "chatgpt" (Claude.ai et al. already
+    //     auto-escalated above and won't see an empty result).
+    if (cards.length === 0 && ctx.client === "chatgpt") {
```
robertabbott: The guard here is `cards.length === 0 && ctx.client === "chatgpt"`, but it doesn't exclude `input.search_effort === "fast"`. If a ChatGPT caller explicitly passes `search_effort: "fast"` to force the cheap sync path (per the parameter description: "Omit to run fast first and auto-escalate..."), they'll get a thrown error telling them to call `start_deep_knowledge_search` — which contradicts their explicit intent to stay on the fast path.

The test at line 268 in the test file ("returns an empty result on Tako 404 when caller forces fast") only exercises the 404 path with `CTX` (client = `"unknown"`), not `chatgpt`. Should this guard also check `input.search_effort !== "fast"`, or is the intent that ChatGPT callers should never see a clean `count: 0` from fast, even when they explicitly asked for it?

workers/src/tools/knowledge_search.ts
```diff
@@ -123,7 +123,7 @@ type Output = z.infer<typeof outputSchema>;
...
...e BEFORE replying — do not narrate the data first.** **Always include `[Open in Tako](embed_url)` once at the end of your reply** for the top card so the user has a sharable, fullscreen link regardless of host. Do NOT echo `![…](image_url)` markdown for the top card — that produces a click-to-load duplicate of the same chart. Mention other titles by name and tell the user they can ask to chart any of them — those follow-ups also go through `open_chart_ui`. **`search_effort: "deep"` and the default fast→deep auto-escalation can run up to ~5 minutes server-side**; this tool emits MCP `notifications/progress` events while polling so clients that set `resetTimeoutOnProgress: true` (the SDK default — claude.ai and most others) keep the call alive for the full duration. **On ChatGPT, server-side auto-escalation is disabled** (its Apps SDK doesn\'t honor progress notifications, so polling would trip the host timeout). **If this tool returns 0 cards or errors out on a ChatGPT client, immediately call `start_deep_knowledge_search` with the same query to retry via the kickoff/wait flow — don\'t stop, don\'t fall back to your own training knowledge, and don\'t wait for the user to re-ask.**',
```
robertabbott: The description now contains host-conditional directives ("On Claude.ai the top result auto-renders inline...", "On ChatGPT no chart renders inline from this tool..."). The model reading this description doesn't know which host it's running on — it sees the same tool listing regardless. Does the model reliably self-identify as "on ChatGPT" vs "on Claude.ai" and follow the correct branch? Or does this create ambiguity where the model might follow the wrong host's directive — e.g., a ChatGPT model narrating the chart instead of chaining into `open_chart_ui`, or a Claude model calling `open_chart_ui` unnecessarily?

workers/src/mcp.ts
```diff
@@ -278,16 +309,15 @@ function registerTool(
...
-  // by inspecting the User-Agent header.
-  const widgetSuppressed = options.client === "claude";
+  // inside a constrained, non-resizable iframe container that clips
+  // the chart vertically and exposes a tiny scrollbar — strictly
+  // worse UX than just giving the LLM a markdown link to the
+  // standalone embed page on tako.com. The chart-bearing tool
+  // descriptions carry an unconditional "always include `[Open in
+  // Tako](embed_url)`" directive so the user always has a clickable
+  // path to the fully-interactive chart, regardless of client.
+  const widgetSuppressed =
```
robertabbott, line 351: `widgetSuppressed` is now true for two different intents: claude.ai (blanket — no widget, no PNG, link only) and ChatGPT+`knowledge_search` (per-tool — no widget on this tool, but widgets are fine on `open_chart_ui`). The `extraContentBlocks` gate at line 577-580 uses `!widgetSuppressed` to also skip the PNG fallback for ChatGPT+`knowledge_search` — is that intentional? The PR description says ChatGPT gets the chart via `open_chart_ui` so it's presumably fine, but the single boolean conflates "skip everything" with "skip widget only" in a way that could surprise a future reader adding a third suppression case.

workers/src/tools/knowledge_search.ts
```diff
@@ -146,13 +146,35 @@ const knowledge_search = {
...
+      try {
+        return await djangoPost<SearchPostResponse>(
+          ctx.env,
+          ctx.token,
+          "/api/v1/knowledge_search",
+          body,
+          { timeoutMs: 130_000 },
+        );
+      } catch (err) {
+        if (err instanceof DjangoNotFoundError) {
```
robertabbott: The catch replaces the 404 body (which contains `detail` / `error_message`) with a bare `{ outputs: { knowledge_cards: [] } }`, discarding the signal. If Tako ever changes the 404 semantics — e.g., `knowledge_search` endpoint gets moved and the reverse proxy returns a genuine 404 — this translation would silently convert a real "not found" into a false empty-result. Is there a way to distinguish "no results matched" 404s from "resource doesn't exist" 404s, for instance by checking for `RelevantResultsNotFoundError` or a known `detail` string in the response body before translating?

robertabbott (approved):

PR #79, opened 2026-06-20
feat(tools)!: GA-API parity — tako_search/answer/contents/agent (TAKO-3221)

robertabbott:
## Summary

Brings the canonical (Workers) MCP tool set to parity with Tako's GA public API. Part of TAKO-3221 (MCP server parity). **Breaking** tool rename → version bump **0.1.0 → 0.2.0**.

- **`knowledge_search` → `tako_search`** — rename only; intentionally keeps the legacy `/api/v1/knowledge_search` endpoint (it has fast + deep + async; `/api/v3/search` is fast-only, TAKO-3183).
- **`grounding` → `tako_answer`** — migrated to GA `/api/v1/answer`; request is the v3 `{query, source_indexes}` shape; result is `{answer, cards, web_results, request_id}` (drops `tako_selected`/`confidence`).
- **`tako_contents`** (new) — `POST /api/v1/contents`; returns the presigned download URL and inlines extracted web text (capped).
- **`tako_agent`** (new) — Agent API `POST /api/v1/agent/runs` + poll `GET /api/v1/agent/runs/{run_id}`, emitting `notifications/progress` each poll. ChatGPT (no progress-reset support) gets a `tako_agent_start` / `tako_agent_wait` split, gated in `mcp.ts` exactly like the existing deep-search split.

Tools stay hand-written; the registry (`_registry.ts` + `registry/server.json`) is regenerated (16 tools). Auth is unchanged (connection-level OAuth/Bearer — tools added zero auth code). The non-GA tools (reports/chart/credit + the ChatGPT deep-search split) are intentionally untouched this pass.

## Lines changed (logic / tests / docs)
- **Logic:** ~+600 — the 4 GA tools + `tako_agent` split + `mcp.ts` gating + regenerated registry
- **Tests:** ~+318 — per-tool vitest (incl. fake-timer agent polling) + `index.test.ts` gating assertions
- **Docs/version:** ~+28 — README/AGENTS/llms\* tool reference + breaking-change note + 0.2.0 bump
- Total: 23 files, ~+970 / −480

## Testing
- `npm test` — **235/235**; `npm run registry:check` — ok (16 tools, no drift); `npm run typecheck` — clean (strict, incl. `noUncheckedIndexedAccess`).
- Built subagent-driven (5 tasks, per-task review + an opus whole-branch review — no Critical/Important findings). Agent-API slash-less route path verified against the backend urlconf.

## ⚠️ Billing caveat
`tako_agent` over MCP is **unbilled for PAYG orgs** until **TAKO-3245** (Agent API metering) lands — the Agent API currently has no `@meters_api_credits`. Caveat is also commented in the tool files. Not a blocker, but don't broadly promote agent-over-MCP to PAYG customers until 3245 ships.

## Deferred (non-blocking, noted in review)
- `tako_contents`: add truncation-cap + fetch-failure-degradation tests.
- `llms-full.txt` `tako_answer` entry: cosmetic api_token note.


workers/src/tools/tako_agent_wait.ts
```diff
@@ -0,0 +1,46 @@
...
+  inputSchema,
+  outputSchema: agentRunSchema,
+  annotations: {
+    title: "Tako: Wait for Agent Run",
+    readOnlyHint: true,
+    destructiveHint: false,
+    openWorldHint: false,
+  },
+  async handler(input, ctx): Promise<AgentRun> {
+    return pollAgentRun(ctx, input.run_id);
```
robertabbott: The whole reason `tako_agent_start`/`tako_agent_wait` exists (per `mcp.ts`: "ChatGPT's Apps SDK doesn't send a progressToken … can't survive ChatGPT's ~60 s ceiling") is so each `wait` call returns *before* that ceiling and the model re-polls. But this handler delegates straight to `pollAgentRun`, which loops until `completed`/`failed` with no per-call budget — so for a 60–90 s run it blocks the single call past ChatGPT's ~60 s timeout, exactly the failure the split was built to prevent. The parallel `wait_for_knowledge_search` caps a single call at `MAX_WAIT_CEILING_S = 50` and returns `timed_out: true` to drive the chain. This tool's own description ("call this tool again with the same `run_id`", "cap the chain at 12 calls") describes that bounded-poll contract, but `agentRunSchema` has no `timed_out`/still-running terminal and `pollAgentRun` never returns early. Should `tako_agent_wait` do a budget-capped poll (~50 s) and return a non-terminal "still running" status so the model can re-call, mirroring `wait_for_knowledge_search`?

workers/src/tools/tako_agent.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+  return data.run_id;
+}
+
+/** Poll an agent run to a terminal state, emitting progress each iteration. */
+export async function pollAgentRun(ctx: ToolContext, runId: string): Promise<AgentRun> {
+  let transient = 0;
+  let pollCount = 0;
+  // Budget: poll every 5s; the SDK timeout is kept fresh by sendProgress.
+  while (true) {
```
robertabbott, line 97: `pollDeep` in `tako_search.ts` enforces `POLL_BUDGET_MS` and throws an actionable timeout when the deep task overruns. `pollAgentRun`'s `while (true)` only exits on `completed`/`failed` (or >2 transient transport errors); a run stuck in `queued`/`running` polls indefinitely. On the Claude single-tool path `ctx.sendProgress` keeps the SDK timeout fresh on every poll, so nothing else bounds it either. Should this carry an absolute budget cap (like `pollDeep`) that throws "agent run didn't complete within Ns" so a wedged backend run can't poll forever?

workers/src/tools/tako_answer.ts
```diff
@@ -0,0 +1,101 @@
...
+    title: "Tako: Answer",
+    readOnlyHint: true,
+    destructiveHint: false,
+    openWorldHint: false,
+  },
+  async handler(input, ctx): Promise<Output> {
+    // GA /api/v1/answer takes the v3 SearchRequest shape: top-level `query`
+    // (NOT inputs.text) + `source_indexes`. Answer runs the fast pipeline +
+    // arbiter (sync, ~120s ceiling) — no async/deep path, so no polling.
+    const body = {
```
robertabbott, line 80: The old `grounding` tool passed `country_code` and `locale` through to the backend; `tako_answer` drops both and sends only `{query, source_indexes}`. Does GA `/api/v1/answer` (the v3 SearchRequest shape) accept `country_code`/`locale`? If it does, is dropping them an intentional removal of localized-answer capability, or worth keeping for parity with `tako_search` (which still has them)?

workers/src/tools/tako_contents.ts
```diff
@@ -0,0 +1,92 @@
...
+        "Tako contents endpoint returned no downloadable content for that URL.",
+      );
+    }
+    const isText = (item.format ?? "").toLowerCase() === "text";
+    let text: string | null = null;
+    if (isText) {
+      // Inline the extracted web text so the model can read it without a
+      // second tool call. Best-effort: a fetch failure still returns the URL.
+      try {
+        const resp = await fetch(item.url);
```
robertabbott: Every other outbound fetch in the Workers code passes `AbortSignal.timeout(...)` (`django.ts:304`, `oauth/identity.ts:96`). This raw `fetch(item.url)` has none, so a slow or hung presigned-URL host stalls the entire tool call — the `try/catch` only catches thrown errors, not a socket that never responds. Should this fetch pass `AbortSignal.timeout(...)` to bound it like the rest of the codebase?

workers/src/tools/tako_agent_wait.ts
robertabbott: Fixed (93c5eef). You're right — delegating straight to `pollAgentRun` defeated the split. `pollAgentRun` now takes `{ budgetMs, onTimeout }`; `tako_agent_wait` gained a `max_wait_seconds` input (cap/default `AGENT_WAIT_CEILING_S = 50`) and calls it with `onTimeout: "return"` — so it does a budget-capped poll and returns the last-known non-terminal run with a new **`timed_out: true`** field (added to `agentRunSchema`) when the deadline elapses, mirroring `wait_for_knowledge_search`. The description now tells the model to re-call with the same `run_id` while `timed_out` is true. A fake-timer test exercises the `timed_out:true` path.

workers/src/tools/tako_agent.ts
robertabbott: Fixed (93c5eef). `pollAgentRun` now carries an absolute deadline (`Date.now() + budgetMs`) like `pollDeep`. The Claude single-tool path (`tako_agent`) uses `{ budgetMs: AGENT_POLL_BUDGET_MS = 295_000, onTimeout: "throw" }` → throws "Agent run <id> did not complete within Ns" if the run never reaches terminal, so a wedged backend can't poll forever. The per-GET request timeout is also bounded (15s) so one socket can't burn the budget. Fake-timer test asserts the throw-on-overrun path.

workers/src/tools/tako_answer.ts
robertabbott: Good catch — restored (93c5eef). Confirmed the GA `/api/v1/answer` (v3 `SearchRequest`) does accept `country_code` (default "US") and `locale` (BCP-47, default "en-US"), so dropping them was an unintended parity regression. Both are back on `tako_answer`'s input schema (with the same defaults as `tako_search`, so callers needn't pass them) and in the POST body. Test asserts they're sent.

workers/src/tools/tako_contents.ts
robertabbott: Fixed (93c5eef). The inline web-text fetch now passes `AbortSignal.timeout(CONTENTS_FETCH_TIMEOUT_MS)` (15s), matching the rest of the codebase — a hung presigned-URL host now aborts and the existing catch degrades to `text: null` instead of stalling the tool call. Added a test that a rejecting fetch yields `text: null` without throwing.

robertabbott:
Addressed all 4 comments from this review round — budget-capped `pollAgentRun` (Claude path throws at 295s; `tako_agent_wait` does a 50s-capped poll returning `timed_out` so ChatGPT re-polls), restored `country_code`/`locale` on `tako_answer`, and bounded the `tako_contents` inline fetch with `AbortSignal.timeout`. 239/239 tests, registry ok (16 tools), typecheck clean. @claude-address

workers/src/tools/tako_agent.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+import { z } from "zod";
+
+import { djangoGet, djangoPost } from "../django.js";
+import type { ToolContext, ToolModule } from "./types.js";
+
+const POLL_INTERVAL_MS = 5_000;
+const MAX_TRANSIENT_ERRORS = 2;
+export const AGENT_POLL_BUDGET_MS = 295_000;
+export const AGENT_WAIT_CEILING_S = 50;
```
noahjax: [Suggestion] `AGENT_WAIT_CEILING_S = 50` plus a single poll-GET hanging up to `AGENT_POLL_REQUEST_TIMEOUT_MS` (15s) near the deadline gives a ~65s worst-case wall-clock before `tako_agent_wait` returns `timed_out`. The `tako_agent_start`/`tako_agent_wait` split exists precisely to return under ChatGPT's ~60s tool-call ceiling, so this edge can undermine its purpose. Consider lowering the wait ceiling to ~40s, or using a shorter request timeout for poll GETs on the wait path. Low probability (needs an anomalous slow status GET), hence non-blocking.

workers/src/tools/tako_agent.ts
```diff
@@ -0,0 +1,155 @@
...
+    card_id: z.string().nullable().optional(),
+    title: z.string().nullable().optional(),
+    embed_url: z.string().nullable().optional(),
+    image_url: z.string().nullable().optional(),
+  })
+  .loose();
+
+export const agentRunSchema = z.object({
+  run_id: z.string(),
+  status: z.enum(["queued", "running", "completed", "failed"]),
```
noahjax, line 49: [Nit] `status` is a closed 4-member enum, so any future backend status addition makes the poll throw "unexpected shape" until both sides update in lockstep. This mirrors the backend `AgentRunStatus` StrEnum exactly (intentional), but the backend type carries a comment noting that lockstep — worth mirroring that note here so a maintainer doesn't treat it as an oversight.

workers/src/tools/tako_contents.ts
```diff
@@ -0,0 +1,93 @@
...
+  url: z
+    .string()
+    .min(1)
+    .describe("The result URL to fetch content for (a Tako card URL → CSV; any other URL → page text)."),
+});
+
+const outputSchema = z.object({
+  format: z.string(),
+  download_url: z.string(),
+  expires_at: z.string(),
```
noahjax, line 29: [Suggestion] The output schema drops the backend `ContentItem.cost`. Web-text contents is metered ($1/1k pages, per `content_types.py`); unlike `tako_agent` (unbilled, caveated), this call actually charges the org — but the agent can't surface what it cost because `cost` isn't in the output. Consider including it.

workers/src/tools/tako_contents.ts
```diff
@@ -0,0 +1,93 @@
...
+ * inline the text so the model can read it directly.
+ */
+import { z } from "zod";
+
+import { djangoPost } from "../django.js";
+import type { ToolModule } from "./types.js";
+
+const DESCRIPTION =
+  "Fetch the underlying content behind a result URL. A Tako card URL returns a CSV of the card's data; any other URL returns the page's extracted full text. Pass a single `url` (a TakoCard.webpage_url or a web result URL). Returns a short-lived `download_url`; for web pages the extracted `text` is also inlined so you can read it directly.";
+
```
noahjax, line 15: [Nit] The `MAX_INLINE_CHARS` truncation path has no test. The fetch-failure-degradation case is covered, but the truncation branch (the other case the PR self-noted as deferred) isn't exercised.

noahjax (approved):
Solid, well-tested GA-parity pass — APPROVE. I cross-checked every tool's request/response mapping against the backend monorepo and they all match: `tako_answer` ↔ `AnswerResponse`, `tako_agent` ↔ `AgentRunRequest`/`AgentRun`/`AgentResult`/`ErrorObject` (status enum is an exact mirror of `AgentRunStatus`; 202 dispatch is handled by `django.ts`'s `response.ok` gate), and `tako_contents` ↔ `ContentsResponse`/`ContentItem` (`ContentFormat` is `csv`/`text`). The `mcp.ts` gating composes correctly with the existing ChatGPT split (no tool in both sets; counts match `index.test.ts`). The author's self-review fixes (contents fetch timeout, restored `country_code`/`locale`, poll-budget/deadline split) all landed. Comments below are non-blocking.

**Two design follow-ups to track:**

1. **Naming inconsistency** — the renamed `tako_search`/`tako_answer` now sit beside the un-renamed `start_deep_knowledge_search`/`wait_for_knowledge_search` deep-search split. Intentionally out of scope this pass, but worth a tracked follow-up so the whole surface is consistently `tako_*`.

2. **Endpoint-version asymmetry** — `tako_search` deliberately stays on legacy `/api/v1/knowledge_search` while `tako_answer`/`tako_agent` move to GA endpoints. I verified the rationale in the backend: `/api/v3/search` runs `run_fast_pipeline` and is fast-only today (`v3/search/types.py`: "Only 'fast' is currently supported"), whereas v1 has the `_delegate_to_async` deep/async path. So a repoint to v3 today would silently drop deep auto-escalation. The inline comment guarding this (repoint once v3 gains deep, TAKO-3183) is load-bearing — keep it.

workers/src/tools/tako_agent.ts
robertabbott: Good call. Lowered `AGENT_WAIT_CEILING_S` 50 → **40** (6b19bc3), so the worst case (a poll-GET hanging the full 15s right at the deadline) returns in ~40+15 = **55s**, comfortably under ChatGPT's ~60s ceiling. Added a comment spelling out that math.

workers/src/tools/tako_agent.ts
robertabbott: Done (6b19bc3) — added a comment above the `status` enum in `agentRunSchema` noting it mirrors the backend `AgentRunStatus` StrEnum and must stay in lockstep (a new backend status must be added here too, or the poll rejects the run as an unexpected shape).

workers/src/tools/tako_contents.ts
robertabbott: Added (6b19bc3). `cost` is now in the `tako_contents` output schema (passed through from the backend `ContentItem.cost`; card CSV → 0, web text → the metered amount), so the agent can report what the call charged. Tests assert cost 0 for CSV and the metered value for web text.

workers/src/tools/tako_contents.ts
robertabbott: Added (6b19bc3) — a test fetches a body one char past `MAX_INLINE_CHARS` and asserts the result ends with `...[truncated]` and is exactly `MAX_INLINE_CHARS + len("\n...[truncated]")` long.

robertabbott:
Addressed noahjax's 4 follow-ups — lowered the agent wait ceiling to 40s (worst-case ~55s, under ChatGPT's ceiling), surfaced `cost` on `tako_contents` output, added a status-enum lockstep comment, and a truncation-path test. 240/240 tests, registry ok (16 tools), typecheck clean. @claude-address

PR #80, opened 2026-06-20
feat(registry): publish tako-mcp to the official MCP Registry as a remote server

robertabbott:
Adds the concrete repo pieces to list tako-mcp on the official [MCP Registry](https://registry.modelcontextprotocol.io) as a hosted remote server, and automates publishing. This is the discoverability follow-up to the GA tool-parity work (#79) and part of TAKO-3221.

## What this does

- **`server.json`** (repo root) — the official-registry descriptor, conforming to the current `2025-12-11` server schema. A remote `streamable-http` entry pointing at `https://mcp.tako.com/mcp`. The registry schema does **not** list tools; hosts discover them at runtime via `tools/list`. This is intentionally **distinct** from `registry/server.json`, the generated in-repo tool catalog (different schema, includes a `tools[]` array, consumed by `registry:gen`/`registry:check` and Smithery).
- **`.github/workflows/publish-mcp.yml`** — publishes `server.json` via `mcp-publisher`, authenticating with **GitHub OIDC** (no secret). **The version lives in code:** the workflow triggers on a **push to `main` that bumps `server.json`'s `version`** (a guard step compares the version at HEAD vs the pre-push commit and skips publishing when it's unchanged, so non-version edits never trigger a duplicate-version republish). Manual `workflow_dispatch` publishes the checked-in version on demand. No `v*` tag required.
- **`registry/metadata.json` + regenerated `registry/server.json`** — corrects the stale legacy-Python catalog (`runtime: python` / pip / `sse:8001` / `api_key`) to the live remote Worker (`streamable-http` at `mcp.tako.com/mcp`, OAuth + Bearer auth) and refreshes the description for the GA tool surface. `npm run registry:check` is green.
- **`registry/smithery.yaml`** — same staleness fix: remote `streamable-http` server, OAuth + optional Bearer token, GA tool surface, refreshed examples (no more connection-level `api_token`).
- **`agent.json`** — bumped to `0.2.0`; advertises OAuth (web hosts) alongside the API-key Bearer path (CLI/desktop).
- **README** — maintainer section covering the listing, the OIDC publish flow, and the `com.tako`/DNS upgrade path.

## Decisions

- **Namespace `io.github.TakoData/tako-mcp`** (not `com.tako/tako-mcp`). GitHub OIDC authorizes the `io.github.<org>/*` namespace automatically because this repo lives in the TakoData org — **zero secrets, CI-native**. Casing confirmed against the registry source: the OIDC grant is built verbatim from `repository_owner` (`io.github.<owner>/*`) and matched case-sensitively (`strings.HasPrefix`), with no name normalization on publish — so the mixed-case `TakoData` is required to match the grant (lowercasing would break it). The branded `com.tako/*` name needs DNS auth (Ed25519 key + apex `TXT` on tako.com); documented in the README as a future upgrade (swap one workflow step).
- **Two `server.json` files on purpose.** Root = official registry (no tools, remote URL). `registry/server.json` = generated internal/Smithery catalog (tools array). They serve different consumers; the README and workflow comments call this out.

## Lines changed
- Logic/config: ~125 (root `server.json`, workflow, `registry/*`, `agent.json`)
- Tests: 0 (metadata/CI only — no runtime code touched; `registry:check` + `typecheck` are the gates)
- Docs: ~20 (README)

## Test plan
- [x] `npm run registry:gen` regenerates cleanly; `npm run registry:check` → `ok (16 tools)`
- [x] `npm run typecheck` passes
- [x] All JSON parses; workflow YAML sanity-checked
- [ ] **First publish:** merge this PR, then bump `server.json`'s `version` and merge (or run the workflow manually) → confirm the entry appears in the registry. **The MCP Registry is in preview** — verify `mcp-publisher`'s latest CLI flags before the first real publish.

### DIFF
--- a/.github/workflows/publish-mcp.yml
+++ b/.github/workflows/publish-mcp.yml
@@ -1,0 +1,46 @@
+name: Publish to MCP Registry
+
+# Publishes ./server.json (the io.github.TakoData/tako-mcp remote-server
+# descriptor) to the official MCP Registry (registry.modelcontextprotocol.io)
+# via GitHub OIDC. OIDC needs NO secret: the io.github.TakoData/* namespace is
+# authorized because this repo lives in the TakoData GitHub org.
+#
+# NOTE: ./server.json is the official-registry descriptor (remote streamable-http
+# at mcp.tako.com/mcp; tools are discovered at runtime). It is distinct from
+# registry/server.json, which is the in-repo generated tool catalog used by
+# `npm run registry:gen`/`registry:check`.
+
+on:
+  push:
+    tags: ["v*"] # publish on a release tag, e.g. v0.2.0
+  workflow_dispatch: {} # or publish on demand
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # required for OIDC authentication
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Install mcp-publisher
+        run: |
+          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
+
+      # On a release tag, stamp server.json's version from the tag so the
+      # published version matches the tag (the registry rejects re-publishing an
+      # already-published version). On workflow_dispatch, the checked-in version
+      # is published as-is.
+      - name: Sync server.json version to the release tag
+        if: startsWith(github.ref, 'refs/tags/v')
+        run: |
+          VERSION="${GITHUB_REF#refs/tags/v}"
+          jq --arg v "$VERSION" '.version = $v' server.json > server.tmp && mv server.tmp server.json
+
+      - name: Authenticate to the MCP Registry (GitHub OIDC)
+        run: ./mcp-publisher login github-oidc
+
+      - name: Publish to the MCP Registry
+        run: ./mcp-publisher publish
--- a/.github/workflows/publish-mcp.yml
+++ b/.github/workflows/publish-mcp.yml
@@ -5,15 +5,23 @@ name: Publish to MCP Registry
 # via GitHub OIDC. OIDC needs NO secret: the io.github.TakoData/* namespace is
 # authorized because this repo lives in the TakoData GitHub org.
 #
+# Trigger model: the version lives in code (server.json's "version"), so a merge
+# to main that BUMPS that version publishes automatically. A merge that touches
+# server.json without changing the version (e.g. editing the description) is a
+# no-op — the guard step below skips publishing so the registry never sees a
+# duplicate-version republish. workflow_dispatch publishes the checked-in
+# version on demand.
+#
 # NOTE: ./server.json is the official-registry descriptor (remote streamable-http
 # at mcp.tako.com/mcp; tools are discovered at runtime). It is distinct from
-# registry/server.json, which is the in-repo generated tool catalog used by
+# registry/server.json, the in-repo generated tool catalog used by
 # `npm run registry:gen`/`registry:check`.
 
 on:
   push:
-    tags: ["v*"] # publish on a release tag, e.g. v0.2.0
-  workflow_dispatch: {} # or publish on demand
+    branches: [main]
+    paths: ["server.json"]
+  workflow_dispatch: {}
 
 jobs:
   publish:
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-mcp.yml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # required for OIDC authentication
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+        with:
```
noahjax, line 35: [Suggestion] `fetch-depth: 2` can misfire on multi-commit pushes. The guard runs `git show ${{ github.event.before }}:server.json`, but with depth 2 only HEAD and its parent are fetched; a push landing >=2 commits leaves `github.event.before` (the pre-push tip) out of history, so `git show` fails -> `PREVIOUS=""` -> `changed=true` -> publish is attempted even when the version is unchanged. The registry's duplicate-version rejection (which you note) is a backstop, but the run would then error on the publish step. `fetch-depth: 0` (or explicitly fetching `event.before`) avoids it. Squash-merges (1 commit/merge) sidestep this, so impact depends on the repo's merge strategy.

server.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
+  "name": "io.github.TakoData/tako-mcp",
```
noahjax, line 3: [Suggestion] `io.github.TakoData/tako-mcp` is schema-valid (the name pattern allows mixed case), but worth confirming GitHub OIDC authorizes that exact casing before the first publish — some registry tooling normalizes `io.github.<owner>` to lowercase, which would mismatch `TakoData`. Fits your existing "verify before first publish" checkbox.

noahjax (approved):
Clean discoverability follow-up to #79 — APPROVE. I verified the root `server.json` against the live `2025-12-11` registry schema: `title` is a valid optional field, the required `name`/`description`/`version` are present, the `name` pattern permits `io.github.TakoData/tako-mcp` (mixed case allowed), and `remotes[].headers[]` with `name`/`description`/`isRequired`/`isSecret` are all valid `KeyValueInput` keys. `isSecret: true` is set correctly on the Authorization header. `registry/metadata.json`, `registry/server.json`, `smithery.yaml`, and `agent.json` are mutually consistent (oauth2 + apiKey, streamable-http) and `registry:check` is green. The two-`server.json` split and the OIDC-namespace-now / DNS-namespace-later decisions are sound and well-documented. Comments below are non-blocking.

**One doc nit:** the PR description is stale on the trigger model — the "What this does" bullet and the test plan describe a `v*` release-tag flow with version stamping, but the implemented workflow triggers on push-to-`main` + `server.json` version-bump detection (no tag, no stamping). The workflow's own comments are correct; just worth fixing the description/test-plan so the first-publish steps match reality.

.github/workflows/publish-mcp.yml
robertabbott: Good catch — fixed in b57e731. Changed `fetch-depth: 2` → `fetch-depth: 0` so `github.event.before` is always in history regardless of merge strategy, so the version-diff guard no longer misfires (and never attempts an unchanged-version publish that the registry would reject). Thanks!

server.json
robertabbott: Confirmed against the registry source rather than deferring it — and the result is reassuring: the OIDC grant is built verbatim from `repository_owner` as `io.github.<owner>/*` (`github_oidc.go:288-294`), matched case-sensitively via `strings.HasPrefix` (`jwt.go:165-170`), and the publish path passes `input.Body.Name` to that matcher with no normalization (`publish.go:53`; `validators.go` does not lowercase the name). GitHub emits the org's canonical login (`TakoData`) in the claim, so `io.github.TakoData/tako-mcp` matches the `io.github.TakoData/*` grant exactly. Lowercasing to `takodata` would actually break the prefix match, so keeping the mixed case as-is.

robertabbott:
Re: the stale PR description — fixed. Rewrote the "What this does" workflow bullet and the test plan to describe the implemented **push-to-`main` + `server.json` version-bump** trigger (with the skip-on-unchanged guard), and removed the `v*` tag / version-stamping language. Thanks for the thorough schema verification and the approve!

robertabbott:
Addressed 3 comments from this review round: fixed the `fetch-depth` guard misfire (→ `fetch-depth: 0`), verified the OIDC namespace casing against the registry source (mixed-case `TakoData` is correct — kept as-is), and updated the stale PR description to the version-bump-on-merge trigger. @claude-address

PR #85, opened 2026-06-22
feat(tako_search): add sources param + web_results output

robertabbott:
Closes the gap where `tako_search` could only return curated Tako cards — it was hardcoded to `source_indexes: ["tako"]` and dropped web results, so MCP users could blend tako+web only via `tako_answer`, not search.

The legacy `/api/v1/knowledge_search` endpoint already returns `outputs.web_results`, so this is purely a Worker change — **no need to move `tako_search` to `/api/v3/search`** (which is fast-only today and would lose deep search; TAKO-3183).

## Changes
- **`sources` input param** (`["tako"]` / `["web"]` / `["tako","web"]`, default `["tako"]`), passed through as `source_indexes`. Default preserves the chart-first behavior.
- **`web_results` in the output** — added to the shared results output shape with `.default([])` so the deep split tools (`wait_for_knowledge_search`) stay valid without changes; threaded through both the fast and deep/async paths.
- **Escalation gating** — the deep auto-escalation and the ChatGPT empty-redirect now require `tako` in `sources` **and** zero cards **and** zero web results, so a web-only search (or one that already returned web results) no longer pointlessly escalates the tako deep pipeline or discards web hits. With the default `["tako"]`, behavior is unchanged.

## Testing
- 4 new tests (sources→source_indexes pass-through, web_results surfaced, web-only doesn’t escalate/throw on 0 cards, default sources `["tako"]`).
- Full suite **251 passed**; `typecheck` + `registry:check` clean; `registry/server.json` regenerated.

Docs follow-up: TakoData/docs#269 updated to document the `sources` param + `web_results` on the search tool.

## Lines changed
- Logic: ~45 (`tako_search.ts`, `_async_search_shape.ts`)
- Tests: ~75
- Generated: `registry/server.json`

### DIFF
--- a/registry/server.json
+++ b/registry/server.json
@@ -394,6 +394,13 @@
           "description": "Maximum number of matching cards to return (1-20).",
           "default": 10
         },
+        "sources": {
+          "type": "array",
+          "description": "Which source(s) to search: `[\"tako\"]` (default) returns curated knowledge cards that render as charts; `[\"tako\",\"web\"]` also returns live web results; `[\"web\"]` returns web results only. Web results come back in `web_results` (they are not chartable cards). Deep research (the fast→deep escalation and `search_effort: \"deep\"`) applies to the `tako` source only.",
+          "default": [
+            "tako"
+          ]
+        },
         "search_effort": {
           "type": "string",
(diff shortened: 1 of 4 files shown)


noahjax (commented):
Would we lose that much by porting to `/api/v3/search` now? Seems like deep mode mcp is confusing given how different it is from the rest of knowledge_search. Also doesn't the web index for knowledge_search currently also include an answer, making it redundant with `/answer`?

PR #92, opened 2026-06-24
feat(workers): serve OpenAI connector-directory domain-verification token

jfassio:
## What

Adds a `GET /.well-known/openai-apps-challenge` route that serves the OpenAI connector-directory domain-verification token as plain text, read from a per-env `OPENAI_APPS_CHALLENGE_TOKEN` binding. Returns 404 when unset so an environment never satisfies a verification it wasn't issued.

This unblocks the OpenAI Apps submission for `mcp.tako.com`, which was failing domain verification because the endpoint 404'd.

## Why it was missing

The route was originally built on `jfassio/resource-bare-origin` (May 6) but that branch was never merged — so it never reached main or prod. This re-applies just the challenge changes on top of current main (the old branch bundled an unrelated OAuth change and predated the tool-surface cleanup), with the **current** dashboard token.

## Tokens
- **production** (`mcp.tako.com`): `uXTBcUGnISWBPU2YXEn7IPIDh2ANMxLa1rkFvv3Z0c0` — the value now shown in the OpenAI dashboard (OpenAI rotated it from the old branch's value)
- **staging** (`mcp.staging.tako.com`): prior issued token, kept for parity
- **dev**: stub the integration test pins

## Verification
- `npm run typecheck` ✅
- `npm run registry:check` ✅
- `src/index.test.ts` — 18/18 pass, incl. new route test ✅

## ⚠️ Deploy note
Merging alone does **not** make the endpoint respond. The route goes live only after `wrangler deploy --env production` runs against `mcp.tako.com`. Since CI prod deploys fail on the zone-route token error, prod ships out-of-band — someone with CF account access needs to deploy before OpenAI's "Verify Domain" will pass.

### DIFF
--- a/workers/src/index.test.ts
+++ b/workers/src/index.test.ts
@@ -1,5 +1,7 @@
 import { SELF } from "cloudflare:test";
 import { describe, expect, it } from "vitest";
+import type { Env } from "./env.js";
+import worker from "./index.js";
 
 // Valid RFC 6750 b64token — any non-empty ASCII token works for these tests
 // because `extractBearer` only validates shape, not value. Django's the one
@@ -33,6 +35,20 @@ describe("worker routing", () => {
     expect(res.status).toBe(200);
     expect(res.headers.get("content-type")).toBe("text/plain; charset=utf-8");
     expect(await res.text()).toBe("dev-stub-not-a-real-challenge-token");
+  });
+
+  it("GET /.well-known/openai-apps-challenge returns 404 when the token is unset", async () => {
+    // The "never satisfy an un-issued challenge" security property: with
+    // no `OPENAI_APPS_CHALLENGE_TOKEN` binding, the route must not answer
+    // and instead falls through to the catch-all 404. `SELF.fetch` always
+    // carries the configured stub, so we invoke the handler directly with
+    // an env lacking the token to exercise the unset path.
+    const env = { DJANGO_BASE_URL: "http://localhost:8000" } as Env;
+    const res = await worker.fetch(
+      new Request("https://example.com/.well-known/openai-apps-challenge"),
+      env,
+    );
+    expect(res.status).toBe(404);
   });
 
   // We deliberately do not alias the OIDC discovery path. ChatGPT's App
--- a/workers/src/index.test.ts
+++ b/workers/src/index.test.ts
@@ -20,6 +20,21 @@ describe("worker routing", () => {
     expect(res.status).toBe(404);
   });
 
+  it("GET /.well-known/openai-apps-challenge returns the configured token as plain text", async () => {
+    // OpenAI's connector-directory domain-verification flow GETs this
+    // URL during submission and matches the response body verbatim
+    // against the token shown in their dashboard. The token comes
+    // from the `OPENAI_APPS_CHALLENGE_TOKEN` binding (per env, in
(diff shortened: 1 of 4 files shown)


jed326 (approved):

workers/src/index.test.ts
```diff
@@ -20,6 +20,21 @@ describe("worker routing", () => {
...
+    // from the `OPENAI_APPS_CHALLENGE_TOKEN` binding (per env, in
+    // `wrangler.jsonc`); for the test runner the top-level `vars`
+    // entry stubs it to a known string.
+    const res = await SELF.fetch(
+      "https://example.com/.well-known/openai-apps-challenge",
+    );
+    expect(res.status).toBe(200);
+    expect(res.headers.get("content-type")).toBe("text/plain; charset=utf-8");
+    expect(await res.text()).toBe("dev-stub-not-a-real-challenge-token");
+  });
```
robertabbott, line 38: The new test covers the 200/token-present path. The behavior the PR description leans on for safety — "when unset, return 404 so an unrelated environment never satisfies a verification it wasn't issued" — is the unset→404 branch (`index.ts:67`), and it has no coverage. Since that's the actual security property, is it worth a second case that asserts 404 when the binding is absent/empty (e.g. via a per-test `vars`/env override in the workers pool)?

workers/src/index.ts
```diff
@@ -52,6 +52,30 @@ export default {
...
+    // `OPENAI_APPS_CHALLENGE_TOKEN`; when unset, return 404 so an
+    // unrelated environment never satisfies a verification it wasn't
+    // issued. Plain text response — no CORS wrapper needed (OpenAI
+    // hits this server-to-server, not from a browser).
+    if (
+      request.method === "GET" &&
+      url.pathname === "/.well-known/openai-apps-challenge"
+    ) {
+      const token = env.OPENAI_APPS_CHALLENGE_TOKEN;
+      if (token === undefined || token === "") {
```
robertabbott: When the token is unset this returns a 404 that's byte-for-byte identical to the fallthrough 404 at line 125. Dropping this branch and gating on `if (token)` would land any unset request on the same fallthrough. Is the explicit branch here purely for documenting the "never satisfy an un-issued challenge" intent at the route, or is a behavioral difference intended?

robertabbott (approved):

jfassio:
Addressed 2 comments from this review round (simplified the unset→404 branch, added unset-path test coverage). @claude-address

workers/src/index.ts
jfassio: Good catch — there was no behavioral difference; the explicit branch returned a 404 identical to the catch-all. Simplified to gate the route on `env.OPENAI_APPS_CHALLENGE_TOKEN` being present and let an unset/empty binding fall through to the catch-all 404. Kept a comment documenting the "never satisfy an un-issued challenge" intent so the reasoning isn't lost. (8668027)

workers/src/index.test.ts
jfassio: Added a dedicated test for the unset→404 path. `SELF.fetch` always carries the configured stub, so the test invokes the handler directly (`worker.fetch(req, env)`) with an env lacking the token and asserts 404 — covering the security property directly. (8668027)

PR #97, opened 2026-06-28
feat(mcp): single-source tool schemas from the sdk OpenAPI spec (contract-guarded)

jed326:
## What this does

Stops hand-maintaining the MCP tool schemas and **generates them from the same `sdk`-audience OpenAPI spec the SDK consumes.** Hand-maintained schemas were the source of recurring MCP↔backend drift — the `sources` default divergence, the missing `contents.mode`, and one-off manual resyncs like #95. After this, the contract is single-sourced and drift fails the build instead of shipping silently.

Design follows the industry-standard **generate-then-curate** pattern (checked against Speakeasy / Neon / FastMCP): **the spec owns structure; the curated, LLM-ergonomic layer stays hand-authored** (rich tool descriptions, the flat `sources` facade). This is Part A of the effort; the monorepo sync leg that auto-delivers spec updates here is the companion PR **TakoData/tako#26951**.

Spec & plan (in the monorepo PR): `docs/superpowers/specs/2026-06-27-tako-mcp-contract-generation-design.md`, `docs/superpowers/plans/2026-06-28-tako-mcp-contract-generation.md`.

## How it works

- **`openapi/sdk.yaml`** is vendored (and, once #26951 lands, auto-synced by the org GitHub App).
- **`workers/scripts/gen-schemas.ts`** (json-schema-to-zod) → **`workers/src/generated/schemas.ts`**: one Zod schema + type per `components.schemas` entry, intra-doc `$ref`s as `z.lazy(() => X)`. A `liftRefSiblings` pre-pass preserves OpenAPI 3.1 `$ref`+sibling keywords (`default` / `description` / optionality) that the bare-`$ref` parser would otherwise drop.
- **Each tool composes the generated contract with its hand-authored `DESCRIPTION` + handler:**
  - **Outputs:** taken from the generated response schema where shapes match, or — where the MCP normalizes the wire (search / answer / agent / visualize / contents) — a hand-written facade **guarded** by `Response.safeParse(rawWire)` before mapping, so backend response drift throws instead of silently mapping to nulls.
  - **Faceted inputs** (the flat `sources` / `include_contents` facade) stay hand-authored; each `build*Body` reshape is typed `satisfies z.input<typeof XRequest>` **and** its nested `sources` object is typed against `z.input<…>["sources"]`, so a renamed/added contract key, a new required field, or a changed enum **breaks compilation at the reshape site.**
  - One documented divergence: `tako_contents.mode` defaults to `inline` (MCP ergonomics) vs the backend's `url`.
- **Guards that replace the old manual parity doc:** `schemas:check` + `registry:check` in CI and `predeploy:*`; a curated tool **allowlist** + **description-completeness** assertion in `gen-registry.ts` that fails closed (adding/removing a tool file without updating the allowlist breaks the build; `classify` is intentionally excluded — it's an API operation, not a tool); `regenerate.yml` regenerates and re-verifies on every synced-spec PR.

## Review

Built TDD task-by-task with per-task + a final whole-branch review, then a pre-merge `pr-review` pass. Findings addressed in-branch:
- **Tightened the `sources` request-guard** in `tako_answer` / `tako_search` — it was structurally cosmetic on `sources` (a `Record<string, …>` local accepted any key), the exact field the effort exists to protect. Now typed against the contract.
- Deduped the `appUiResource` widget closure into a shared `_chart_widget.ts` helper.
- Distinct error strings for wire-guard vs post-map failures; misc comment cleanup.
- Earlier review caught and fixed the keystone `$ref`-sibling drop.

### Smoke test (local Worker → staging backend)

Verified end-to-end by running the branch's Worker code locally (`npm run dev:staging` → `localhost:8787`, proxying to `staging.trytako.com`) and driving it with the MCP smoke client (`scripts/smoke.ts`). All steps green:

```
✓ /health → 200 "ok"
✓ initialize → tako-mcp 0.5.0
✓ tools/list → 6 tools (get_credit_balance, tako_agent, tako_answer, tako_contents, tako_search, tako_visualize)
✓ tako_search → MCP Apps widget at ui://tako/embed/chart (text/html;profile=mcp-app)
✓ tako_search "US GDP" → 1 cards
✓ tako_answer "US GDP" → answer
✓ tako_contents {url} → inline data (65 rows)
✓ tako_contents {url, mode:"url"} → download_url present
✓ get_credit_balance → details.credit_balance
✓ tako_visualize → card created
✅ smoke passed
```

This directly confirms the two findings with runtime surface against a real backend: the widget assertion exercises the deduped `buildChartAppUiResourceFromOutputPubId` helper, and the green `tako_search` / `tako_answer` calls confirm the retyped `sources` reshape still produces a body the backend accepts. (The live surface shows 6 tools — the unsplit `tako_agent` — because the agent split is gated by client user-agent; the smoke client gets the unsplit tool, as expected.)

## Lines changed (by category)

- **Logic** (~+1.0k): `scripts/gen-schemas.ts`, `scripts/gen-registry.ts`, the 5 `tako_*.ts` tools, `src/generated/schemas.ts`, the shared `_chart_widget.ts` helper, `regenerate.yml`, `workers-ci.yml`, `package.json`.
- **Tests** (~+310): `gen-schemas.test.ts`, `gen-registry.test.ts`, and the 5 tool `*.test.ts` suites.
- **Generated / vendored / config** (~+2.2k): `openapi/sdk.yaml`, `registry/server.json`, `package-lock.json`.

Full suite **214/214 passing**; typecheck (3 tsconfigs) + `schemas:check` + `registry:check` green.

## Merge order / prerequisites

- **Merge this PR first**, before the monorepo sync-leg PR (#26951) — it must land `openapi/sdk.yaml` + `regenerate.yml` so the monorepo's auto-sync PR has somewhere to regenerate.
- #26951 additionally requires the **`OPENAPI_SYNC_APP` GitHub App installed on `TakoData/tako-mcp`** (Contents + PRs: write) before its sync leg can fire.
- Version bump / publish remains a manual step when ready.

https://claude.ai/code/session_01MGrCGxySeRXqq1rC5QTdaT


.github/workflows/regenerate.yml
```diff
@@ -0,0 +1,63 @@
...
+        working-directory: .
+        env:
+          HEAD_REF: ${{ github.head_ref }}
+        run: |
+          git config user.name "tako-bot"
+          git config user.email "bot@tako.com"
+          git add workers/src/generated/schemas.ts workers/src/tools/_registry.ts registry/server.json
+          if ! git diff --cached --quiet; then
+            git commit -m "chore: regenerate schemas + registry from synced spec"
+            git push origin "HEAD:$HEAD_REF"
```
robertabbott, line 71: This job checks out the synced-spec PR branch and pushes the regenerated `schemas.ts` / `_registry.ts` / `server.json` back with `secrets.GITHUB_TOKEN`. GitHub deliberately does not trigger `on: pull_request` / `on: push` workflows for commits pushed by the default `GITHUB_TOKEN` (anti-recursion). So `workers-ci.yml` won't run against the bot's regen commit — it only ran against the App's original spec-only commit. If branch protection requires the `workers-ci` checks, the PR's head SHA will be missing them and can't merge; if it doesn't, the regenerated code ships verified only by this job's own `Verify` step. Is that the intent (rely on the in-job `typecheck`/`test`/`*:check`), or should this push use the `OPENAPI_SYNC_APP` token / a PAT so the standard checks re-run on the final SHA?

workers/src/tools/tako_contents.ts
```diff
@@ -12,24 +12,25 @@
...
 
-const inputSchema = z.object({
-  url: z
-    .string()
-    .min(1)
-    .describe("The result URL to fetch content for (a Tako card URL → CSV; any other URL → page text)."),
-  mode: z
-    .enum(["inline", "url"])
+// Generated contract, with the one documented MCP divergence: default mode → inline.
+const inputSchema = ContentsRequest.extend({
```
robertabbott: `tako_contents` builds its caller-facing `inputSchema` from `ContentsRequest.extend(...)`, whereas `tako_search`/`tako_answer`/`tako_agent`/`tako_visualize` hand-author the flat input and use the contract only for the `satisfies z.input<…>` body type. Because `.extend` spreads the whole contract object, any field later added to `ContentsRequest` in the synced spec will silently join this tool's MCP input surface with no curation step — the opposite of the "spec owns structure; the curated layer stays hand-authored" principle stated in the description. Given the tool surface is otherwise guarded by `MCP_TOOL_ALLOWLIST`, is auto-adopting new request fields here intended, or would picking the fields explicitly (`ContentsRequest.pick({ url: true }).extend({ mode })`) keep the input curated the same way the others are?

workers/src/tools/tako_contents.ts
```diff
@@ -12,24 +12,25 @@
...
+  mode: ContentsDeliveryMode
```
robertabbott, line 29: The previous hand-written schema had `url: z.string().min(1)`; the generated `ContentsRequest.url` is a bare `z.string()` (no `minLength` in the spec). So an empty-string `url` now passes MCP-layer validation and is forwarded to `/api/v1/contents/`, where it was previously rejected up front. The backend will presumably 400/404 it anyway, but the tighter local validation is gone. Worth re-adding `.min(1)` in the `.extend({...})` alongside `mode` to preserve the prior behavior, or is deferring to the backend acceptable here?

workers/src/tools/tako_search.ts
```diff
@@ -129,8 +162,8 @@ const tako_search = {
     return buildSearchOutput(
       cards.data,
       webResults.data,
-      data.request_id ?? "",
-      data.contents_total_cost ?? 0,
+      wire.request_id,
```
robertabbott, line 165: The generated `SearchResponse` and `AnswerResponse` declare `request_id: z.string()` (required, no default) and `AnswerResponse.answer: z.string()` (required). The old code degraded a missing field to a default (`data.request_id ?? ""`, `data.answer ?? ""`); now `*.safeParse(data)` fails and the whole call throws "unexpected wire shape" if the backend ever returns a 200 without them. That's the intended fail-closed contract behavior — just confirming these are guaranteed present on every successful response (including zero-result `tako_search` and any answer-less `tako_answer` case), since this turns a previously-tolerated omission into a hard error. (Same pattern at `tako_answer.ts:117`.)

.github/workflows/regenerate.yml
jed326: Intended, with a caveat. The `GITHUB_TOKEN` push is deliberate — it's what prevents an infinite regen→regen loop, and the in-job `Verify` step (`schemas:check` / `registry:check` / `typecheck` / `test`) runs the exact same checks `workers-ci.yml` would, so the regenerated code *is* verified before the commit lands.

The one real gap you're pointing at is the status check on the final SHA. I checked: `main` currently has no branch protection and no rulesets, so there are no required status checks today — the regen commit isn't blocked from merging. So this works as-is right now.

The forward path, if/when we add branch protection that requires `workers-ci`: switch this push to the `OPENAPI_SYNC_APP` installation token so the standard checks re-fire on the regen commit. That needs the App's ID + private key as secrets in this repo (they currently live only in the monorepo), so I'm deferring it until branch protection actually requires it rather than wiring the extra secret now. I'll add a comment in the workflow noting the intentional `GITHUB_TOKEN` choice + this upgrade path.

workers/src/tools/tako_search.ts
jed326: Confirmed — this is the intended fail-closed contract behavior. I checked the generated schemas: `SearchResponse.request_id`, `AnswerResponse.request_id`, and `AnswerResponse.answer` are all `z.string()` with no `.optional()`/`.default()`, i.e. spec-required. They're required in the backend's own Pydantic models too, so they're guaranteed on every 200 — `request_id` is always minted (including zero-result `tako_search`, which returns an empty `cards` array but still a `request_id`), and the answer endpoint always synthesizes an `answer` string (the 'no answer' case is still a non-empty string, not an omission).

So the old `?? ""` degradations were papering over a case that can't legitimately occur on a success response; turning that into a hard error is exactly what we want — if any of these ever go missing on a 200, that's a real backend contract break we want surfaced loudly rather than silently coerced to an empty string. Same reasoning for the `tako_answer.ts` site.

robertabbott (approved):

PR #131, opened 2026-07-17
fix: expose max_rows in tako_contents (inline mode was silently capped at 20 rows)

25eliu:
## Problem

`tako_contents` in **inline** mode silently returned only the **free 20-row preview** of a Tako card. The generated `ContentsRequest` contract has a `max_rows` field, but the tool never exposed it — so a caller had no way to fetch more rows inline.

Worse, the response reported truncation correctly while the description lied about the cap. Reproduced live against the deployed connector (silver-price card):

```
total_rows: 300, truncated: true    ← but `data` contained exactly 20 rows
```

The tool description claimed *"CSV capped at 1000 rows"*, so an agent reads 20 rows and believes it has the whole (≤1000-row) dataset. `url` mode was unaffected (full download, no cap).

## Fix

- **Expose `max_rows`** from the contract — parse-don't-cast: reuse `ContentsRequest.shape.max_rows` so it keeps the contract's `.gte(1)` guard, re-described for the MCP surface. Omit → backend's 20-row free default; raise up to the 2,000-row ceiling (billed per 1,000 rows beyond the free 20). Flows straight through `body = input`; no handler logic change.
- **Fix the misleading "1000 rows" claim** in the tool description, `mode` help, file docstring, and output-schema comment — real numbers are **20 (free default) / 2,000 (ceiling)**.
- **Correct the `cost` comment** — a Tako-card CSV is metered (per-export baseline + per-1,000-row rate beyond the free 20), not free (live call billed `0.001`).
- **Regenerate `registry/server.json`** for the new param + updated descriptions.

## Test plan

- [x] `npx vitest run` — 247/247 pass, incl. 3 new tests (max_rows exposed & optional, `.gte(1)` guard, body passthrough)
- [x] `npm run typecheck` — clean (all 3 tsconfigs)
- [x] `npm run registry:check` — in sync
- [ ] Live end-to-end with `max_rows` can only be verified **after deploy** — the currently-deployed tool doesn't expose the param. Passthrough is covered by unit tests; the semantics are the synced contract's.

## Notes
- Scoped to `tako_contents` only (branched off `main`, not the graph-API work).
- Separately observed while reproducing: `tako_search` is currently throwing `"unexpected shape"` in prod (likely stale generated schemas). **Out of scope for this PR** — flagging for a follow-up.

### DIFF
--- a/workers/src/tools/tako_contents.ts
+++ b/workers/src/tools/tako_contents.ts
@@ -2,11 +2,13 @@
  * `tako_contents` — fetch the downloadable content behind a result URL.
  * Wraps `POST /api/v1/contents`. A Tako card URL resolves to a CSV of the
  * card's data; any other URL resolves to the page's extracted text. One URL
- * per call. `mode` controls delivery: "inline" (default) returns the content
- * in the response body (for a Tako card, only the free 20-row preview unless
- * `max_rows` is raised — up to a 2000-row ceiling — with total_rows/truncated;
- * or web text) so the model can read it directly; "url" returns a short-lived
- * presigned download URL instead. The "inline" default is an intentional
+ * per call. A Tako card CSV is capped at a 20-row free default in BOTH modes;
+ * `max_rows` raises that up to a 2000-row ceiling — there is no uncapped export.
+ * `mode` controls only delivery, not the row count: "inline" (default) returns
+ * the (capped) content in the response body — total_rows/truncated report the
+ * true size — so the model can read it directly; "url" returns a short-lived
+ * presigned download URL to the same capped CSV instead. The "inline" default is
+ * an intentional
  * MCP-ergonomics divergence from the backend default ("url") — an agent almost
  * always wants to read the data, not hand back a link.
  */
@@ -17,7 +19,7 @@ import { ContentsDeliveryMode, ContentsRequest, ContentsResponse } from "../gene
 import type { ToolModule } from "./types.js";
 
 const DESCRIPTION =
-  "Fetch the underlying data behind a result URL — a Tako card URL yields a CSV of the card's data; any other URL yields the page's extracted full text. Pass a single `url` (a TakoCard.webpage_url or a web-result URL). `mode` controls delivery: `inline` (default) returns the content directly in the response so you can read and reason over it — for a Tako card this returns only the free 20-row preview unless you raise `max_rows` (up to 2,000; billed per 1,000 rows beyond the free 20), so always check `total_rows` / `truncated` to know if it's partial; `url` instead returns a short-lived presigned `download_url` (no row cap), for handing the user a download/embed link or for large datasets you don't need to read yourself. Use `inline` when you need the numbers; use `url` when the user just wants the file or the full dataset.";
+  "Fetch the underlying data behind a result URL — a Tako card URL yields a CSV of the card's data; any other URL yields the page's extracted full text. Pass a single `url` (a TakoCard.webpage_url or a web-result URL). A Tako card CSV is capped at a 20-row default in both delivery modes; raise `max_rows` (up to 2,000) to get more — there is no uncapped export. `mode` controls only how the content is delivered, not how many rows come back: `inline` (default) returns the CSV/web text directly in the response so you can read and reason over it — always check `total_rows` / `truncated` to know if it's partial; `url` returns a short-lived presigned `download_url` to the same (capped) file, for handing the user a download/embed link or when you don't need to read it inline. Use `inline` when you need the numbers; use `url` when the user just wants the file.";
 
 // Curate the input from the contract explicitly: `.pick` only the fields we
 // expose (so a new field added to ContentsRequest in the synced spec does NOT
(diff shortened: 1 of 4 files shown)


workers/src/tools/tako_contents.ts
```diff
@@ -16,19 +17,26 @@ import { ContentsDeliveryMode, ContentsRequest, ContentsResponse } from "../gene
...
+  "Fetch the underlying data behind a result URL — a Tako card URL yields a CSV of the card's data; any other URL yields the page's extracted full text. Pass a single `url` (a TakoCard.webpage_url or a web-result URL). `mode` controls delivery: `inline` (default) returns the content directly in the response so you can read and reason over it — for a Tako card this returns only the free 20-row preview unless you raise `max_rows` (up to 2,000; billed per 1,000 rows beyond the free 20), so always check `total_rows` / `truncated` to know if it's partial; `url` instead returns a short-lived presigned `download_url` (no row cap), for handing the user a download/embed link or for large datasets you don't need to read yourself. Use `inline` when you need the numbers; use `url` when the user just wants the file or the full dataset.";
```
noahjax: **[Issue]** This description commits the tool to CSV-only ("a Tako card URL yields a CSV"). The backend `content_format` param supports `json_records` (→ `records`) and `json_compact` (→ `dataset`) too, and both are in the synced contract — but neither is reachable through the MCP. Since you're curating the input surface anyway, let's expose `content_format` so agents can request structured JSON, not just CSV text.

workers/src/tools/tako_contents.ts
```diff
@@ -16,19 +17,26 @@ import { ContentsDeliveryMode, ContentsRequest, ContentsResponse } from "../gene
...
   // an empty-string url is rejected at the MCP layer instead of hitting the API.
   url: ContentsRequest.shape.url.min(1),
   mode: ContentsDeliveryMode
     .default("inline")
-    .describe('Delivery mode: "inline" (default) returns the content in the response body (CSV capped at 1000 rows, with total_rows/truncated; or web text) so you can read it directly; "url" returns a short-lived presigned download_url (no row cap).'),
+    .describe('Delivery mode: "inline" (default) returns the content in the response body (for a Tako card, only the free 20-row preview unless you raise max_rows; with total_rows/truncated; or web text) so you can read it directly; "url" returns a short-lived presigned download_url (no row cap).'),
+  // Expose the contract's row cap so an agent can pull more than the 20-row free
+  // preview inline. parse-don't-cast: reuse the generated shape (carries the
+  // .gte(1) guard) and just re-describe it for the MCP surface.
+  max_rows: ContentsRequest.shape.max_rows.describe(
```
noahjax: **[Issue]** "Ignored in url mode (full download)" contradicts backend `main`: `build_content` (`generation.py:346-352`) defaults an omitted `max_rows` to `CSV_FREE_ROWS` (20) and passes it into `build_card_csv` **in url mode**, which serializes only the capped `window` (`generation.py:563-575`). The URL tests assert this (`generation_unit_test.py:45-103`). So `max_rows` is *not* ignored in url mode, and url mode is *not* a full download — it's 20 rows by default, with no uncapped path through `/contents` at all. Please verify the deployed backend, then fix this line (and the same "no row cap"/"full dataset" phrasing at line 33 and in the DESCRIPTION/docstring).

workers/src/tools/tako_contents.ts
```diff
@@ -43,11 +51,12 @@ const outputSchema = z.object({
...
   source_url: z.string(),
-  // USD charged for this artifact (web text is metered ~$1/1k pages; Tako-card
-  // CSV is free → 0). Surfaced so the agent can report what the call cost.
+  // USD actually charged for this artifact. Web text is metered per page; a
+  // Tako-card CSV bills a per-export baseline plus a per-1,000-row rate on rows
+  // beyond the free 20-row allowance. Surfaced so the agent can report the cost.
   cost: z.number(),
-  // Inline content — populated in "inline" mode (CSV text capped at 1000 rows, or
-  // web page text), null in "url" mode. total_rows/truncated describe CSV truncation.
+  // Inline content — populated in "inline" mode (CSV text, up to max_rows / the
```
noahjax: **[Issue]** The `outputSchema` (and the handler, which reads only `item.data`) has no slot for `records` or `dataset`, so even if `content_format` were added, the JSON payloads would be dropped to null. To surface the `json_records`/`json_compact` formats end-to-end, add `records`/`dataset` (and consider `export_pricing`) here and map them from the wire `ContentItem`.

noahjax (changes requested):
Solid, well-tested fix for the inline row-cap gap — the parse-don't-cast reuse of `ContentsRequest.shape.max_rows` and the honesty pass on the descriptions are exactly right.

Two things before merge, both about surfacing the *full* `/contents` output-format contract (which the synced `generated/schemas.ts` already carries in full):

1. **The MCP still exposes only CSV.** `content_format` (`json_records` → `records`, `json_compact` → `dataset`) is never requestable, and the `outputSchema`/handler have no slot for `records`/`dataset` — so structured JSON output is unreachable. `quote_only` + `export_pricing` (price an export before paying) are likewise unsurfaced. Let's add these here so the tool covers the whole contract in one change.
2. **The "url mode = full download, no row cap" claim looks wrong against backend `main`.** `views.py` → `build_content` defaults an omitted `max_rows` to 20 and applies it in url mode too (`generation.py:563-575`; asserted in `generation_unit_test.py:45-103`). Please confirm the prod deploy state — your live test may have hit an older backend — and correct the wording either way. As written, an agent told to "use url for the full dataset" hands back a 20-row file.

workers/src/tools/tako_contents.ts
noahjax: @25eliu I can fix this on the tako side of things so we have better max rows behavior

workers/src/tools/tako_contents.ts
25eliu: Done — exposed `content_format` (`csv` default | `json_records` | `json_compact`), reusing `ContentsRequest.shape.content_format`. Description no longer commits to CSV-only. (b42e8f7)

workers/src/tools/tako_contents.ts
25eliu: Done — added `records` (json_records → row objects) and `dataset` (json_compact → TakoDataset, reusing the generated schema) to the outputSchema, and the handler now maps `item.records`/`item.dataset`. Exactly one of data/records/dataset is populated per call. Left `quote_only`/`export_pricing` for a separate follow-up. (b42e8f7)

noahjax (approved):

PR #132, opened 2026-07-17
feat: Tako data-graph tools (search/related/node) + node pinning; fix search 'unexpected shape' outage

25eliu:
## Summary

Adds the **Tako data-graph** capability to the MCP server as three thin primitive tools, wires graph-node pinning into the two retrieval tools, and — while diagnosing along the way — **fixes a live prod outage** in `tako_search`/`tako_answer`.

### New tools
- **`tako_graph_search`** — resolve a name → graph node(s) (`GET /api/beta/graph/search`)
- **`tako_graph_related`** — a node → what it connects to, with **multi-`q` parallel fan-out** that unions/dedupes results by node id (`GET /api/beta/graph/related`)
- **`tako_graph_node`** — hydrate a node id → full detail (`GET /api/beta/graph/node/{id}`)

### Extended tools
- **`tako_search`** + **`tako_answer`** gain `node_ids` (max 20) + `strict`, mapped into `sources.data` so agents can pin resolved graph nodes into retrieval
- **`tako_search`** now surfaces each card's graph `nodes` (id/name/type), closing the discover→pin loop

The workflow guidance (parallel-for-narrow, overview-as-coverage-map, multi-`q`-for-broad, "graph is free") lives in the tool descriptions, since raw MCP clients don't load the `tako-graph-agent` skill.

### 🚑 Prod outage fixed (found via live diagnosis)
`tako_search`/`tako_answer` were throwing **"unexpected shape" on every query** in prod. Root cause: the hand-written `resultContentSchema` still required a `format` field, but the backend renamed it to `content_format` (the generated schema was updated in `d21b000`; this hand-written one was missed). A `content` preview object rides on every result even when `include_contents=false`, so every result tripped the handler's second-stage guard. Fixed by renaming the field and making `resultContentSchema` fully optional/nullable (it's a read-only response the tools only surface for the model). Verified end-to-end against live prod: every query threw before, all succeed after.

## Design & architecture
- **Thin primitives + guidance** (not a server-side orchestrator) — the agent keeps control of parallelism/node-selection.
- **Wire-guard / facade split:** each handler `safeParse`s the raw response against the generated schema, then re-validates through a flat hand-written output facade (parse-don't-cast, no `z.lazy` in advertised JSON schema).
- Graph endpoints are GET → left slash-less, the robust choice under Django `APPEND_SLASH`.
- No `django.ts` transport changes; no `openapi/sdk.yaml`/schema edits (the graph contract already existed there).

Design spec + implementation plan: `docs/superpowers/specs/` and `docs/superpowers/plans/`.

## Test plan
- [x] `npm run test` — 248 passing
- [x] `npm run registry:check` — clean (11 tools)
- [x] `npm run schemas:check` — clean
- [x] Graph tool schemas verified against live API reference docs (graph-search + graph-related match exactly)
- [x] Search hotfix verified end-to-end against **live prod** (all queries OK after fix)
- [x] Regression tests for the `content_format` fix in both `tako_search` and `tako_answer`, built from the real live payload

## Notes
- The parallel `tako_contents` `max_rows` work lives on its own branch (`fix/contents-max-rows-inline`) and is **not** included here.


docs/superpowers/specs/2026-07-17-graph-api-integration-design.md
```diff
@@ -0,0 +1,264 @@
...
+Maps to `GET /api/beta/graph/search`.
+
+**Input:**
+
+| field | type | notes |
+|---|---|---|
+| `q` | string, required | search text, min 2 chars |
+| `types` | `"entity" \| "metric"` | single choice; sent as the API's comma-separated string. Decide "thing vs measure" up front — don't mix. |
+| `label` | NER enum, optional | boost (not a filter): `PERSON, ORG, GPE, LOC, PRODUCT, EVENT, LANGUAGE, MONEY, METRIC, STOCK_TICKER, WEBSITE`. Omit to let `infer_label` run. |
+| `infer_label` | boolean, optional, default true | auto-detect labels from `q` |
```
noahjax, line 74: Torn on whether we should surface this to mcp or just default it to false; since mcp's are always going to be used by agents, they can always just provide a label instead of relying on us to infer one

docs/superpowers/specs/2026-07-17-graph-api-integration-design.md
```diff
@@ -0,0 +1,264 @@
...
+
+Maps to `GET /api/beta/graph/related`.
+
+**Input:**
+
+| field | type | notes |
+|---|---|---|
+| `node_id` | string, required | opaque public id from a search/related result or a card node |
+| `relation` | string, optional | relation key to drill: `metrics`, `entities`, `siblings`, `part_of`, `members`, or `rel:<phrase>`. Omit for the overview. |
+| `q` | `string \| string[]`, optional | case-insensitive substring filter on name+aliases. See multi-`q` below. |
```
noahjax, line 96: Instead of allowing `string[]` as an input it may be simpler to just tell the agent to call related multiple times if it has multiple search strings. This avoids confusing label/cursor/limit logic when multiple query strings are provided

workers/src/tools/_search_results.ts
```diff
@@ -46,6 +54,20 @@ export const takoCardSchema = z
...
+    // Graph nodes (entities/metrics) this card was built from, returned by the
+    // backend by default. Slim shape (id/name/type) — pass these ids into
+    // sources.data.node_ids to pin the same nodes in a follow-up search, or
+    // hydrate with tako_graph_node for full detail.
+    nodes: z
+      .array(
+        z.object({
+          id: z.string(),
+          name: z.string(),
+          type: z.enum(["metric", "entity"]),
```
noahjax: **[Issue]** This constrains the card node `type` to a hand-written strict enum — the same drift-prone strictness that caused the outage this PR fixes.

When the backend adds a third `GraphNodeType`, the **generated** `SearchResponse` schema gets regenerated to accept it (so the wire guard `SearchResponse.safeParse` passes), but this **hand-written** enum won't be updated — so `z.array(takoCardSchema).safeParse(wire.cards)` fails and *every* `tako_search`/`tako_answer` response throws "unexpected shape". That's the identical mechanism to `format`→`content_format`: generated schema updated, hand-written facade missed.

Nothing branches on this field (it's read-only, surfaced to the model), so loosen `type` to `z.string()` — matching the principle stated three lines up in this same file, and how `graphNodeSchema` in `_graph.ts` keeps `subtype`/`label` loose.

workers/src/tools/tako_search.ts
```diff
@@ -72,6 +72,19 @@ const inputSchema = z.object({
...
+    .max(20)
+    .optional()
+    .describe(
+      "Graph node ids (from tako_graph_search / tako_graph_related, or a card's nodes) to PIN into the Tako data source. Pinned nodes get a strong retrieval boost. Max 20. Applies only to the 'data' source.",
+    ),
+  strict: z
+    .boolean()
+    .default(false)
+    .describe(
+      "Documented to return only cards matching a pinned node (requires node_ids). NOTE: not reliably enforced upstream — rely on the pin boost, not strict, for hard exclusion.",
```
noahjax: **[Issue]** This description is inverted. Per the generated contract itself — `DataSourceSettings.strict`: *"When true, only return data cards that match at least one node in node_ids (which must then be non-empty). When false (default), pinned nodes are preferred but organic results still return."* — and confirmed with the team, `strict` **does** hard-filter to pinned nodes.

Telling the agent it's "NOT reliably enforced upstream — rely on the pin boost, not strict, for hard exclusion" steers it away from a working exclusion feature. Please rewrite to describe strict as the hard-filter it is (true = only cards matching ≥1 pinned node, requires non-empty `node_ids`; false = pinned preferred, organic still returns).

The same inverted copy appears in 5 places: this param description, the `tako_search` tool description (line 35), and the `tako_answer` equivalents (lines 9 and 44) — then regenerate `registry/server.json`.

workers/src/tools/_graph.ts
```diff
@@ -0,0 +1,235 @@
...
+  DjangoUnauthorizedError,
+} from "../django.js";
+import type { GraphNode, GraphRelatedResponse } from "../generated/schemas.js";
+
+/** Advertised graph-node facade. subtype/label are stringified enums on the
+ *  wire; the facade keeps them as loose strings so a new enum value never
+ *  breaks the advertised contract. */
+export const graphNodeSchema = z.object({
+  id: z.string(),
+  type: z.enum(["metric", "entity"]),
```
noahjax: **[Suggestion]** Same inconsistency as the card `nodes` finding, lower blast radius (a per-call error rather than a global search outage). This file's own comment says the facade keeps enums as loose strings "so a new enum value never breaks the advertised contract," but `type` is constrained to a strict enum. Consider `z.string()` for symmetry, or a short comment noting `type` is intentionally kept strict.

noahjax (commented):
Reviewed the code (skimmed the design docs). Really clean PR — thin primitives, the wire-guard/facade split is applied consistently, `mergeRelatedResponses` and `graphErrorMessage` are well-covered by tests, and the `content_format` outage fix is well-motivated. CI is green and nothing here is currently broken.

Two things I'd fix before merge:

1. **Latent outage (same class this PR fixes):** the new card `nodes[].type` uses a hand-written strict enum. When the backend adds a third node type, the *generated* `SearchResponse` will be regenerated to accept it (wire guard passes) but this hand-written enum won't be — and `z.array(takoCardSchema).safeParse(wire.cards)` failing throws "unexpected shape" on *every* search/answer response. Exactly the `format`→`content_format` drift mechanism. One-word fix (loosen to `z.string()`).
2. **`strict` docs are inverted:** per the generated `DataSourceSettings.strict` contract (and confirmed with the team), `strict` *does* hard-filter to pinned nodes. The current copy tells agents it doesn't work, steering them away from a working exclusion feature.

Inline for both, plus one lower-priority consistency suggestion. Neither is a hard blocker, so flagging as comments rather than requesting changes.

docs/superpowers/specs/2026-07-17-graph-api-integration-design.md
25eliu: I think surface to mcp and default to true makes sense since non-agents also have a graph api use case and the mcp should def include this context

workers/src/tools/_search_results.ts
25eliu: Fixed — loosened card `nodes[].type` to `z.string()` in `_search_results.ts` with a comment explaining exactly this (nothing branches on it; hard-coding the enum reintroduces the drift failure the content_format fix removed). Good catch — I'd added the same class of bug this PR fixes. (da03677)

workers/src/tools/_graph.ts
25eliu: Done — `graphNodeSchema.type` is now `z.string()` too, and the facade comment now covers type/subtype/label together. Consistent with the file's stated principle. (da03677)

workers/src/tools/tako_search.ts
25eliu: Fixed, and thank you for the correction — you're right. I **verified it live on prod** to be sure: with a broad query and NVIDIA pinned, `strict:true` returned 0 cards (organic non-matching results excluded) vs 2 with `strict:false`, and empty `node_ids` + `strict:true` → 400. So strict genuinely hard-filters. My original copy came from a stale skill note claiming staging didn't enforce it. Rewrote all inverted copy — `tako_search` param + description, `tako_answer` param + description, and the design spec — describing strict as the hard filter it is (true = only cards matching ≥1 pinned node, requires non-empty node_ids; false = pinned preferred, organic still returns). Regenerated server.json. (da03677)

docs/superpowers/specs/2026-07-17-graph-api-integration-design.md
25eliu: Agreed — removed the `string[]` support. `q` is now a single string, and the description tells the agent to call `related` once per name-variant (graph calls are free). This drops the fan-out + `mergeRelatedResponses` + the cursor/limit-per-fan-out-call ambiguity you flagged. (da03677)

docs/superpowers/specs/2026-07-17-graph-api-integration-design.md
25eliu: Kept `infer_label` exposed, but took your point — strengthened the `label` copy (field + tool description) to actively encourage the agent to supply an explicit label when it can categorize the term (company→ORG, place→GPE, ...), with inference as the fallback for when it can't. (da03677)

noahjax (approved):

PR #136, opened 2026-07-19
feat: make the Tako agent an opt-in tool via the tools query parameter

25eliu:
## What

Makes the Tako **Answer Agent** an **opt-in** tool instead of a default one. It's enabled per-connection the standard MCP way — a `tools` query parameter on the server URL — mirroring how Exa exposes its optional tools (`?tools=agent_tools`):

```
https://mcp.tako.com/mcp?tools=agent
```

`agent` is an alias for the whole agent feature. With it enabled the server exposes the single-call `tako_agent` on Claude/unknown clients and automatically switches to the `tako_agent_start` / `tako_agent_wait` split pair on ChatGPT. Without it, none of the three appear.

| URL | Claude / unknown | ChatGPT |
|---|---|---|
| `.../mcp` | 8 defaults, **no agent** | 8 defaults, no agent |
| `.../mcp?tools=agent` | + `tako_agent` | + `tako_agent_start` / `tako_agent_wait` |

## Why

The agent is a slow (~30–90 s), opinionated, multi-step research tool — the right choice for some questions but not a sensible *default* sitting next to the fast one-shot tools. Making it opt-in keeps the default surface lean and lets callers turn the agent on deliberately.

## ⚠️ Breaking behavior change

Connectors that receive `tako_agent` by default **today** will stop seeing it until they append `?tools=agent` to their MCP URL. This is the intended product change. Existing installs that rely on the agent must update their connector URL to re-enable it. (Titled `feat:` for a patch bump — retitle to `feat!:` if you want release-please to cut a minor version instead.)

## How it works

- **`workers/src/tools/_optional.ts`** (new) — single source of truth: the `agent` alias → `[tako_agent, tako_agent_start, tako_agent_wait]` mapping, a derived `OPTIONAL_TOOL_NAMES` set, and a pure, tolerant `parseEnabledOptionalToolNames` parser (comma-split, trim, lowercase, dedupe; **only aliases** recognized; unknown tokens ignored, never fatal — a typo can't break `initialize`).
- **`workers/src/mcp.ts`** — `handleMcpRequest` reads `?tools=`, resolves it, and logs the raw value + resolved set for `wrangler tail`. `createMcpServer` gains an `enabledOptionalToolNames` option; the registration loop skips any tool in `OPTIONAL_TOOL_NAMES` unless enabled, **ahead of** the existing per-client filters so the right agent variant is still chosen.
- The generated registry (`registry/server.json`) still lists all 11 tools — discovery lists everything, the runtime filters per request (same precedent as the ChatGPT-split tools). No `gen-registry` change needed.

## Docs

- **README** — new "Enabling the Tako agent (opt-in)" section + cross-references, plus a broader accuracy pass (see below).
- **llms.txt / llms-full.txt / AGENTS.md** — document the opt-in and how to enable it.

### README accuracy pass (bundled)

While documenting the agent I audited the whole README against the code and fixed pre-existing inaccuracies:

- Documented the **graph tools** (`tako_graph_search` / `tako_graph_related` / `tako_graph_node`), which shipped but were never listed.
- Fixed `tako_search` / `tako_answer` **`sources` default** — it's both `data`+`web`, not the documented `["tako"]`; canonical values are `data`/`web`.
- Reconciled the self-contradictory OAuth **Prerequisites** ("mint a token" step vs "no longer required") and corrected a stale **`trytako.com` → `tako.com`**.
- Corrected the Architecture diagram's Django host (**`tako.com`**, not `api.tako.com`).
- Corrected the `tako_search` inline-render claim (inline widget on ChatGPT; Open-in-Tako link on Claude.ai, where the widget is suppressed).
- Removed a misleading reused screenshot + an unfilled screenshot placeholder; all remaining image references verified present.

## Testing

- **Unit** (`workers/src/tools/_optional.test.ts`, new) — parser: null/empty/whitespace, `agent`, case+whitespace, dupes, unknown token, raw-tool-name (alias-only), mixed list; plus `OPTIONAL_TOOL_NAMES` derivation.
- **Integration** (`workers/src/index.test.ts`) — updated the default-set + ChatGPT tests and added no-param and `?tools=agent` cases across claude / chatgpt / unknown clients.
- Full suite green: **274 passing**, typecheck clean, `registry:check` + `schemas:check` OK.

## Known follow-up (not in this PR)

`tako_search` / `tako_answer` tool descriptions still tell the model to "use the Tako Answer Agent" for figure-it-out questions. When a connection hasn't enabled the agent that guidance dangles (the model just falls back to search). Worth softening to "(if enabled)" in a follow-up, since it's model-facing prompt copy with its own test coverage.


workers/src/index.test.ts
```diff
@@ -332,6 +326,70 @@ describe("worker routing", () => {
...
+    };
+    const names = new Set(body.result.tools.map((t) => t.name));
+    // Claude / unknown clients get the single-call tool...
+    expect(names.has("tako_agent")).toBe(true);
+    // ...and never the ChatGPT-only split pair.
+    expect(names.has("tako_agent_start")).toBe(false);
+    expect(names.has("tako_agent_wait")).toBe(false);
+    expect(names.has("tako_search")).toBe(true);
+    // 8 default tools + tako_agent = 9.
+    expect(body.result.tools).toHaveLength(9);
```
noahjax, line 390: [Suggestion] The unit tests prove `parseEnabledOptionalToolNames("nope") → ∅`, but there's no e2e case asserting an *unknown* `?tools=` value still yields the default 8 tools at the request layer. A one-line `?tools=nope → toHaveLength(8)` case here would guard the "a typo never breaks the connection" promise end-to-end, alongside these no-param / `?tools=agent` cases.

noahjax (approved):
Clean, well-specced change — the opt-in gate runs ahead of the per-client filters (verified), routing is by `pathname` so `?tools=agent` never perturbs the OAuth `.well-known` paths, and coverage is solid. Approving. Two non-blocking items:

1. **`llms-full.txt` missed the `sources` accuracy fix.** This PR corrects the `sources` default in the README (`["data","web"]`, with `"tako"` a legacy synonym), but `llms-full.txt:30` still documents `tako_search`'s `sources` as `["tako"] (curated, default)` — which is factually wrong (the default is both data+web, not curated-only) and is the exact class of inaccuracy this PR's accuracy pass set out to fix. `llms-full.txt:41` (`tako_answer`) also still uses the legacy `tako` value. Since these are model-facing docs, worth folding into the same PR.

2. **Connector-continuity — the crux of the rollout.** The unit/integration tests can't cover whether Claude.ai / ChatGPT actually *preserve* `?tools=agent` on every `tools/list` POST after the OAuth connect — that host behavior is the sole path by which an existing install re-enables the agent. Recommend a staging smoke with a real Claude.ai + ChatGPT connector on `?tools=agent` before cutting the release.

docs/superpowers/specs/2026-07-18-opt-in-agent-via-tools-param-design.md
jed326, line 1: Sorry I missed this in previous PRs, but the MCP repo is a public repo so we shouldn't commit the superpowers specs in there.

PR #139, opened 2026-07-20
feat(tools): slim inline data + clarify search vs answer routing

25eliu:
## Problem

Two related issues surfaced while dogfooding the MCP:

1. **Token bloat.** A single `tako_answer`/`tako_search` call returned hundreds of time-series rows inlined per cited card (`content.dataset.rows`), dwarfing the actual answer and burning context/tokens. Both channels the model sees — `content.text` and `structuredContent` — derive from the tool output, and `structuredContent` counts toward context on claude.ai, so the rows were tokenized in full.
2. **Routing confusion.** A model synthesizing data to build a thesis would call `tako_search`, get useful-looking cards, and **stop** — never calling `tako_contents` to pull the underlying data. The copy described `tako_contents` as a *capability*, not a *step in a workflow*, and `tako_search` implied the card **is** the data.

### Behavior — slim the model-facing payload
- **New shared helpers** (`_search_results.ts`): `slimCardContent(content, capRows)` drops (`null`) or caps (last N) the `data`/`records`/`dataset` row payloads, always preserving metadata (`content_format`/`cost`/`total_rows`/`truncated`) so the "more data available, priced" signal survives. Plus `slimCard` / `slimWebResult` and `INLINE_PREVIEW_ROW_CAP = 5`.
- **`tako_answer`** — removed `include_contents` entirely. The synthesized prose answer is the payload; cited cards never inline rows and web text is never inlined. Model fetches specifics on demand via `tako_contents`.
- **`tako_search`** — `include_contents` now defaults **true** (data-first): each card carries a free 5-row preview; set `false` for a pointers-only response (cheap parallel fan-out). The **web source is pinned `include_contents=false`** — web page text is billed per page, so it is never auto-inlined (fetch via `tako_contents`).

### Routing — clarify the mental model
- **`tako_answer`** = ask one *specific, self-contained* data question, get the answer written for you (you don't touch the rows).
- **`tako_search`** = pull the *actual data* to work with, and **parallelize it** — decompose a broad/multi-part request into several narrow single entity+metric searches run concurrently (e.g. four searches for "CPI / core CPI / PCE / core PCE"), not one broad query.
- Kept the earlier `tako_contents` handoff clarification (search → contents when you need the rows).

### Docs + generated
- Updated `README.md`, `llms.txt`, `llms-full.txt`, and both tool descriptions.
- Re-ran `npm run registry:gen` → `registry/server.json` + `_registry.ts` in sync.

## Test plan
- [x] `npm test` — **280/280 pass** (added row-cap / row-drop / web-text-drop coverage; `gen-registry.test.ts` confirms `server.json` in sync)
- [x] `npm run registry:check` / `schemas:check` — ok
- [ ] Spot-check on a real host: `tako_answer` returns compact prose+pointers; `tako_search` shows the 5-row preview + inline chart; model escalates to `tako_contents` for full data


workers/src/tools/_search_results.ts
```diff
@@ -93,6 +93,79 @@ export const webResultSchema = z
...
+    !!dataset && Array.isArray(dataset.rows) && dataset.rows.length > capRows;
+  return {
+    ...meta,
+    // CSV text is always dropped on a cap — the structured rows are the
+    // cheaper, cleaner peek; a caller wanting CSV calls tako_contents.
+    data: null,
+    records: Array.isArray(records) ? records.slice(-capRows) : (records ?? null),
+    dataset:
+      dataset && Array.isArray(dataset.rows)
+        ? { ...dataset, rows: dataset.rows.slice(-capRows) }
```
robertabbott: The cap keeps the tail (`dataset.rows.slice(-capRows)`, and `records.slice(-capRows)` on line 143), which only yields the "most-recent" rows if the backend returns them oldest-first. Backend ordering is card-type-dependent: timeseries is ascending (tail = newest ✓), but **stock cards are sorted descending / newest-first** (`configs/data/modelml/modelml_stocks.py:337`; the contract spells it out at `api/ga/content_types.py:184-186` — "ascending for timeseries, descending for stocks"). So for a stock card this preview hands the model the *oldest* 5 rows of the recent window and drops the latest price — on a core query type, and contradicting the "latest rows" wording in the tool description. Should the cap be applied backend-side (where the ordering is known), or the slice direction detected rather than assumed tail? No test covers descending input.

robertabbott (commented):
Super-review (risk: HIGH public MCP contract / MEDIUM payload reshaping + hot path). CI green, 280/280. Two suspicions cleared during verification against the backend: the free-preview billing (confirmed free at any row count) and the CSV `data:null` drop (safe today — the worker never requests `content_format:"csv"`). Three findings below; #1 is a correctness bug I'd fix before merge, #2/#3 are non-blocking.

workers/src/tools/tako_search.ts
```diff
@@ -63,9 +66,9 @@ const inputSchema = z.object({
     .describe("Maximum number of results to return per source (1-20)."),
   include_contents: z
     .boolean()
-    .default(false)
+    .default(true)
```
robertabbott, line 69: Defaulting `include_contents` to `true` puts inline-preview materialization on every default search. The preview is free (confirmed), but `_inline_tako_card` runs `get_dataset_result()` **serially per card** and can trigger a cold DB query for fact-table-backed cards (`api/ga/v3/search/auto_contents.py:99-107`). Previously (default `false`) searches skipped this; now every default search pays it across up to `count` cards. The backend comment argues it's negligible because the pipeline just warmed those queries — was the added per-search latency measured on this default flip?

workers/src/tools/_search_results.ts
```diff
@@ -93,6 +93,79 @@ export const webResultSchema = z
...
+  if (capRows === null) {
+    return { ...meta, data: null, records: null, dataset: null } as ResultContent;
+  }
+  const slicedRecords = Array.isArray(records) && records.length > capRows;
+  const slicedRows =
+    !!dataset && Array.isArray(dataset.rows) && dataset.rows.length > capRows;
+  return {
+    ...meta,
+    // CSV text is always dropped on a cap — the structured rows are the
+    // cheaper, cleaner peek; a caller wanting CSV calls tako_contents.
```
robertabbott: The unconditional `data: null` on a positive cap silently empties a CSV-format preview. Safe today because the worker never sends `content_format` and the backend defaults to `json_compact` — but `content_format:"csv"` is a first-class backend option whose payload lives entirely in `content.data`, so if a format param is ever threaded into `tako_search`, an `include_contents:true` caller would get zero rows and no error. Worth a guard, or at least a test pinning the CSV+cap path (the `records`-array slice branch is also untested).

robertabbott (approved):

PR #142, opened 2026-07-20
fix(mcp): surface upstream 4xx error bodies (403/401/404/…) to the model

25eliu:
## Problem

Client-error (4xx) responses from the Tako backend carry actionable detail the model should relay to the user, but the MCP boundary was dropping most of it. The model saw only opaque messages like `Django returned 403 for POST /api/v1/contents/`.

This started from a specific report — a `403` on a **protected source** surfaced without Tako's real message (`Data export is not available for this card (protected source).`) — but a review of the whole error path found the same class of gap across **401, 403, and 404**.

## What was being dropped (before)

| Status | `_meta.body` | model-visible text | Notes |
|--------|:---:|:---:|-------|
| 400 | ✅ | ✅ (raw JSON) | only status spliced into text |
| 401 | ❌ | ❌ | body was **never even read** in `django.ts` |
| 403 | ✅ | ❌ | reason stranded in `_meta` (most clients don't surface it) |
| 404 | ❌ | ❌ | `django.ts` read the body, `djangoErrorToToolResult` **dropped** it |
| 5xx | ✅ | ❌ | correct — server errors stay body-free |

## Changes

1. **`django.ts`** — `DjangoUnauthorizedError` now captures the 401 response body (like the 400/404/catch-all subtypes already do).
2. **`mcp.ts` `djangoErrorToToolResult`**:
   - Surface the captured body on `_meta["tako/error"].body` for **every** body-bearing subtype (400/401/404/catch-all), via a single `errorResponseBody()` helper.
   - Splice the body into the model-visible text for **all 4xx** client errors (400/401/403/404/409/429/…), replacing the previous 400-only + 4xx-`HttpError` special-case.
   - `extractErrorDetail()` lifts the message out of DRF JSON envelopes (`{"detail":"…"}`, `{"error":"…"}`, `{"detail":["…"]}`) so the model sees clean text, not raw JSON. Falls back to the raw body for unrecognised shapes (field-keyed 400 validation maps, non-JSON text, truncated bodies).
   - **5xx server errors and transport errors (timeout) stay body-free** — no actionable detail, and 5xx bodies are often noisy HTML error pages.

The full raw envelope always stays on `_meta`. `err.message` remains body-free by construction (log-injection guard in `django.ts`) — the splice happens only at the MCP boundary. Graph tools already map Django errors to actionable messages via `graphErrorMessage()` and are unaffected.

The model now sees, e.g.:

> `Django returned 403 for POST /api/v1/contents/: Data export is not available for this card (protected source).`

## Test plan

- [x] `npm test` — full suite green (**273 passing**), incl. 17 mcp error-mapping tests
- [x] `tsc --noEmit` clean
- [x] Regression tests: protected-source 403, 401 auth-failure body, 404 not-found body — each splices the lifted `detail` into text and keeps the raw envelope on `_meta`
- [x] No-body 401/404 stay body-free with no noisy `body: ""` in `_meta`
- [x] Non-JSON 4xx body (429) falls back to raw body
- [x] `extractErrorDetail` unit tests: `detail` / `error` / nested list / field-keyed 400 map / non-JSON / truncated body
- [x] `django.ts`: 401 now captures the body
- [x] Existing 400-validation and 5xx body-free behaviour preserved

### DIFF
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -734,25 +734,32 @@ function registerTool(
  */
 /**
  * Lift a human-readable message out of an upstream error body for the
- * model-visible text channel. Django/DRF errors arrive as JSON envelopes —
- * `{"detail": "…"}` (APIException), `{"error": "…"}` (custom handlers), or
- * `{"detail": ["…"]}` — and the raw JSON reads poorly to the model. When the
- * shape is recognisable we return just the message; otherwise (unrecognised
- * keys like a field-keyed 400 validation map, non-JSON text, or a body
- * `safeReadText` truncated so `JSON.parse` fails) we fall back to the raw body.
+ * model-visible text channel, or `undefined` when the body is NOT a
+ * recognisable structured envelope.
  *
- * Best-effort only: the full raw body always stays on
+ * Django/DRF errors arrive as JSON envelopes — `{"detail": "…"}`
+ * (APIException), `{"error": "…"}` / `{"message": "…"}` (custom handlers),
+ * `{"detail": ["…"]}` (nested list), or a field-keyed validation map
+ * `{"field": ["…"], …}`. We lift the message out of those.
+ *
+ * Anything else returns `undefined` so the caller keeps it OUT of the text
+ * channel: non-JSON bodies (e.g. a Cloudflare/edge HTML challenge or block
+ * page on a 4xx, or a body `safeReadText` truncated so `JSON.parse` fails)
+ * and JSON objects with no recognisable message (e.g. a bare discriminator
+ * like `{"error_type": "RELEVANT_RESULTS_NOT_FOUND"}`) would otherwise flood
+ * the model with raw HTML/JSON. The full raw body always stays on
  * `_meta["tako/error"].body` for clients that want the untouched envelope.
  */
-export function extractErrorDetail(body: string): string {
+export function extractErrorDetail(body: string): string | undefined {
   let parsed: unknown;
   try {
     parsed = JSON.parse(body);
   } catch {
-    return body;
+    // Not JSON — an HTML edge page or truncated body. Don't surface it.
+    return undefined;
   }
-  if (typeof parsed === "string" && parsed !== "") return parsed;
-  if (parsed === null || typeof parsed !== "object") return body;
+  if (typeof parsed === "string") return parsed !== "" ? parsed : undefined;
+  if (parsed === null || typeof parsed !== "object") return undefined;
   const obj = parsed as Record<string, unknown>;
   for (const key of ["detail", "error", "message"]) {
     const value = obj[key];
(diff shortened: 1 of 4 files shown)


workers/src/mcp.ts
```diff
@@ -744,22 +777,34 @@ export function djangoErrorToToolResult(err: DjangoError): {
...
+  // Keeping it in `_meta` alone isn't enough — not every MCP client surfaces
+  // structured detail to the model.
+  //
+  // SERVER errors (5xx) and transport errors (timeout — no status) stay
+  // body-free: no LLM-actionable detail, and 5xx bodies are often noisy HTML
+  // error pages that would flood the text channel. `extractErrorDetail` lifts
+  // the message out of DRF JSON envelopes; `err.message` stays body-free by
+  // construction (log-injection guard in `django.ts`), so the splice happens
+  // only here at the MCP boundary.
+  const is4xx =
```
robertabbott, line 833: The splice is gated on status class (`is4xx`) alone. The 5xx exclusion is justified as avoiding "noisy HTML error pages" — but since we now call `tako.com` through Cloudflare, an edge-level 403 (WAF block) or 429 (rate-limit) returns an HTML challenge/block page too, and those are 4xx, so they'd get spliced raw into the model-visible text (`extractErrorDetail` fails `JSON.parse` on HTML → returns the raw body, up to the 8 KiB cap). That's the same "noisy HTML floods the text channel" failure mode the 5xx exclusion guards against, just on 4xx. Is the cap considered sufficient here, or is it worth gating the splice on a recognised/JSON-ish body (the `extractErrorDetail`-found-an-envelope case) so only DRF-shaped detail reaches the model?

workers/src/mcp.ts
```diff
@@ -732,6 +732,39 @@ function registerTool(
...
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(body);
+  } catch {
+    return body;
+  }
+  if (typeof parsed === "string" && parsed !== "") return parsed;
+  if (parsed === null || typeof parsed !== "object") return body;
+  const obj = parsed as Record<string, unknown>;
+  for (const key of ["detail", "error", "message"]) {
```
robertabbott, line 764: The key loop recognises `message` too, but the docstring (and the PR description) only list `detail` / `error` / nested-`detail`-list. Worth adding `message` to the documented shapes so the recognised-key set matches the code?

workers/src/mcp.ts
```diff
@@ -744,22 +777,34 @@ export function djangoErrorToToolResult(err: DjangoError): {
...
+  // the message out of DRF JSON envelopes; `err.message` stays body-free by
+  // construction (log-injection guard in `django.ts`), so the splice happens
+  // only here at the MCP boundary.
+  const is4xx =
+    err.status !== undefined && err.status >= 400 && err.status < 500;
   const text =
-    err instanceof DjangoBadRequestError
-      ? `${err.message}: ${err.body}`
+    is4xx && body !== undefined
+      ? `${err.message}: ${extractErrorDetail(body)}`
```
robertabbott: 404 bodies are now spliced for every 404, including the `knowledge_search`/no-results overload the `django.ts` comment describes (`{"error_type": "RELEVANT_RESULTS_NOT_FOUND", …}`). No tool inspects that body, and `extractErrorDetail` finds no `detail`/`error`/`message` key in it, so the model would see the raw discriminator JSON. Does any endpoint the tools actually call return that shape on 404 — and if so, should the no-results case map to a cleaner "no results" message rather than the raw envelope? (Couldn't confirm from the MCP repo alone whether `/api/v3/search` etc. emit it.)

robertabbott (approved):

PR #144, opened 2026-07-21
fix(tools): only call tako_contents on cards that carry a content attribute

25eliu:
## Summary

A Tako knowledge card returned by `tako_search`/`tako_answer` is exportable via `tako_contents` **only when it carries a non-null `content` attribute** (the export descriptor). The backend's export-safe gate rejects cards without one (403; the OpenAPI contract also documents 404 as "does not exist or has no exportable data"). Models were calling `tako_contents` on unexportable cards because no model-facing surface stated the rule — every surface said unconditionally "for the full data, call `tako_contents` on the result URL".

The plumbing already preserved the signal (`slimCardContent` strips row payloads but keeps the descriptor object, and never fabricates one), so this PR makes the rule explicit everywhere and self-correcting at runtime:

- **Descriptions** — `tako_search` gains an explicit "Exportability gate" sentence; `tako_answer` conditions its `tako_contents` guidance on `content` presence; `tako_contents` states the precondition up front. `takoCardSchema.content` gets a `.describe()` so the rule rides in the advertised output schema for both search and answer.
- **Runtime backstop** — the `tako_contents` handler maps the endpoint's 403 (export-safe gate) and 404 (no exportable data / missing) to self-correcting error messages that include the backend's own detail (truncated) and steer the model back to the gate, instead of the opaque "Django returned 403". Framed as the likely cause, not asserted fact, mirroring how `_graph.ts` treats ambiguous 403s.
- **Docs** — same rule in `llms.txt`, `llms-full.txt`, `agent.json`, and README (all three `tako_contents` mentions). Web URLs are consistently described as "exempt from this gate" rather than the overclaim "always fetchable".
- **Tests** — `slimCard` never fabricates nor drops the `content` descriptor (4 new tests); 403/404 mapping + non-403 passthrough (3 new tests; the django mock now spreads `importOriginal` so the real error classes survive for `instanceof`).
- **Registry** — `registry/server.json` regenerated from the updated descriptions (`registry:check` deploy gate passes).

Stacked on #143 (`docs/graph-fallback-guidance`) because the description strings build on that branch's wording; it will retarget to `main` when #143 merges.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run` — 303/303 passing (7 new tests)
- [x] `npm run registry:check` passes (was failing before regen)
- [x] Swept all doc surfaces (`llms.txt`, `llms-full.txt`, `agent.json`, README, tool descriptions) for remaining unconditional `tako_contents` guidance — none left
- [ ] Live spot-check against a running worker: card **with** `content` → descriptor survives in output (both `include_contents` modes); card **without** `content` → no descriptor fabricated; `tako_contents` on an unexportable card → new self-correcting message


workers/src/tools/_search_results.ts
```diff
@@ -52,8 +56,15 @@ export const takoCardSchema = z
...
-    // Inline card CSV — present only when include_contents was set for the tako source.
-    content: resultContentSchema.nullable().optional(),
+    // Export descriptor + inline preview. Present iff this card's data is
+    // exportable via tako_contents (it rides along even with include_contents
+    // off, carrying the cost quote). Missing/null → NOT exportable (403).
+    content: resultContentSchema
+      .nullable()
+      .optional()
+      .describe(
+        "Export descriptor + inline data preview. Present (non-null) ONLY when this card's underlying data is exportable via tako_contents. Missing or null means the card has NO exportable data — do NOT call tako_contents on its URL (the export gate rejects it).",
```
robertabbott: The `.describe()` states `content` is "Present (non-null) ONLY when this card's underlying data is exportable via tako_contents." In the backend the adapter sets `content` via the lenient `supports_data_export()` (excludes only `PROTECTED_SOURCES`), while `/api/v1/contents` gates on the stricter `export_safe()` — which additionally requires a non-empty source set where every source resolves to a known source or is web-marked (its docstring calls itself "the fail-closed inverse of supports_data_export"). So a card with an unresolvable/legacy/blank source name comes back with `content` present yet still 403s on export. The "absent ⇒ not exportable" direction holds; "present ⇒ exportable" does not. Should this description (and the parallel doc surfaces) frame `content` presence as necessary-but-not-sufficient — and should the 403 message at `tako_contents.ts:135` stop asserting the card "carried no `content` attribute," given a content-present card is exactly what reaches that branch surprisingly?

workers/src/tools/tako_contents.ts
```diff
@@ -109,13 +111,38 @@ const takoContents = {
...
+      // Map the two "this URL has no exportable content" statuses to
+      // self-correcting messages so the model stops retrying and falls back
+      // to the card's preview/metadata. Per the OpenAPI contract: 403 is the
+      // export-safe gate (e.g. protected-source export); 404 is "does not
+      // exist or has no exportable data". Both are framed as the LIKELY
+      // cause — not asserted fact — and carry the backend's own detail,
+      // since 403 in particular can have other causes (cf. _graph.ts, where
+      // a 403 on /beta/graph is an edge block, not a query problem).
+      if (err instanceof DjangoHttpError && err.status === 403) {
+        const detail = err.body.slice(0, 200);
```
robertabbott: `err.body.slice(0, 200)` splices the raw upstream body into the model-visible error text (same at line 139 for 404). The comment just above notes a 403 here "can have other causes (cf. `_graph.ts`, where a 403 on `/beta/graph` is an edge block)". For an edge/WAF 403 the body is a raw HTML block page, so `403: <!DOCTYPE html>…` lands in the model text — the exact "noisy body floods the text channel" failure that `extractErrorDetail()` (added and exported in this same PR) was built to prevent, and that `_graph.ts` avoids by not splicing 403 bodies at all. Would routing `err.body` through `extractErrorDetail()` — splice only when it recognizes structured detail, otherwise omit — be better than a raw slice here?

workers/src/mcp.ts
```diff
@@ -770,6 +770,63 @@ function registerTool(
...
+ *
+ * Anything else returns `undefined` so the caller keeps it OUT of the text
+ * channel: non-JSON bodies (e.g. a Cloudflare/edge HTML challenge or block
+ * page on a 4xx, or a body `safeReadText` truncated so `JSON.parse` fails)
+ * and JSON objects with no recognisable message (e.g. a bare discriminator
+ * like `{"error_type": "RELEVANT_RESULTS_NOT_FOUND"}`) would otherwise flood
+ * the model with raw HTML/JSON. The full raw body always stays on
+ * `_meta["tako/error"].body` for clients that want the untouched envelope.
+ */
+export function extractErrorDetail(body: string): string | undefined {
```
robertabbott: Inserting `extractErrorDetail` (with its own JSDoc at 773–790) directly below the block ending at line 772 ("Exported for unit testing — … the `kind` strings here.") leaves that block stranded above `extractErrorDetail`, and `djangoErrorToToolResult` (line 830) now has no doc comment. That "Exported for unit testing / `kind` strings" note describes `djangoErrorToToolResult`'s `_meta`/`kind` contract, not `extractErrorDetail`. Should it move back down onto `djangoErrorToToolResult`? (This originates in the #142 code carried in this branch, so fine to drop as out of scope.)

PR #146, opened 2026-07-21
fix(tools): only call tako_contents on cards that carry a content attribute

25eliu:
## Summary

Re-lands #144, which was accidentally merged into its stale stacked base branch (`docs/graph-fallback-guidance`) after that branch's own PR (#143) had already merged to main — so its changes never reached main. This PR merges that branch (now containing exactly main + the #144 merge) into main; the diff is precisely #144's changes, including the review-round fixes.

From #144:

- **Exportability gate** — a Tako card is exportable via `tako_contents` only when its search/answer result carried a non-null `content` attribute (the export descriptor). Stated in the tool descriptions, output schema `.describe()`, and all doc surfaces (`llms.txt`, `llms-full.txt`, `agent.json`, README).
- **Runtime backstop** — the `tako_contents` handler maps the endpoint's 403 (export-safe gate) and 404 (no exportable data) to self-correcting messages that steer the model back to the gate.
- **Review round** (robertabbott): `content` presence framed as necessary-but-not-sufficient (lenient `supports_data_export()` sets it; the stricter `export_safe()` gates the endpoint); the 403 message no longer asserts the card lacked a `content` attribute; 403/404 bodies route through `extractErrorDetail()` (moved to `django.ts` to avoid an import cycle) so edge/WAF HTML never floods the model text.
- **Tests** — `slimCard` never fabricates nor drops the descriptor; 403/404 mapping incl. unstructured-body omission; registry regenerated.

## Test plan

- [x] `npx vitest run` — 321/321 passing
- [x] `npm run registry:check` passes
- [x] Verified `origin/docs/graph-fallback-guidance` tree is identical to `origin/docs/gate-contents-on-content-attr` tip (the merged #144 head incl. review fixes)


workers/src/tools/tako_contents.ts
```diff
@@ -109,13 +120,46 @@ const takoContents = {
...
+      // would flood the model text with an edge/WAF HTML block page.
+      //
+      // Note the 403 message does NOT claim the card lacked a `content`
+      // attribute: the search/answer adapter sets `content` via the lenient
+      // supports_data_export(), while this endpoint gates on the stricter
+      // export_safe() — so the card that lands here may well have carried
+      // one (presence is necessary for export, not sufficient).
+      if (err instanceof DjangoHttpError && err.status === 403) {
+        const detail = extractErrorDetail(err.body);
+        throw new Error(
```
robertabbott: On `main`, a 403/404 from `/api/v1/contents/` propagated as a `DjangoError` up to `registerTool` (mcp.ts:627) and became a structured result via `djangoErrorToToolResult` — `isError: true` plus `_meta["tako/error"]` (kind/status/body), with the 4xx detail already spliced into the model-visible text (mcp.ts:815-821). This handler now catches those two statuses and `throw`s a plain `Error`, which falls to the `throw err` fallthrough (mcp.ts:630) and is wrapped generically — so contents 403/404 lose the `_meta["tako/error"]` envelope that every other tool and status still emits. Relatedly, `extractErrorDetail`'s own JSDoc (django.ts:403) still states "The full raw body always stays on `_meta["tako/error"].body`", which is no longer true for this new caller. Is dropping the structured error channel for contents 403/404 intended (the richer self-correcting text being the goal), or should the handler preserve `_meta` too?

workers/src/tools/tako_visualize.ts
```diff
@@ -51,11 +51,16 @@ const COMPONENT_TYPES = [
...
-  "For per-`component_type` `config` shapes and more worked examples, see Tako's 'Agent Skills → Visualize Your Data' docs and the Thin-Viz chart-creation reference. NOTE: `person_card` must be the ONLY component when used. **Always include `[Open in Tako](embed_url)` once at the end of your reply.**";
+const DESCRIPTION = [
+  "Create an embeddable Tako chart/card from data you ALREADY HAVE — use `tako_search` to find existing Tako data instead. Auto-renders inline; returns `webpage_url` / `embed_url`.",
+  "",
+  "Input: one or more typed `components`, each `{component_type, config}` — e.g. `header`, `generic_timeseries`, `categorical_bar`, `table`, `financial_boxes`.",
+  "",
+  "Example — a titled bar chart is two components:",
+  '{"components": [{"component_type": "header", "config": {"title": "Revenue"}}, {"component_type": "categorical_bar", "config": {"datasets": [{"label": "Sales", "data": [{"x": "NA", "y": 500}, {"x": "EU", "y": 300}]}]}}]}',
```
robertabbott: This trimmed example omits `units` from the `categorical_bar` dataset, but the `config` param `.describe()` at line 81 still shows `{"label": "Sales", "units": "USD", "data": [...]}` for the same `component_type`. A model reading the two together sees conflicting shapes for one component. Is `units` required server-side (in which case the trimmed example would generate an invalid config), or optional (in which case, worth making the two examples agree)?

robertabbott (approved):

PR #148, opened 2026-07-21
feat(tools): add tako_available_data; demote graph primitives

25eliu:
## Summary

Adds **`tako_available_data`** — a free, fast, single-call tool that answers "what data does Tako have on X?" and makes it the default graph-discovery surface, replacing the three lower-level graph primitives (which are demoted, not deleted).

Given a name, it resolves the top graph matches and, for each, reports the live data Tako holds as a natural-language summary:
- an **entity** node (e.g. *Tesla*) → the **metrics** Tako tracks for it
- a **metric** node (e.g. *Inflation Rate*) → the **entities** it is tracked across

The summary is built deterministically in the Worker (no LLM call), so it's low-latency and free. Canonical names inline double as the exact terms to reuse in a follow-up `tako_search`.

### Pipeline
1. `graph/search` to resolve the name (~0.25s)
2. For the top 2 hits, a **type-aware** parallel `graph/related` drill — `relation=metrics` for entity nodes, `relation=entities` for metric nodes
3. Deterministic prose in `_available_data.ts`

Total latency is typically sub-second (mega-nodes like *United States* dropped from ~4s on the old overview path to ~0.6s on the targeted drill).

### Behavior details
- **Type-aware coverage** — drilling `relation=metrics` on a metric node returns empty, so metric nodes report their entity coverage instead (avoids a false "no metrics" answer).
- **Headline-first metrics** — internal/accounting names (`… (Normalized)`, `Account Code …`) are deprioritized in the preview (never dropped from the total).
- **Capped totals** render as `N+` (honors the API's `total_capped`).
- **Per-node error isolation** — one node's failed drill yields an "unavailable" match rather than sinking the whole call.
- **Empty search** fast-exits with a "no node, try tako_search" summary and zero drill calls.

### Demotion (kept in-tree for rollback)
`tako_graph_search`, `tako_graph_related`, `tako_graph_node` renamed with a `_` prefix (registry codegen skips them), removed from `MCP_TOOL_ALLOWLIST` and the `?tools=` opt-in aliases. Still compiled and tested. Default tool surface is now `tako_search`, `tako_answer`, `tako_contents`, `tako_available_data`.

### Docs & registry
Updated all sibling tool descriptions, `README.md`, `llms.txt`, `llms-full.txt`, `AGENTS.md`, `agent.json`, regenerated `registry/server.json` + `_registry.ts`, and added a `tako_available_data` smoke canary.

## Test plan
- [x] `npm test` — **355 pass** (27 for the new tool: pure formatter + handler, incl. metric-node regression + capped rendering)
- [x] `npm run registry:check` — ok (9 tools)
- [x] `npm run schemas:check` — ok
- [x] **Live-verified against staging**: Apple, Nvidia, Tesla, US GDP (entity→metrics), Inflation Rate / Unemployment Rate (metric→entities), Bitcoin, empty/nonsense queries — all render correctly, sub-second.
- [ ] `npm run smoke` against staging **after deploy** (new canary calls `tako_available_data`)

## Review
Passed an adversarial code review (2 rounds): a HIGH (metric-node false negative) and a MEDIUM (`total_capped`) were found, fixed, tested, and re-confirmed clean.


workers/src/tools/_available_data.ts
```diff
@@ -0,0 +1,244 @@
...
+}): string {
+  const { query, matches, otherMatches } = input;
+
+  if (matches.length === 0) {
+    return `Tako has no data-graph node matching "${query}". Tako may still have relevant public/web data — try tako_search directly, or rephrase the entity or metric name.`;
+  }
+
+  const n = matches.length;
+  const header = `Tako has live data on ${n} ${plural(n, "match", "matches")} for "${query}":`;
+  const lines = matches.map(matchLine);
```
jed326, line 251: **Finding (Important):** the header `Tako has live data on ${n} matches` is emitted whenever `matches.length > 0`, regardless of coverage — and `found` is set `true` in `tako_available_data.ts` on node *resolution* alone. This contradicts the per-match lines below (`resolved, but Tako holds no metrics for it yet` / `couldn't load its coverage right now`) **and the tool's own published contract** (`llms-full.txt`: *"a match with no coverage is a real answer — report the gap rather than silently falling back to web"*).

**Evidence:** an entity that resolves with 0 metrics → header asserts *live data* while the body says there is none. If both top drills fail (e.g. 503) → `found:true` + "has live data on 2 matches" over two "couldn't load coverage" lines. An agent keying off the prose or `found` may wrongly narrow `sources` to `["data"]`.

**Proposed fix:** gate the header on real coverage.
```ts
const withData = matches.filter((m) => !m.unavailable && m.coverage.total > 0).length;
const header = withData > 0
  ? `Tako has live data on ${withData} of ${n} ${plural(n, "match", "matches")} for "${query}":`
  : `Resolved ${n} ${plural(n, "match", "matches")} for "${query}", but no live coverage loaded:`;
```
Consider also deriving `found` from `withData > 0` rather than node-resolution, since consumers read it as "has data."

workers/src/tools/_available_data.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (m.coverage.total === 0) {
+    return `${head} — ${emptyClause(m.coverage.kind)}.`;
+  }
+  return `${head} — ${coverageClause(m.coverage)}.`;
+}
+
+// Pick a real next-step example. For an entity's metrics: "Tesla, Inc. Revenue".
+// For a metric's entities: "United States Inflation Rate" (entity + metric).
+function nextStepExample(matches: CoverageMatch[]): string | null {
+  const m = matches.find((x) => !x.unavailable && x.coverage.names.length > 0);
```
jed326, line 214: **Finding (Minor, compounds the header issue):** when no match has coverage names, `nextStepExample` falls back to `any.name`, so `buildSummary` appends a `call tako_search using "X"` hint for an entity the tool just reported has **no data** — steering the agent into a priced dead-end.

**Proposed fix:** only emit the example when the chosen match actually has coverage; drop the fallback.
```ts
const m = matches.find((x) => !x.unavailable && x.coverage.names.length > 0);
if (!m) return null;
```

workers/src/tools/_available_data.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (!group) return emptyGroup(kind);
+  const raw = group.items.map((i) => i.name);
+  const ordered = kind === "metrics" ? orderMetricNames(raw) : raw;
+  const total = group.total ?? ordered.length;
+  return {
+    kind,
+    names: ordered.slice(0, PREVIEW),
+    total,
+    truncated: total > Math.min(ordered.length, PREVIEW),
+    capped: group.total_capped ?? false,
```
jed326: **Finding (Minor, invariant):** `truncated: total > Math.min(ordered.length, PREVIEW)` reports `false` when `capped === true` but `total === shown count`, which violates the field's documented contract ("more names exist than are shown").

**Evidence:** not currently triggerable — the drill uses `limit=50`, `PREVIEW=8`, and a capped total is >= 50 > 8, so `truncated` is always `true` when capped. `truncated` is also structured-output-only (never used in the prose). Flagging for the invariant only.

**Proposed fix:** `truncated: capped || total > names.length`.

workers/src/index.test.ts
```diff
@@ -364,8 +362,8 @@ describe("worker routing", () => {
     // The other opt-in tools stay absent for ChatGPT too.
     expect(names.has("tako_graph_node")).toBe(false);
     expect(names.has("get_credit_balance")).toBe(false);
```
jed326, line 365: **Finding (Consistency, cleanup):** leftover assertions `expect(names.has("tako_graph_node")).toBe(false)` at **lines 363 and 467** assert the absence of a tool name that no longer exists as a registered tool. Harmless but dead. **Proposed fix:** prune both, or replace with a positive `expect(names.has("tako_available_data")).toBe(true)`.

workers/src/tools/_graph.ts
```diff
@@ -96,8 +96,11 @@ export function graphErrorMessage(
   err: unknown,
   op: GraphOp,
   ref?: string,
+  toolName?: string,
 ): string {
-  const tool = `tako_graph_${op}`;
+  // `toolName` lets a composite caller (tako_available_data) label the error
+  // with its own name; the demoted primitives fall back to `tako_graph_<op>`.
+  const tool = toolName ?? `tako_graph_${op}`;
   const idOf = ref ? `"${ref}"` : "the given id";
```
jed326, line 105: **Finding (Minor, note-only — no action required):** with the primitives unregistered, `graphErrorMessage`'s `related` and `node` branches (`op !== "search"`) are now unreachable at runtime — `tako_available_data` swallows its `graph/related` failures into `unavailableMatch` and never routes them here; only the `search` branch is live. Not misleading. Keeping the dead branches eases the documented rollback, so leaving them is fine — noting for awareness.

workers/src/tools/tako_available_data.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  },
+});
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  vi.restoreAllMocks();
+});
+
+describe("tako_available_data", () => {
+  it("tool name is tako_available_data", () => {
```
jed326, line 38: **Finding (Test coverage):** no test asserts the summary is non-contradictory when every match is empty or unavailable — the exact case in the header finding above. The `no metrics` and `unavailable` tests assert only the per-match *line*, never that the header avoids claiming "live data."

**Proposed additions:**
- all-empty: one entity hit whose `graph/related` returns empty items → assert `out.summary` does **not** contain "live data" (locks in the header fix).
- a node whose `type` is neither `entity` nor `metric`, driven end-to-end through the handler (only `coverageKindFor` is unit-tested in isolation today).
- assert `label` is omitted from the `graph/search` query when the caller omits it (only the positive "forwards label" case exists).

jed326 (commented):
**Layered pre-merge review.** Six inline findings below. The one I'd fix before merge is #1 (the summary header asserts "Tako has live data" over an empty/unavailable body, contradicting the tool's own `llms-full.txt` contract).

**Design note — flexibility gap (not a blocker, but worth a decision).** Demoting the three graph primitives isn't just "off the default surface" — they're removed from the allowlist *and* the `?tools=` aliases, so with codegen skipping `_`-prefixed files they are **completely unreachable at runtime**. "Demoted, not deleted" holds only for the source. `tako_available_data` re-exposes a narrow slice of `graph/related` (hardcoded `relation=metrics|entities`, top-2 nodes, `limit=50`, no `q`), so these capabilities now have **no MCP path at all**:
- non-coverage relations: `siblings`, `part_of`, `members`, `rel:<phrase>` (graph traversal — "what's in the S&P 500", "who supplies Apple")
- server-side `q` substring filter within a relation (precise "does Apple track *free cash flow*?")
- cursor paging past `limit=50` (names beyond 50 render as `50+` and are unreachable)
- node-detail lookup (`aliases`/`description`/`subtype`, dropped by `buildMatch`)
- coverage for matches beyond `EXPAND_TOP_N=2`

For the stated goal (fast, free, guided coverage discovery) this is a reasonable trade-off, and rollback is a filename rename. **Cheap insurance:** re-add a `graph` opt-in alias in `_optional.ts` mapping to the three primitives — restores the escape hatch for power-user graph traversal at zero cost to the default 4-tool surface, exactly like `agent`/`visualize`/`credits`.

jed326 (approved):
Approving — af406de resolves all six review findings, verified against the diff (not just the message):

- #1 header/`found` now gated on `hasLiveCoverage` (gap / "N of M" / full), so a resolved-but-empty node no longer overclaims "live data".
- #2 next-step hint suppressed when no match has coverage.
- #3 `truncated` ORs `capped`.
- #4 dead `tako_graph_node` assertions replaced with positive presence checks.
- #5 moot now that the branches are reachable again.
- #6 solid new unit + end-to-end regression coverage.

Flexibility gap closed as suggested: `?tools=graph` opt-in alias restores the primitives (traversal relations, `q` filter, cursor paging, node detail) with the default surface unchanged at 4 tools. CI green.

PR #149, opened 2026-07-21
feat(mcp): render charts inline as images on Claude and generic MCP clients

25eliu:
## Summary

Tako cards previously reached every non-ChatGPT host as a bare markdown link. This PR makes charts render **inline as images** on Claude clients (claude.ai, Claude desktop, Claude Code) **and** the unknown-client long tail (Cursor, Windsurf, Gemini CLI, LibreChat, …), by reworking the render gates in `registerTool` (`workers/src/mcp.ts`).

### Render matrix after this PR

| Client | Chart surface |
|---|---|
| ChatGPT | Interactive MCP Apps widget (unchanged) |
| Claude (claude.ai / desktop / Code) | Inline PNG `image` content block (was: link only) |
| Unknown (Cursor, Windsurf, Gemini CLI, …) | Inline PNG `image` content block (was: unrenderable widget metadata + link) |
| All clients | `embed_url` in structured content as the interactive click-through |

### How
Two changes to the gates in `registerTool`:
1. `inlinePngFallbackSuppressed` is decoupled from widget suppression — it now only fires for per-tool ChatGPT suppression (`CHATGPT_NO_WIDGET_TOOL_NAMES`), so the `extraContentBlocks` PNG ships wherever the widget doesn't.
2. `widgetSuppressed` flips from `client === "claude"` to `client !== "chatgpt"` — ChatGPT is the only host that actually renders the widget; everywhere else the metadata just blocked the portable image fallback.

Applies to the chart-bearing tools (`tako_search`, `tako_visualize`). Besides charts rendering in-chat, the model can now **see** the chart while composing its answer on all non-ChatGPT hosts.

## Test plan
- [x] End-to-end gate tests in `workers/src/mcp.test.ts` over an in-memory MCP transport (real server, real `tools/call`, upstream fetch stubbed):
  - claude → exactly one `image` block, no widget `_meta.ui`
  - unknown → exactly one `image` block, no widget `_meta.ui`
  - chatgpt → widget `_meta.ui` present, no image block, no PNG prefetch (single-item fetch queue throws on any second call)
  - claude + zero-card search → no image block, no stray PNG fetch
- [x] `index.test.ts` widget-surface tests updated: widget listing metadata + `resources/list`/`resources/read` asserted under a ChatGPT UA; non-ChatGPT listings asserted widget-metadata-free
- [x] Full suite: 326 tests pass; `npm run typecheck`, `registry:check`, `schemas:check` clean
- [ ] Manual: verify chart renders (not cropped) on claude.ai web with a staging connector
- [ ] Manual: spot-check one generic host (e.g. Cursor) renders the image block
- [ ] Watch response sizes: PNG inlining adds ~70-400 KB base64 per chart-bearing call (capped at 4 MB by `MAX_INLINE_PNG_BYTES`)

## Notes for review
- An old comment claimed the PNG block "renders cropped" on claude.ai — that observation predates the current image-block plumbing; the manual checklist re-verifies it before rollout.
- Rare MCP-Apps-capable non-ChatGPT hosts (Goose, VS Code Insiders) trade the interactive widget for a static image — accepted trade-off for universal rendering.
- Follow-up candidates (not in this PR): extend the same hooks to `tako_answer`/`tako_agent` top cards; re-evaluate the baked-image widget (`ui://tako/embed/chart/{pub_id}`) for claude.ai.


workers/src/mcp.ts
```diff
@@ -372,35 +406,45 @@ function registerTool(
...
+  // blank/broken on ChatGPT in some legitimate success state — the
+  // intended fallback there is the plain text + markdown-link answer,
+  // not a PNG. (The image-block/widget mutual exclusion is separately
+  // guaranteed by the `ui === undefined` condition at the call-time
+  // `extraContentBlocks` gate.)
   const widgetSuppressed =
-    options.client === "claude" || options.widgetSuppressedForTool === true;
-  const inlinePngFallbackSuppressed = widgetSuppressed;
+    options.client !== "chatgpt" || options.widgetSuppressedForTool === true;
+  const inlinePngFallbackSuppressed = options.widgetSuppressedForTool === true;
```
robertabbott, line 447: On `main` this was `inlinePngFallbackSuppressed = widgetSuppressed`, which made the `extraContentBlocks` PNG path effectively dead for every client. This PR activates it for claude + unknown, and the block is `await`ed before the tool result returns (line ~710) — so every `tako_search`/`tako_visualize` call now carries a synchronous PNG fetch with up to an 8s `PNG_FETCH_TIMEOUT_MS` ceiling, where claude previously got a link only with no fetch. Is that added wall-clock on the primary search path acceptable, or worth making the image block best-effort / non-blocking so a slow render can't stall an otherwise-fast search up to 8s?

workers/src/mcp.ts
```diff
@@ -685,11 +727,18 @@ function registerTool(
...
-      // above. `tako_search` uses `extraMeta` to ship `image_data_url`
-      // for the widget to read via `params._meta`. When the widget is
-      // suppressed (claude.ai), no widget will consume `_meta`, so
-      // running this hook would inflate the JSON-RPC response with a
-      // ~330 KB unused data URL.
+      // above. When the widget is suppressed (every non-ChatGPT client),
+      // no widget will consume `_meta`, so running this hook would
+      // inflate the JSON-RPC response with a ~330 KB unused data URL.
+      //
+      // NB: with the widget now ChatGPT-only, this hook only fires for
```
robertabbott, line 734: This comment notes `_meta` was chosen to avoid inflating the response, and the old widget path capped the baked PNG at `MAX_INLINE_DATA_URL_BYTES` = 250 KB specifically to stay under claude.ai's "tool result too large for context" guard. The newly-activated content-block path (`fetchPngContentBlock`) instead caps at `MAX_INLINE_PNG_BYTES` = 4 MB (~5.3 MB base64) and does enter the model context (that's the goal — the model sees the chart). Is 4 MB intended for a context-entering block, or should it match the tighter 250 KB cap the earlier design deliberately used? And does the manual checklist cover size-rejection at the top of the range, not just the cropping case?

robertabbott (approved):

PR #151, opened 2026-07-22
feat(tools): explicit `exportable` flag + reframe `tako_available_data` descriptions

25eliu:
## Problem

A Tako card signaled export-eligibility only by the **presence** of a `content` attribute — i.e. an *absent key* meant "not exportable." LLM clients routinely don't notice a missing key, so they call `tako_contents` anyway and eat a self-correcting **403**.

Real example that prompted this: an S&P "Alphabet Quarterly EBIT Margin" card came back from `tako_search` with **no `content` key**, and the model called `tako_contents` on it regardless → 403.

## Fix

Emit a **positive `exportable` boolean** on every card, on both `tako_search` and `tako_answer`.

Since TakoData/tako#27989 (merged) the base API emits the flag **authoritatively** — computed with the same fail-closed `export_safe` gate `/api/v1/contents` enforces, with `content` gated on it too (that PR also closed the lenient `supports_data_export` vs strict `export_safe` mismatch, where a rare content-bearing search/answer card could still 403 on export). `slimCard` therefore **passes the wire value through untouched** and only falls back to deriving it from `content != null` when an older backend omits the flag. Pure in-memory either way — no extra backend call, no added latency.

Semantics:
- `exportable: false` → authoritative, never call `tako_contents`
- `exportable: true` → eligible but **not** guaranteed (a rare card can still 403 at export time; fall back on error, don't retry)

`tako_contents` is taught about the flag: its precondition and both the 403/404 self-correcting messages now lead with `exportable` (keeping the `content` equivalence for accuracy). Docs (`llms.txt`, `llms-full.txt`, `README.md`) and the generated `registry/server.json` are synced.

**Scoped to cards only** — web URLs are exempt from the export gate, so tagging them would risk nudging models toward unnecessary (billed) page fetches.

---

## Also in this PR: reframe `tako_available_data` + tool-description cleanup

Follow-on refinement to the tool descriptions this branch introduced.

**Problem.** `tako_available_data` was framed as "see what data **Tako** has on X." An MCP client has no prior on what "Tako" is, so that's a weak signal — and it under-sold the tool's real value: answering the general question *"what proprietary/structured data exists on X?"* and serving as a cheap **pre-flight accuracy check** before spending a priced `tako_search` / `tako_answer`.

**Changes.**
- **`tako_available_data`** — concise lead line (drops "entity + metric"); positions the tool as (a) the answer to *"what structured/proprietary data is there on X?"* and (b) a free accuracy check to confirm a figure exists *and its exact name* before a priced call.
- **Framing convention** — lead every model-facing description with the capability (**"proprietary, structured data"**) and keep "Tako" only for tool cross-references. Neutralized "Tako covers / Tako's data source" phrasing in `tako_search`, `tako_answer`, and `tako_agent` (`sources` + `node_ids`).
- **Brief cross-references** — added short available-data pointers to `tako_search` / `tako_answer` where they apply, at negligible token cost.
- **Coverage prose** — `PREVIEW` 8 → 50 (the coverage names *are* the payload the model reuses), and the summary no longer repeats those names — they live once in `matches[].coverage.names`, so the prose carries counts + a single next-step example. Cuts token duplication.
- **Docs** — `README.md`, `AGENTS.md`, `agent.json`, `llms.txt`, `llms-full.txt` synced to the new framing (branding/intro text keeps "Tako").

## Test plan
- [x] `npm run typecheck` — clean (all three tsconfigs)
- [x] `npm run test` — 378 pass, incl. 6 `slimCard — explicit exportable flag` cases + updated `buildSummary` / `tako_available_data` coverage-prose assertions (counts only, names not repeated)
- [x] `npm run registry:check` / `schemas:check` — in sync
- [x] Verified `exportable` survives output validation (`takoCardSchema` is `.loose()` + now explicit) and doesn't collide with any generated/backend field
- [x] `agent.json` / `server.json` valid JSON after description edits


noahjax (approved):
Approving — the `exportable` change correctly consumes the authoritative backend field from TakoData/tako#27989.

- `card.exportable ?? card.content != null` has the right precedence and, importantly, preserves an authoritative `false` (nullish coalescing does not fall back on `false`), with a content-presence fallback for older backends.
- Strong test coverage: six `slimCard` cases including the "backend `exportable: false` wins even when content is present" case.
- `tako_contents` precondition + 403/404 guidance and the generated `registry/server.json` are synced and accurate against the backend contract.

Two non-blocking notes: (1) the PR bundles the unrelated `tako_available_data` reframe — cleaner history if split, but both halves are low-risk; (2) web-native metrics (e.g. SimilarWeb traffic) are retrievable but come back `exportable: false` — the flag correctly steers the model away from a guaranteed export 403, so the messaging is consistent.

PR #152, opened 2026-07-22
feat(tools): full coverage names in tako_available_data; dedupe from summary

25eliu:
## Why

`tako_available_data` was silently truncating coverage: an entity with 15 metrics showed only 8 names (`PREVIEW = 8`), even though the backend drill already fetched 50. The names are the tool's primary payload — each is the exact term an agent reuses in a follow-up `tako_search` — so the cap hid real, requestable data.

Separately, every shown name was paid for **twice** in the calling agent's context: once enumerated in the prose `summary` ("15 metrics incl. AI Adoption Rate, …") and again in `matches[].coverage.names`, both riding in the same serialized text block.

## What

1. **Cap raised 8 → 50**, matched to the `graph/related` drill's fetch limit. The drill now passes `limit: PREVIEW` instead of a hardcoded 50, so the fetch limit and display cap can't drift (and the test asserts `String(PREVIEW)`, not a magic `"50"`).
2. **Names de-duplicated out of the summary.** The prose is now a narrative shell — header, per-match counts (`47 metrics.` / `tracked for 63 entities.` / `250+ metrics.` when capped), gap/unavailable phrasing, and a next-step line pointing at `coverage.names` with one concrete example (e.g. `"Apple Inc. Revenue"`). Names appear exactly once, in `matches[].coverage.names`.
3. Tool description tip updated: exact names live in `coverage.names` (previously "in the summary").

Net effect: a typical entity response carries **all** fetched names instead of 8, at roughly the same token cost as before. `truncated`/`capped` semantics, low-signal metric reordering (`(Normalized)`/`Account Code` pushed to the back), and the `found`/gap-header contract are unchanged.

## Test plan

- [x] `npx vitest run` — 372/372 pass (44 in the two touched suites)
- [x] `npm run typecheck` — clean across all three tsconfigs
- [x] Code review pass: no CRITICAL/HIGH/MEDIUM findings; the one LOW (magic `"50"` in the limit assertion) fixed in this PR
- [x] Repo grep: no stale references to the 8-cap or the old "incl. … (+N more)" phrasing in code, tests, README, or llms docs
- [ ] Spot-check on staging: `tako_available_data("OpenAI")` returns all 15 metric names with `truncated: false`


jed326 (approved):
Approved. Correctness, tests, and internal consistency are clean; `registry:check` confirms the `server.json` description is in lockstep with the tool source. The only nits are two wording over-claims in the PR description (no code impact):

1. "Names appear exactly once" — the first name still rides in the next-step example (`nextStepExample`), so it appears twice. Intentional and honestly commented in code; just the summary wording overstates.
2. "roughly the same token cost" — closer to ~3x for large entities (47 metrics ≈ 48 vs ~16 name instances before). The completeness trade-off is defensible; consider reframing.

Neither blocks merge.

PR #154, opened 2026-07-22
refactor(tools): Tako-proprietary-data attribution in available-data summary + fan-out tip

25eliu:
## Summary

Two small wording changes to `tako_available_data`, both aimed at steering downstream LLM behavior:

1. **Summary attribution** — `buildSummary`'s positive headers now read `Tako's proprietary data has live, continuously-updated coverage of N matches for "X":` instead of `Tako has live data on ...`. Models echo this header nearly verbatim when answering users, so making "Tako's proprietary data" the grammatical subject stops them from attributing results to a generic "this dataset". The gap header (no coverage) is unchanged — the tool still never claims coverage it doesn't have.

2. **Fan-out tip** — new first tip in the tool description: *"One metric across many entities → one metric-first call; one entity across many metrics → one entity-first call. The coverage.names answer all of them at once — never loop one call per name."* Observed agents checking e.g. app downloads for ChatGPT/Claude/Gemini/Grok/Perplexity with five separate entity-first calls when one metric-first call answers all of them.

Registry regenerated for the description change (`registry/server.json`, `_registry.ts` unchanged apart from the description string).

## Test plan

- [x] `npx vitest run` — 378/378 pass (header assertions updated; gap-regression tests now guard `not.toContain("Tako's proprietary data has")`)
- [x] `npm run registry:check` — ok (12 tools)


workers/src/tools/_available_data.ts
```diff
@@ -231,19 +231,23 @@ export function buildSummary(input: {
...
+  // The header only claims coverage for matches that actually have some — a
+  // resolved node with no coverage (or a failed drill) must not be advertised
+  // as data, per the tool's contract ("a match with no coverage is a real
+  // answer — report the gap"). "Tako's proprietary data" is the grammatical
+  // subject on purpose: downstream models echo this header nearly verbatim to
+  // the user, and this framing keeps them from attributing the answer to a
+  // generic "dataset".
   const withData = matches.filter(hasLiveCoverage).length;
   const matchesOf = `${n} ${plural(n, "match", "matches")} for "${query}"`;
+  const covers = "Tako's proprietary data has live, continuously-updated coverage of";
```
robertabbott, line 243: The header now leads with "live, continuously-updated coverage" for every match that has any coverage, and `selectCoverage` counts coverage regardless of each metric's real update cadence — quarterly financials, annual macro series, and one-off historical datasets all count as "covered." Since the whole point is that downstream models echo this header nearly verbatim, is escalating the old "live data" to "live, continuously-updated" intended across all match types, or should it stay closer to the freshness the tool can actually vouch for per match?

workers/src/tools/tako_available_data.ts
```diff
@@ -49,6 +49,7 @@ const DESCRIPTION = [
   "Works on an entity (a company, person, or place → the metrics tracked on it, e.g. Tesla) or a metric (→ the entities it is tracked across, e.g. Inflation Rate).",
   "",
   "Tips:",
+  "One metric across many entities → one metric-first call; one entity across many metrics → one entity-first call. The coverage.names answer all of them at once — never loop one call per name.",
```
robertabbott, line 52: The tip tells the agent "The coverage.names answer all of them at once — never loop one call per name," but `coverage.names` is capped at `PREVIEW` (50) and sets `truncated`/`capped` when more exist server-side. For a metric tracked across more than 50 entities, one metric-first call does not return every name. Should the tip point at the `truncated`/`capped` signal so an agent doesn't treat a truncated list as complete, or is >50 names rare enough here that the simpler phrasing wins?

robertabbott (approved):

PR #158, opened 2026-07-23
feat: Claude Code plugin — hosted MCP server + bundled research skills

25eliu:
## What

Adds a Claude Code plugin so users can install Tako in one step:

```bash
claude plugin marketplace add TakoData/tako-mcp
claude plugin install tako@tako
```

- **`.claude-plugin/plugin.json`** — declares the hosted server (`https://mcp.tako.com/mcp?tools=agent`, Bearer auth via `${TAKO_API_KEY}` env interpolation). `?tools=agent` because the bundled skills use `tako_agent` for ranking questions.
- **`.claude-plugin/marketplace.json`** — makes this repo installable directly as a marketplace (`tako@tako`), and is the artifact needed to submit to `claude-plugins-community` (and pitch for `claude-plugins-official`, the Exa-style `tako@claude-plugins-official` path).
- **`skills/*/SKILL.md`** — the three README agent skills (financial research, web & app traffic, macroeconomics) extracted verbatim, so the plugin ships them natively instead of copy-paste blocks.
- **README** — plugin install path added to the Claude Code section (with the `TAKO_API_KEY` vs `TAKO_API_TOKEN` distinction and a note to remove a pre-existing `claude mcp add` server).

## Verification

- `claude plugin validate .` passes.
- End-to-end tested locally: `claude plugin marketplace add <path>` → `claude plugin install tako@tako` → skills present in the plugin cache → uninstalled cleanly.
- Skills diffed byte-identical against the README source blocks.
- Reviewed: no CI/codegen collisions (`skills/`, `.claude-plugin/` untouched by gen-registry/regenerate workflows).

## Follow-ups (deliberately out of scope)

- ~~plugin.json version~~ — resolved per review discussion: `version` is pinned to release-please (`.claude-plugin/plugin.json` added to `extra-files`, seeded at 0.11.0). Users receive plugin updates when a release is cut and bumps the version, with changelog — not on every push to main.
- No CI drift-check between README skill blocks and `skills/*/SKILL.md` yet; consider making `skills/` the source of truth and generating the README blocks.
- AGENTS.md repo-layout section doesn't mention `.claude-plugin/` / `skills/` yet.
- Submit to `claude-plugins-community` via https://platform.claude.com/plugins/submit once merged.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,6 @@
 {
   "name": "tako",
   "description": "Tako's proprietary knowledge graph as MCP tools — live financial, macroeconomic, and company data as interactive, citation-backed charts — plus bundled research skills.",
-  "version": "0.1.0",
   "author": {
     "name": "Tako",
     "url": "https://tako.com"
@@ -9,7 +8,15 @@
   "homepage": "https://docs.tako.com/documentation/integrations/mcp-server",
   "repository": "https://github.com/TakoData/tako-mcp",
   "license": "MIT",
-  "keywords": ["tako", "mcp", "search", "finance", "macroeconomics", "charts", "data"],
+  "keywords": [
+    "tako",
+    "mcp",
+    "search",
+    "finance",
+    "macroeconomics",
+    "charts",
+    "data"
+  ],
   "mcpServers": {
     "tako": {
       "type": "http",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,0 +1,22 @@
+{
+  "name": "tako",
+  "description": "Tako's proprietary knowledge graph as MCP tools — live financial, macroeconomic, and company data as interactive, citation-backed charts — plus bundled research skills.",
+  "version": "0.1.0",
+  "author": {
+    "name": "Tako",
+    "url": "https://tako.com"
+  },
+  "homepage": "https://docs.tako.com/documentation/integrations/mcp-server",
+  "repository": "https://github.com/TakoData/tako-mcp",
+  "license": "MIT",
+  "keywords": ["tako", "mcp", "search", "finance", "macroeconomics", "charts", "data"],
+  "mcpServers": {
+    "tako": {
+      "type": "http",
+      "url": "https://mcp.tako.com/mcp?tools=agent",
+      "headers": {
+        "Authorization": "Bearer ${TAKO_API_KEY}"
+      }
+    }
+  }
+}
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,7 @@
 {
   "name": "tako",
   "description": "Tako's proprietary knowledge graph as MCP tools — live financial, macroeconomic, and company data as interactive, citation-backed charts — plus bundled research skills.",
+  "version": "0.11.0",
   "author": {
     "name": "Tako",
     "url": "https://tako.com"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -18,12 +18,21 @@
     "charts",
     "data"
   ],
+  "userConfig": {
+    "tako_api_key": {
+      "type": "string",
+      "title": "Tako API key",
+      "description": "Your Tako API key from https://tako.com/console/api-keys",
+      "sensitive": true,
+      "required": true
+    }
+  },
   "mcpServers": {
     "tako": {
       "type": "http",
       "url": "https://mcp.tako.com/mcp?tools=agent",
       "headers": {
-        "Authorization": "Bearer ${TAKO_API_KEY}"
+        "Authorization": "Bearer ${user_config.tako_api_key}"
       }
     }
   }
--- a/README.md
+++ b/README.md
@@ -34,17 +34,14 @@ Every request authenticates with a Bearer token. **[Get your API key](https://ta
 <details>
 <summary><b>Claude Code</b></summary>
 
-**Plugin (recommended)** — installs the MCP connection plus Tako's bundled [research skills](#agent-skills) in one step. The plugin reads your key from `TAKO_API_KEY` (not `TAKO_API_TOKEN`):
+**Plugin (recommended)** — installs the MCP connection plus Tako's bundled [research skills](#agent-skills) in one step. Claude Code prompts for your API key when the plugin is enabled and stores it securely — no `export` or environment variable to manage:
 
 ```bash
-# Persist your key so every Claude Code launch can read it
-echo "export TAKO_API_KEY='<your-token>'" >> ~/.zshrc && source ~/.zshrc
-
 claude plugin marketplace add TakoData/tako-mcp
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -34,6 +34,19 @@ Every request authenticates with a Bearer token. **[Get your API key](https://ta
 <details>
 <summary><b>Claude Code</b></summary>
 
+**Plugin (recommended)** — installs the MCP connection plus Tako's bundled [research skills](#agent-skills) in one step. The plugin reads your key from `TAKO_API_KEY` (not `TAKO_API_TOKEN`):
+
+```bash
+export TAKO_API_KEY='<your-token>'
```
robertabbott: `export TAKO_API_KEY=...` sets the key only for the current shell, but `plugin.json` stores the header as the literal `${TAKO_API_KEY}` and expands it at **connection time on every Claude Code launch** — not once at install. The direct `claude mcp add --header "Authorization: Bearer $TAKO_API_TOKEN"` path below differs in a way that matters: the shell expands `$TAKO_API_TOKEN` at add-time, so the resolved key is baked into the stored config and survives restarts. With the plugin path, a user who runs this `export` in one terminal, installs, then later starts Claude Code from a fresh shell ends up sending `Authorization: Bearer ${TAKO_API_KEY}` verbatim — a silent auth failure surfaced only as a missing-variable warning in `claude mcp list`. Should the README tell users to persist `TAKO_API_KEY` in their shell profile (or use a `${TAKO_API_KEY:-...}` fallback), rather than a one-off `export`?

robertabbott (approved):

.claude-plugin/plugin.json
```diff
@@ -0,0 +1,29 @@
...
+    "macroeconomics",
+    "charts",
+    "data"
+  ],
+  "mcpServers": {
+    "tako": {
+      "type": "http",
+      "url": "https://mcp.tako.com/mcp?tools=agent",
+      "headers": {
+        "Authorization": "Bearer ${TAKO_API_KEY}"
```
noahjax: **[Issue]** `${TAKO_API_KEY}` is not interpolated in a plugin manifest — only `${CLAUDE_PLUGIN_ROOT}` / `${CLAUDE_PLUGIN_DATA}` / `${CLAUDE_PROJECT_DIR}` expand here, so the server receives the literal string and Bearer auth fails.

Use `userConfig` instead — Claude Code prompts for the key at enable time and stores it securely:

```json
"userConfig": {
  "tako_api_key": {
    "type": "string",
    "title": "Tako API key",
    "description": "Your Tako API key from https://tako.com/...",
    "sensitive": true,
    "required": true
  }
},
"mcpServers": {
  "tako": {
    "type": "http",
    "url": "https://mcp.tako.com/mcp?tools=agent",
    "headers": { "Authorization": "Bearer ${user_config.tako_api_key}" }
  }
}
```

Note the reference token is `${user_config.tako_api_key}` (lowercase `user_config`), distinct from the camelCase `userConfig` manifest key. Worth confirming with one real `tako_search` through the installed plugin before merge.

README.md
```diff
@@ -34,6 +34,19 @@ Every request authenticates with a Bearer token. **[Get your API key](https://ta
 <details>
 <summary><b>Claude Code</b></summary>
 
+**Plugin (recommended)** — installs the MCP connection plus Tako's bundled [research skills](#agent-skills) in one step. The plugin reads your key from `TAKO_API_KEY` (not `TAKO_API_TOKEN`):
```
noahjax: **[Issue]** Once the key comes from `userConfig`, this whole block changes: there is no `TAKO_API_KEY` env var and no `export` step — Claude Code prompts for the key when the plugin is enabled. Drop the "reads your key from `TAKO_API_KEY`" note and the `export` line, and just show `marketplace add` → `install`, mentioning the interactive key prompt. (The "Or add the MCP server directly" fallback below stays as-is — `${TAKO_API_TOKEN}` is correct there because standard `.mcp.json` *does* expand shell vars.)

noahjax (changes requested):
Nice, clean packaging PR — verified the three `SKILL.md` files are byte-identical to their README source blocks, and confirmed the new files sit outside all repo automation (`regenerate.yml` triggers only on `openapi/sdk.yaml`; release-please tracks a fixed 7-file list). The `version` omission and `?tools=agent` are both correct and intentional.

One blocker: `${TAKO_API_KEY}` in the plugin's `mcpServers` headers won't expand. Shell-env `${VAR}` interpolation works in standard `.mcp.json` (what `claude mcp add` writes) but **not** in a plugin manifest — per the [plugins reference](https://code.claude.com/docs/en/plugins-reference), plugin MCP configs substitute only `${CLAUDE_PLUGIN_ROOT}`, `${CLAUDE_PLUGIN_DATA}`, `${CLAUDE_PROJECT_DIR}`. The literal string `Bearer ${TAKO_API_KEY}` gets sent, so auth fails on the first tool call. The E2E test in the description (install → skills present → uninstall) never exercised an authenticated call, which is why it passed.

Recommended fix is `userConfig` — it also removes the `TAKO_API_KEY`-vs-`TAKO_API_TOKEN` divergence entirely and prompts the user for the key at install (no manual `export`). Details inline.

jed326 (approved):
Approving — plugin, marketplace, and release-please wiring are correct and validated end-to-end (`claude plugin validate` passes; `userConfig` + `${user_config.tako_api_key}` interpolation is right; version 0.11.0 matches the manifest).

One non-blocking doc note (follow-up, not for this PR):

**README `## Agent Skills` section (~line 312):** now that the new **Plugin (recommended)** block auto-bundles these three skills, this section still tells users to `claude mcp add` and paste each block manually — two overlapping install paths with no framing of which to use. Consider a one-line note framing it as the manual alternative, e.g. _"If you installed the Tako plugin above, these skills are already bundled — the blocks below are the manual alternative for users who added the server with `claude mcp add`."_

noahjax (approved):
Re-reviewed at `85f5460`. Both blockers from my previous review are resolved:

- **Auth** — `plugin.json` now uses `userConfig.tako_api_key` (`sensitive`/`required`) with `Bearer ${user_config.tako_api_key}`, so Claude Code prompts for the key at enable time and injects it. This is the correct plugin-context mechanism and also removes the `TAKO_API_KEY`/`TAKO_API_TOKEN` split.
- **README** — rewritten to describe the interactive prompt (no `export`); the direct `claude mcp add` fallback correctly keeps `${TAKO_API_TOKEN}` (standard `.mcp.json` does expand shell vars).

The `version` reversal (omit → pin `0.11.0`) is wired correctly — `.claude-plugin/plugin.json` was added to release-please `extra-files`, so it bumps in lockstep and the value matches the current manifest. Skills remain byte-identical to their README source blocks; CI green.

Approving. One non-blocking note: the config is correct per the docs, but the auth path itself has not been exercised yet — worth one real `tako_search` through the installed plugin to confirm the prompt → Bearer injection works end-to-end. The PR-body follow-ups (README/skills drift-check, AGENTS.md layout, community-marketplace submission) remain good post-merge items.

PR #164, opened 2026-07-24
docs: align registry descriptions with README reframe (web search + licensed data)

25eliu:
## Summary
- Replaces the old "Search, answer, and fetch data from Tako's knowledge base of 100K+ charts…" description with the new README framing in all registry discovery files:
  - `registry/metadata.json` (source of truth)
  - `registry/server.json` (regenerated via `npm run registry:gen`)
  - `registry/smithery.yaml`
  - New description: "Tako MCP gives your agent industry-leading live web search plus licensed data that the open web does not have. That includes company financials, macroeconomic indicators, web and app traffic, sports, US government spending, and more."
  - These changes were originally pushed to #163 but landed after it was merged, so this PR carries them.
- Adds a **LobeHub listing**: new `registry/lhm.plugin.json` descriptor + LobeHub badge in the README.
- Types `tako_visualize` component configs as a **discriminated union** on `component_type` — `header`, `categorical_bar`, `generic_timeseries`, `table`, `financial_boxes`, and `pie` carry their required fields inline; other types remain documented passthrough. Untyped optional styling fields are preserved (`.passthrough()`), and the registry catalog is regenerated to match.

## Test plan
- [x] `npm run registry:check` passes (12 tools)
- [x] `tako_visualize` schema tests: typed configs accepted, missing required fields rejected, untyped optional fields preserved, passthrough types still accept freeform configs


workers/src/tools/tako_visualize.ts
```diff
@@ -54,36 +54,193 @@ const COMPONENT_TYPES = [
...
+
+// --- Config sub-shapes for the common component types ---
+// Each mirrors its backend Pydantic model in
+// app/backend/knowledge/api/ga/v1/thinviz/types.py and stays `.passthrough()`
+// so the many optional styling fields (and any additive backend fields) flow
+// through untyped rather than being rejected. Only the required fields and the
+// most-used optionals are typed here — enough for an agent to build a valid
+// call without guessing.
+
+const categoricalDataPoint = z.object({
```
robertabbott: `categoricalDataPoint` is a plain `z.object()`, while every sibling schema here (`categoricalDataset`, `timeseriesDataset`, `financialBoxItem`, `tableColumn`, and all six typed configs) is `.passthrough()`. Zod strips unknown keys on a plain object rather than rejecting them, and the SDK passes the *parsed* result to the handler (`validateToolInput` returns `parseResult.data`, and `mcp.ts:436` registers `inputSchema.shape`), so the stripped value is what `buildVisualizeBody` sends to Django.

The backend's `CategoricalDataPoint` (`thinviz/types.py:748`) accepts two more fields: `highlight: bool` and `color: str | None`. So this parses cleanly, loses `highlight`, and renders every bar muted with nothing accented — no error anywhere:

```json
{"component_type": "categorical_bar", "config": {"series_color_theme": "uniform", "datasets": [{"label": "Revenue", "data": [{"x": "Q1", "y": 100}, {"x": "Q2", "y": 140, "highlight": true}]}]}}
```

The new passthrough test only asserts config-level fields, so it stays green. Should `categoricalDataPoint` (and `timeseriesDataPoint`, which has the same omission but no fields to lose today) be `.passthrough()` too?

workers/src/tools/tako_visualize.ts
```diff
@@ -54,36 +54,193 @@ const COMPONENT_TYPES = [
...
+        `Config for \`${component_type}\` — shape varies by type; see Tako's "Visualize Your Data" docs. Passed through and validated server-side.`,
+      ),
+  });
+
+// Discriminated on `component_type` so each type advertises its own config
+// shape (the six common types typed; the rest documented passthrough). This
+// replaces a single untyped `config` bag, so an agent — or a tool scanner —
+// sees the required fields per type instead of "untyped object".
+const componentSchema = z
+  .discriminatedUnion("component_type", [
```
robertabbott, line 219: I measured the emitted JSON Schema on both refs with the repo's zod: `tako_visualize`'s `inputSchema` goes from 2,022 to 16,568 chars (~506 → ~4,142 tokens), and the whole 12-tool `tools/list` payload from 22,413 to 37,137 chars (~5.6k → ~9.3k tokens) — +66%, essentially all from this one tool.

That is context every connected agent carries on every turn, including the majority of sessions that never call `tako_visualize`. Is the scanner fix worth ~3.7k tokens of standing context, or would a smaller union (typing only `categorical_bar` and `generic_timeseries`, or trimming the per-field `.describe()` strings) satisfy it?

workers/src/tools/tako_visualize.ts
```diff
@@ -54,36 +54,193 @@ const COMPONENT_TYPES = [
...
+const typedComponent = <T extends (typeof COMPONENT_TYPES)[number]>(
+  component_type: T,
+  config: z.ZodTypeAny,
+) => z.object({ component_type: z.literal(component_type), component_variant: componentVariant, config });
+
+// A member whose config stays a passthrough record — the shape is documented in
+// Tako's "Visualize Your Data" docs and validated server-side, but not typed
+// here (these types are either rarely used from an agent or have large, highly
+// variable configs). The typed common types above cover the everyday path.
+const passthroughComponent = <T extends (typeof COMPONENT_TYPES)[number]>(component_type: T) =>
```
robertabbott, line 205: All 20 members are covered today — I checked programmatically. But `COMPONENT_TYPES` is now referenced only as a generic constraint (`T extends (typeof COMPONENT_TYPES)[number]`), which validates that a listed name is *valid* without checking that every name is *listed*.

The `// Keep in sync if the backend adds a builder` note at line 27 used to mean "add to this array" and `z.enum` did the rest; now it silently also means "add a union member," and skipping the second half rejects the new type at the edge with no compile error and no failing test. Would a type-level exhaustiveness assert, or a test that parses one component per `COMPONENT_TYPES` entry, close that gap?

registry/lhm.plugin.json
```diff
@@ -0,0 +1,244 @@
+{
+  "identifier": "takodata-tako-mcp",
+  "name": "Tako MCP",
+  "version": "0.12.0",
```
robertabbott: Three things drift here with nothing to catch them:

- `version` is `0.12.0` while the repo is `0.13.0` (`version.txt`, `workers/package.json`, `registry/metadata.json`).
- The file is absent from `release-please-config.json` `extra-files`, which lists the other eight version-bearing files — so it will stay pinned.
- `tools[]` carries 4 of the 12 tools (`tako_answer`, `tako_available_data`, `tako_contents`, `tako_search`), hand-copied, while `gen-registry.ts` only writes `registry/server.json` and `_registry.ts`.

Since `lhm plugin publish` pushes this to a public listing, should it get a `release-please` entry and be generated by `gen-registry.ts` — and is the 4-tool subset deliberate curation or an incomplete copy?

workers/src/tools/tako_visualize.ts
```diff
@@ -54,36 +54,193 @@ const COMPONENT_TYPES = [
...
-          .describe(
-            "Data/configuration object for this `component_type`; its shape varies by type and is validated server-side. " +
-              'Examples — `header`: `{"title": "Monthly Revenue"}`. ' +
-              '`categorical_bar`: `{"datasets": [{"label": "Sales", "units": "USD", "data": [{"x": "NA", "y": 500}, {"x": "EU", "y": 300}]}]}` ' +
-              "(each dataset is a labeled series; `data` is an array of `{x, y}` points). " +
-              "For shapes of other types (e.g. `generic_timeseries`, `table`, `financial_boxes`, `pie`) see Tako's 'Visualize Your Data' docs.",
-          ),
-      }),
-    )
+    .array(componentSchema)
```
robertabbott, line 262: The SDK's `validateToolInput` throws `McpError(InvalidParams)` *before* the callback registered in `registerTool` runs, and `mcp.ts:1227` notes that the outer catch never sees these ("The SDK handles JSON-RPC validation errors internally"). So a call the new typed configs reject produces a `-32602` to the client and zero Worker-side signal — no `console.error`, no metric.

Given this PR narrows what the tool accepts, is there a way to see whether the tightening is rejecting real agent traffic, or would that only show up as a user complaint?

registry/server.json
```diff
@@ -4,7 +4,7 @@
   "name": "tako-mcp",
   "version": "0.13.0",
   "title": "Tako MCP Server",
-  "description": "Search, answer, and fetch data from Tako's knowledge base of 100K+ charts covering economics, finance, demographics, sports, markets, and weather; and run a multi-step Answer Agent for deeper questions.",
+  "description": "Tako MCP gives your agent industry-leading live web search plus licensed data that the open web does not have. That includes company financials, macroeconomic indicators, web and app traffic, sports, US government spending, and more.",
```
robertabbott, line 7: The commit message says the regen was so "the typed schema reaches the metadata scanners read," but `flattenParameters` in `gen-registry.ts` emits top-level properties only — `components` here is still `{"type": "array", "description": ..., "required": true}` with no item schema, and the only change to this tool in the diff is the description text.

If OpenAI's scanner reads the live `tools/list`, the union does reach it and this file is irrelevant to the fix; if it reads `server.json`, the fix does not land here at all. Which one flagged it? That also decides whether the schema growth on `tako_visualize.ts` bought anything.

workers/src/index.ts
```diff
@@ -71,6 +71,27 @@ export default {
...
+    // the OpenAI challenge above: unset binding falls through to the
+    // catch-all 404, and no CORS wrapper (server-to-server fetch).
+    if (
+      request.method === "GET" &&
+      url.pathname === "/.well-known/glama.json" &&
+      env.GLAMA_MAINTAINER_EMAIL
+    ) {
+      return new Response(
+        JSON.stringify({
+          maintainers: [{ email: env.GLAMA_MAINTAINER_EMAIL }],
```
robertabbott: The repo root already has a `glama.json` (`$schema: glama.ai/mcp/schemas/server.json`) with `maintainers: ["jed326", "robertabbott", "25eliu"]` — bare GitHub usernames. This route serves a different shape at a different location: `maintainers: [{"email": ...}]`, with one entry.

Does Glama read the repo file for the source listing and the well-known URL for the remote-server claim, or do these two now compete? If both are live, should the root file gain Eric, or this one gain the other maintainers?

workers/wrangler.jsonc
```diff
@@ -118,7 +124,12 @@
         // `/.well-known/openai-apps-challenge`. If OpenAI rotates the
         // value, update here and redeploy. Not a secret — the URL is
         // public by design.
-        "OPENAI_APPS_CHALLENGE_TOKEN": "uXTBcUGnISWBPU2YXEn7IPIDh2ANMxLa1rkFvv3Z0c0"
+        "OPENAI_APPS_CHALLENGE_TOKEN": "uXTBcUGnISWBPU2YXEn7IPIDh2ANMxLa1rkFvv3Z0c0",
+        // Glama remote-server listing claim for `mcp.tako.com`. Glama
+        // matches this against the claiming Glama account's email —
+        // served as JSON from `/.well-known/glama.json`. Not a secret —
+        // the URL is public by design.
+        "GLAMA_MAINTAINER_EMAIL": "eric@trytako.com"
```
robertabbott: This is the first appearance of `eric@trytako.com` in the repo, and it lands in a public repo *and* on a public URL that exists to be fetched by third parties. Unlike `OPENAI_APPS_CHALLENGE_TOKEN` above it, this is a real personal mailbox rather than an opaque token.

Would a role alias (`mcp@trytako.com`) work for the Glama claim — same verification, no scraping surface, and ownership survives a personnel change?

workers/src/tools/tako_visualize.ts
```diff
@@ -54,36 +54,193 @@ const COMPONENT_TYPES = [
...
+const pieConfig = z
+  .object({
+    datasets: z
+      .array(categoricalDataset)
+      .min(1)
+      .describe("Slices as a labeled series (x = slice label, y = value); only the first dataset renders."),
+    title: z.string().optional().describe("Title shown in the hover tooltip."),
+  })
+  .passthrough();
+const timeseriesConfig = z
```
robertabbott, line 166: The `.min(1)` comment says the schema is deliberately stricter than the backend because "an empty chart is a meaningless card, so we reject it early." But `generic_timeseries` has both `datasets` and `datasets_by_interval` optional with no cross-field refine, so `{"component_type": "generic_timeseries", "config": {}}` passes; and `headerConfig` makes `title` optional even though the backend's own template (`views.py:183`) lists it as the one required field.

Is the either-or on the timeseries datasets meant to be enforced, or is the `.min(1)` strictness intentionally scoped to just the four configs that have it?

workers/src/tools/tako_visualize.ts
25eliu: Fixed in ebeae4f — both `categoricalDataPoint` and `timeseriesDataPoint` are now `.passthrough()`, and the tests include your exact repro: `highlight: true` (and `color`) on a point now survives all the way into the `buildVisualizeBody` output, asserted at point level rather than config level.

workers/src/tools/tako_visualize.ts
25eliu: We kept the full 6-config union but trimmed it (ebeae4f): describe strings cut down and the per-member `component_variant` describe dropped entirely — it was duplicated across all 20 union members. Measured with the repo's zod: the emitted input schema goes 17,322 → 11,567 chars (~4,331 → ~2,892 tokens), a 33% cut from the state you measured. The remaining delta over the pre-PR 2,022 chars is the structural cost of the union itself, which we're keeping for the per-type required fields it advertises.

workers/src/tools/tako_visualize.ts
25eliu: Both halves added in ebeae4f: a compile-time assert next to `componentSchema` — removing a member now fails tsc with `Type 'boolean' is not assignable to type '{ missingUnionMembersFor: "bubble" }'` (verified by deleting a member) — plus a runtime test that parses one minimal component per `COMPONENT_TYPES` entry, so the "add to the array" comment is enforced in both directions again.

registry/lhm.plugin.json
25eliu: All three fixed in de0b857: version bumped to 0.13.0; the file is in `release-please-config.json` `extra-files` now; and `gen-registry.ts` generates it (static descriptor fields stay in the file, `version` comes from `metadata.json`, `tools[]` is emitted from the live tool modules with full draft-7 schemas — `registry:check` guards drift). The 4-tool subset is deliberate curation of the core retrieval surface, now encoded as `LOBEHUB_TOOL_ALLOWLIST` with a comment saying exactly that.

workers/src/tools/tako_visualize.ts
25eliu: Addressed in 488db96. Since the SDK rejects before any callback runs, the signal comes from a tap after `transport.handleRequest`: the buffered JSON-RPC response (safe to clone under `enableJsonResponse: true`) is checked for `-32602`, and each rejection is `console.error`-ed with its method + tool name, matched back to the request by JSON-RPC id. Workers Logs now shows whether the tightening rejects real agent traffic — for every tool, without moving validation across the SDK boundary. Best-effort and never throws.

registry/server.json
25eliu: You're right — `flattenParameters` is unchanged and this file only picked up the description text; the flat parameter map is deliberate for this registry format. The surface the typed union actually lands on is the live `tools/list`, which is what connected agents (and any scanner that speaks MCP rather than reading our registry card) see — so that's where the fix does its work, and this file is incidental to it. The standing cost of that schema also just dropped ~33% (see the thread on line 217).

workers/src/index.ts
25eliu: They don't compete — they're two different Glama claim mechanisms: the repo-root `glama.json` claims the GitHub/source listing by GitHub username (Eric is already present there as `25eliu`), while `/.well-known/glama.json` claims the hosted server at mcp.tako.com by account email. Also, per the sibling thread on wrangler.jsonc, the served email is now the `mcp@trytako.com` role alias, which decouples the remote-server claim from any individual maintainer.

workers/wrangler.jsonc
25eliu: Switched to `mcp@trytako.com` in de0b857, with a comment in wrangler.jsonc documenting why (public-by-design URL, no personal scraping surface, survives personnel changes). One operational note: the alias must exist and be attached to the Glama account doing the claim before the listing claim is made.

workers/src/tools/tako_visualize.ts
25eliu: The either-or is now enforced (ebeae4f): `timeseriesConfig` has a `.refine` requiring exactly one of `datasets` / `datasets_by_interval` — `{}` and both-present each fail with a clear message — and `headerConfig.title` is now required to match the backend template. The comment above the configs now describes the full strictness scope (`.min(1)` + the refine + required title). Tests cover empty config, both-present, by-interval-only, and title-less header.

workers/wrangler.jsonc
25eliu: Correction to the above: reverted back to `eric@trytako.com` — the `mcp@trytako.com` alias doesn't exist, and the served email has to match the claiming Glama account. If we set up a role alias on the Glama account later, it's a one-line var change.

25eliu: Update: the whole Glama ownership-verification feature is removed in ae8b35a — no role alias exists, and we'd rather drop the remote-listing claim than publish a personal mailbox on a public URL. No email appears in the repo or on any served route now.

workers/src/index.ts
25eliu: Moot as of ae8b35a — the `/.well-known/glama.json` route, the `GLAMA_MAINTAINER_EMAIL` var, and their tests are removed entirely, so the repo-root `glama.json` (GitHub-username source listing) is the only Glama surface again.

robertabbott (approved):

PR #166, opened 2026-07-24
feat(mcp): render charts as inline MCP Apps widgets on Claude clients

25eliu:
## TL;DR
Charts now render inline in the claude.ai / Claude Desktop chat body via the existing MCP Apps widget. Previously Claude clients got a PNG image content block, which claude.ai renders collapsed inside the tool-call expander (anthropics/claude-code#53256) — users saw no charts.

## What changed
- Widget gate admits `claude` clients (widget-only; the existing `ui === undefined` mutual exclusion drops the PNG block automatically). Unknown clients keep the PNG block; ChatGPT keeps its interactive iframe widget, unchanged.
- **Client bucketing is evidence-based, not substring-broad**: `claude-code/…` UAs (Claude Code's HTTP MCP client and the Agent SDK it powers — verified empirically: `claude-code/2.1.220 (sdk-cli)`) are carved out to the `unknown` bucket and keep the portable inline PNG. The widget bucket matches the claude.ai / Claude Desktop custom-connector traffic, which is server-to-server from Anthropic's backend (`160.79.104.0/21`) as `Claude-User`. Direct `detectMcpClient` unit tests pin the observed UAs.
- Widget emits the open-spec `ui/notifications/size-changed` (claude.ai's height mechanism; `notifyIntrinsicHeight` covers ChatGPT only) — the missing piece behind the 2026-05 clipping that led to suppressing the widget. Applied to both the static and baked widget variants; the baked variant also re-notifies on `img` load.
- Widget CSP declares `resourceDomains` so the remote-`image_url` fallback `` loads when a chart exceeds the inline `image_data_url` cap. The cap itself is raised 250 KB → 400 KB raw: Tako charts run 50–300 KB, and the 250–300 KB band otherwise degrades to a text link on claude.ai (whose outer CSP blocks the cross-origin `image_url`). The ~400 KB-encoded failure that motivated the old cap was observed on ChatGPT, whose `extraMeta` hook is skipped entirely now, so the field never ships there.
- Security: the widget's inbound `ui/notifications/tool-result` / init-handshake postMessage handler is source-gated to `window.parent` **or `window.top`** (nested-wrapper hosts) so a co-installed sibling connector can't inject a spoofed chart image or click-through URL — a sibling's `event.source` is never parent or top. Rejected JSON-RPC messages `console.warn` instead of failing silently, since this gate guards Claude's only chart data path.
- **The widget bundle now executes in CI.** A second vitest project (`test/widget/`, plain Node + jsdom) mounts the bundle in a nested-iframe topology and drives the postMessage protocol: handshake completion, parent/top tool-result rendering, sibling-forgery rejection, `javascript:` scheme guard, no-chart collapse, and the baked variant's size-changed. (jsdom types are a local shim — `@types/jsdom` cannot be installed in this repo; vitest's `optional-types.d.ts` resolves it on sight and drags `lib.dom` into every tsc program, colliding with `@cloudflare/workers-types`. `npm run typecheck` is now four passes.)
- Observability: `console.warn` on `image_data_url` fetch failure (now Claude's primary chart path).
- Docs: swept stale "ChatGPT-only widget" comments; reconciled dynamic-URI docstrings; claude `tools/list` widget `_meta` pinned e2e (claude.ai loads the widget URI from registration metadata, so that's the load-bearing wire surface).

## Not changed
- ChatGPT tool results: byte-identical **except one additive change**: `_meta.ui.csp` now also carries `resourceDomains` (client-agnostic construction; ChatGPT's iframe branch never reads it). This is now pinned by an explicit test so a regression or a strict-validating Apps SDK build has a signal; the ChatGPT staging spot-check remains the behavioral gate. Everything else is pinned byte-identical (widget `_meta`, zero image blocks, zero PNG prefetch).
- Unknown clients (Cursor, Windsurf, …) **and now Claude Code / Agent SDK**: byte-identical to pre-PR (PNG block, no widget).
- Fetch count on Claude: still exactly 2 per chart call (PNG now fetched by `extraMeta` for the widget instead of `extraContentBlocks`) — enforced by tests that throw on any extra fetch.

## Known trade-offs (accepted)
- **Model visibility**: the PNG content block was model-visible; `_meta.image_data_url` is not. On Claude clients the model can no longer see the chart image while composing its answer — it works from the structured content (values, embed/image URLs) instead. Accepted: the user-visible chart is the point of this PR.
- **Messages API `mcp_servers` connector**: also egresses from Anthropic's backend; its UA is not publicly pinned. If it presents as `Claude-User` it is indistinguishable from claude.ai here and would get widget `_meta` it can't render. No UA-level fix available; revisit if log evidence shows it.

## Test plan
- [x] `npm test` — 456 workers-pool tests across 25 files + 8 executed-widget DOM tests; chart-gate matrix: claude → widget `_meta` + `_meta.image_data_url`, no image block, graceful degradation when the PNG fetch fails; `claude-code` → PNG path; chatgpt/unknown cases unmodified
- [x] `npm run typecheck` (four passes)
- [x] Local UA capture: `claude mcp add` against a header-logging server confirms Claude Code sends `claude-code/2.1.220 (sdk-cli)`
- [ ] Staging (auto-deploys on merge, or `npm run deploy:staging`): add `https://mcp.staging.tako.com/mcp` as a custom connector on **both claude.ai and Claude Desktop** (Desktop caches widget URIs beyond connector lifecycle — verify before the URI is burned in), run a chart query — chart renders inline in the chat body, height fits, no clipping; confirm a zero-result query collapses (no empty box)
- [ ] **Blocking** staging check: a large (>250 KB) chart renders on claude.ai — this is the empirical gate on the 400 KB cap and on whether claude.ai honors `csp.resourceDomains`
- [ ] ChatGPT spot-check on staging — widget still renders (gates the additive `resourceDomains` key)

Rollback: revert the gate commit (`bce1238`) alone — the widget/CSP/security commits are inert without it.

### DIFF
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -123,9 +123,10 @@ const JSON_SCHEMA_VALIDATOR = new CfWorkerJsonSchemaValidator();
  *
  * The match is intentionally loose: we don't care about exact UA
  * strings, just whether the request smells like one of the major
- * MCP-app hosts. Unknown UAs fall through to the "render the widget"
- * default — better to over-render than to hide the chart from a host
- * that supports it.
+ * MCP-app hosts. Unknown UAs fall through to the inline-PNG `image`
+ * content block, not the widget — that path is the portable fallback
+ * that works on the long tail of MCP hosts that don't (yet) implement
+ * MCP Apps, so an unrecognized client still gets a chart it can render.
  */
 // `McpClientKind` is defined in `tools/types.ts` (re-exported below)
 // so tool modules can reference it without a circular import on
@@ -610,11 +611,15 @@ function registerTool(
       );
     }
     // Optional dynamic-resource variant. When defined, the same widget
-    // also gets a `ResourceTemplate` registration, and per-call tool
-    // results point claude.ai's `_meta.ui.resourceUri` at a specific
-    // instance of that template (so the widget HTML can have the
-    // chart's image + dimensions baked in at fetch time, sidestepping
-    // claude.ai's "snapshot offsetHeight once on mount" behavior). See
+    // also gets a `ResourceTemplate` registration so a per-call
+    // `_meta.ui.resourceUri` could point at a specific instance of that
+    // template (baking the chart's image + dimensions into the widget
+    // HTML at fetch time, sidestepping a host's "snapshot offsetHeight
+    // once on mount" behavior). claude.ai does NOT use this today — see
+    // the "Conclusion" comment below: it ignores per-call resourceUri
+    // overrides and stays on the static URI + baked
+    // `_meta.image_data_url` from `extraMeta`. The template stays
+    // registered for a future host that does honor per-call URIs. See
     // `AppUiResource.dynamic` for the rationale.
     if (
       ui.dynamic !== undefined &&
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -111,11 +111,15 @@ const JSON_SCHEMA_VALIDATOR = new CfWorkerJsonSchemaValidator();
  * Detect the calling MCP client from the HTTP `User-Agent` header.
  *
  * Used to gate per-client behavior that we'd otherwise have to ask the
- * LLM to figure out from prose — specifically, suppressing the chart
- * widget on claude.ai (where the constrained iframe container clips
- * the chart and the LLM's markdown link is strictly cleaner UX) and
- * routing ChatGPT through the agent split pair (its Apps SDK doesn't
- * reset tool-call timeouts on progress notifications).
+ * LLM to figure out from prose — specifically, routing the chart to
+ * the MCP Apps widget for ChatGPT and Claude (the two hosts that render
+ * it inline — ChatGPT via the interactive iframe branch, Claude via the
+ * image branch, since claude.ai's host-side CSP ignores declared
+ * `frameDomains` and the widget's runtime `window.openai` check falls
+ * back to the static image there) while unknown clients fall back to
+ * the inline PNG `image` content block, and routing ChatGPT through
+ * the agent split pair (its Apps SDK doesn't reset tool-call timeouts
+ * on progress notifications).
  *
  * The match is intentionally loose: we don't care about exact UA
  * strings, just whether the request smells like one of the major
@@ -266,8 +270,8 @@ export function createMcpServer(
     "tako_agent",
   ]);
   // Tools whose `appUiResource` should NOT ship on ChatGPT (separate
-  // from the blanket non-ChatGPT suppression in `widgetSuppressed` —
-  // ChatGPT is the only client that gets the widget at all).
+  // from the blanket unknown-client suppression in `widgetSuppressed` —
+  // ChatGPT and Claude are the only clients that get the widget at all).
   // The mechanism is kept in place for future per-tool gating, but
   // is currently empty: `tako_search` ships its widget on ChatGPT
   // and handles the empty-result case by throwing an actionable
@@ -329,11 +333,12 @@ export function createMcpServer(
   }
 
   // Resources parity for server instances that registered NO widget
-  // resource (every non-ChatGPT client, since the chart widget is
-  // ChatGPT-only). The SDK only wires the `resources` capability and its
-  // request handlers on the first `registerResource` call — without this
-  // block, `resources/list` / `resources/read` answer JSON-RPC -32601 on
-  // non-ChatGPT instances, which capability-probing clients (Smithery's
+  // resource (unknown clients only — ChatGPT and Claude both register
+  // it, and the fallback below only fires when neither did). The SDK
+  // only wires the `resources` capability and its request handlers on
+  // the first `registerResource` call — without this block,
+  // `resources/list` / `resources/read` answer JSON-RPC -32601 on
+  // unknown-client instances, which capability-probing clients (Smithery's
   // scan, some hosts) surface as a failure. Same rationale as the empty
   // `prompts` registration above: an empty list is the spec-clean answer.
   // `resources/read` still errors (there is nothing to read) but with the
@@ -544,8 +549,12 @@ function registerTool(
       uiMeta.csp = csp;
     }
     // Resource registration. CSP-allowed iframe domains live on
-    // `_meta.ui.csp.frameDomains` (open MCP Apps spec). The bundle's
-    // `_meta.ui` is set in TWO places by design (matches the official
+    // `_meta.ui.csp.frameDomains` (open MCP Apps spec); its sibling
+    // `_meta.ui.csp.resourceDomains` allow-lists the origins the
+    // remote-image fallback (`<img src=image_url>`) may fetch/embed
+    // from — the widget's image branch is what actually renders on
+    // Claude, where the iframe branch never gets a chance to load. The
+    // bundle's `_meta.ui` is set in TWO places by design (matches the official
     // `@modelcontextprotocol/ext-apps` helper):
     //
     //   1. Resource registration metadata (third arg to
@@ -1144,9 +1153,9 @@ export async function handleMcpRequest(
     // connectors from referencing prod URLs and vice versa.
     const url = new URL(request.url);
     const requestOrigin = url.origin;
-    // Detect calling client from User-Agent so we can suppress the
-    // chart widget on claude.ai (constrained iframe container) and
-    // route ChatGPT through the agent split pair. See
+    // Detect calling client from User-Agent so we can route the chart
+    // widget to ChatGPT and Claude (unknown clients fall back to the
+    // inline PNG) and route ChatGPT through the agent split pair. See
     // `detectMcpClient` for the matching rules.
     const client = detectMcpClient(request.headers.get("user-agent"));
     // Opt-in tools: the agent is off by default and enabled per-connection
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -515,8 +515,15 @@ function registerTool(
 
   if (ui !== undefined) {
     const uiMeta: Record<string, unknown> = {};
+    const csp: Record<string, unknown> = {};
     if (ui.frameDomains && ui.frameDomains.length > 0) {
-      uiMeta.csp = { frameDomains: ui.frameDomains };
+      csp.frameDomains = ui.frameDomains;
+    }
+    if (ui.resourceDomains && ui.resourceDomains.length > 0) {
+      csp.resourceDomains = ui.resourceDomains;
+    }
+    if (Object.keys(csp).length > 0) {
+      uiMeta.csp = csp;
     }
     // Resource registration. CSP-allowed iframe domains live on
     // `_meta.ui.csp.frameDomains` (open MCP Apps spec). The bundle's
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -138,11 +138,27 @@ export type { McpClientKind };
 export function detectMcpClient(userAgent: string | null): McpClientKind {
   if (userAgent === null || userAgent === "") return "unknown";
   const ua = userAgent.toLowerCase();
-  // Claude.ai's MCP server-to-server connector identifies itself as
-  // either `Claude-User`, `claude-mcp-client`, or similar — match on
-  // any "claude" / "anthropic" substring. The user's own browser UA
-  // never reaches /mcp directly (claude.ai proxies through its
-  // backend), so this won't false-positive on user browsers.
+  // Claude Code's HTTP MCP client (and the Agent SDK, which runs Claude
+  // Code under the hood) sends `claude-code/<version> (sdk-cli)` —
+  // verified empirically 2026-07-27 against claude-code 2.1.220. It is
+  // a terminal, not an MCP Apps host: it cannot render the widget, and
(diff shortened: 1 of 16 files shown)


workers/src/mcp.ts
```diff
@@ -506,7 +529,8 @@ function registerTool(
   // guaranteed by the `ui === undefined` condition at the call-time
   // `extraContentBlocks` gate.)
   const widgetSuppressed =
-    options.client !== "chatgpt" || options.widgetSuppressedForTool === true;
+    (options.client !== "chatgpt" && options.client !== "claude") ||
```
noahjax, line 548: **Highest-risk item in the PR.** `detectMcpClient` (line 146) returns `"claude"` for any UA containing `claude` **or** `anthropic`. So this condition admits more than claude.ai and Claude Desktop — it also admits Claude Code's HTTP MCP client, the Agent SDK, and the Messages API `mcp_servers` connector. None of those render MCP Apps widgets, and because widget attachment suppresses the PNG block via `ui === undefined`, each of them now gets widget `_meta` it cannot render *and* loses the chart entirely.

The PR body accepts this for "Claude Code (terminal CLI)", but the blast radius is the whole Anthropic-first-party non-Apps surface, and the safe-default direction inverted: the PNG path worked nearly everywhere; the widget path works on ~2 of N Claude surfaces.

Suggestions, in preference order:

1. **Check what each surface actually sends before merging.** If Claude Code sends `claude-cli/...`, the discriminator already exists and a third bucket is cheap now. If it sends no UA at all it lands in `unknown` and keeps the PNG — in which case the stated trade-off is moot and the PR body should say so.
2. **Consider inverting the match** — allowlist the two known Apps-host connector UAs rather than the broad `claude`/`anthropic` substring, so anything unverified keeps the portable PNG default.

Deferring to "a future UA special-case" leaves a silently chart-less path on a public API surface.

workers/src/tools/_chart_widget.ts
```diff
@@ -978,6 +1047,11 @@ export function buildChartAppUiResource(
     // tool also writes into `embed_url`, so the two move together. No
     // wildcards: the widget only ever embeds Tako's own embed page.
     frameDomains: [webBase],
+    // Remote-image fallback (`<img src=image_url>`) loads from the API
+    // base; the primary data-URI path needs no CSP. webBase covers any
+    // future web-hosted asset. Deduped — staging/prod use one host for
+    // both.
+    resourceDomains: [...new Set([webBase, resolvePublicApiBase(env)])],
```
noahjax, line 1083: I do not think this achieves what the test plan assumes. Two facts in this same file argue against it:

- Line ~543 documents that claude.ai's outer-document CSP is `img-src 'self' blob: data:` and blocks the cross-origin `image_url`.
- The PR's core premise is that claude.ai **ignores** the declared `csp.frameDomains`. If the host ignores one `_meta.ui.csp` key, honoring its sibling needs evidence rather than assumption.

The numbers make this concrete: `MAX_INLINE_DATA_URL_BYTES` is 250 KB (line 64) while the comment at line 52 states "Tako chart PNGs run 50-300 KB". Charts in the 250–300 KB band fetch fine, get dropped by the cap, fall through to `validHttpImage`, get CSP-blocked, and land in the `image.error` handler — the user sees "Open interactive chart →" text, no chart. Not a regression from `main`, but a real hole in what is now Claude's *primary* chart path, with no lever in this PR.

- Treat the staging check "confirm a large (>250 KB) chart still renders" as blocking, not cosmetic.
- If `resourceDomains` turns out to be ignored: either raise the cap (the 250 KB rationale is claude.ai rejecting oversize tool results, but `_meta` is not tokenized — the real ceiling is the host's message-size limit, worth measuring), or re-request the PNG at reduced width when the first fetch comes back oversize. Either beats a text link.

workers/src/tools/_chart_widget.ts
```diff
@@ -727,19 +760,28 @@ const WIDGET_HTML = `<!doctype html>
...
     var msg = event.data;
     if (!msg || typeof msg !== "object") return;
+    // MCP Apps host messages (handshake + tool-result) come from the
+    // parent frame. Reject anything else — a co-installed sibling
+    // connector's iframe can postMessage to us via parent.frames[] and
+    // would otherwise be able to inject a spoofed tool-result (fake chart
+    // image + arbitrary click-through URL; render() is one-shot). The
+    // \`tako-embed-height\` branch below is separately origin-gated to the
+    // embed iframe, so leave it reachable regardless.
+    var fromHost = event.source === window.parent;
```
noahjax: The gate is the right idea and it does block the sibling-connector injection it targets: a sibling connector's iframe is a *child* of the host, so its `event.source` is never `window.parent`.

The risk is that this now guards the **only** data path Claude has, and nothing executes it in CI. If any Claude surface delivers `ui/notifications/tool-result` from a context that is not `window.parent` — a nested wrapper frame, or `window.top` directly — `render()` never fires and the widget sits on "Loading chart…" forever. Silent failure, and the unit tests (string `toContain` assertions on a template literal) cannot see it.

That matters more than usual here because of the note at lines 71-83 in this file: Claude Desktop caches resource URIs beyond connector lifecycle and "once a URI ships, it's effectively permanent". A broken bundle on Claude Desktop is expensive to walk back.

- Verify Claude Desktop explicitly, not just claude.ai.
- Consider widening to `event.source === window.parent || event.source === window.top` if that holds for the hosts you test, or add a `console.warn` on a rejected JSON-RPC message so `wrangler tail` shows the drop instead of failing silently.

workers/src/tools/_chart_widget.test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, expect, it } from "vitest";
+
+import type { Env } from "../env.js";
+import { buildChartAppUiResourceFromOutputPubId } from "./_chart_widget.js";
+
+const ENV: Env = { DJANGO_BASE_URL: "https://staging.trytako.com" };
+
+describe("chart widget HTML", () => {
+  it("notifies height via the MCP Apps size-changed notification", () => {
+    const ui = buildChartAppUiResourceFromOutputPubId(ENV);
```
noahjax, line 13: The PR body says size-changed was "applied to both the static and baked widget variants", but `buildChartAppUiResourceFromOutputPubId(...).html` is `WIDGET_HTML` — the **static** variant only. The baked variant's change has zero coverage. `buildBakedWidgetHtml` is already exported via `__chart_widget_test_only__` (line ~1126), so covering it is one line.

More broadly: these are `toContain` assertions against a template string the test suite never executes. They confirm the notification text is present, not that the widget works. A jsdom test that `eval`s the widget script and drives `postMessage` would cover the source gate, the init handshake, and the image branch in one go — cheap relative to what it protects, and it turns the source gate from an untested assumption into a regression test.

workers/src/tools/_chart_widget.ts
```diff
@@ -197,12 +197,28 @@ function buildBakedWidgetHtml(opts: {
...
         if (h > 0) window.openai.notifyIntrinsicHeight(h);
       }
     } catch (e) { /* host gone — nothing to do */ }
+    // MCP Apps open-spec mechanism (claude.ai, VS Code Insiders, Goose):
+    // the static widget's notifyHeight sends this same notification —
+    // mirrored here so open-spec hosts get sizing info from the baked
+    // per-pub_id variant too, not just the static iframe/img bundle.
+    // Hosts that don't implement it (ChatGPT) ignore unknown messages.
+    try {
+      window.parent.postMessage({
```
noahjax, line 223: Minor, low priority: `notify()` runs at script-exec time, before the `` has loaded, so `document.documentElement.offsetHeight` may be a pre-layout value. Unlike the static variant this one has no `load` listener to re-notify — only `resize`. The `` does carry explicit `width`/`height` attributes so layout is usually reserved, but with `width:100%;height:auto` in a narrower container the first measurement can still be off.

Inert today since claude.ai stays on the static URI, so this only matters if a host starts honoring per-call `resourceUri` overrides. Worth a `notify` on `img.load` for symmetry with the static variant.

noahjax (commented):
## Review

The engineering here is careful, the comments are unusually good, and the gate flip itself is a 2-line change with a clean rollback story. Three things concern me — details in the inline comments.

### 1. The `claude` bucket conflates renderers with non-renderers

`detectMcpClient` maps **any** UA containing `claude` or `anthropic` to `"claude"`. That is not just claude.ai and Claude Desktop — it also catches Claude Code's HTTP MCP client, the Agent SDK, and the Messages API `mcp_servers` connector. None of those are MCP Apps hosts, and after this change each gets widget `_meta` it cannot render **and** loses the PNG block: chart gone, text link only.

The PR body scopes this to "Claude Code (terminal CLI)... accepted for this PR", but the blast radius is the whole Anthropic-first-party non-Apps surface, and the fallback direction inverted: it used to be "PNG works nearly everywhere", now it is "widget works on ~2 of N Claude surfaces".

Related and unmentioned: the PNG content block was model-visible; `_meta.image_data_url` is not. On Claude the model can no longer see the chart while composing its answer — a property the surrounding comments explicitly cite as a benefit for the unknown-client path. Worth stating in the PR body as an accepted consequence.

### 2. `resourceDomains` probably does not do what the test plan assumes

This file already documents that claude.ai's outer CSP is `img-src 'self' blob: data:`, and the PR's core premise is that claude.ai **ignores** declared `frameDomains`. If the host ignores one `_meta.ui.csp` key, assuming it honors its sibling needs evidence. Combined with `MAX_INLINE_DATA_URL_BYTES = 250 KB` against the file's own "Tako chart PNGs run 50-300 KB", charts in the 250–300 KB band likely degrade to a text link on Claude.

### 3. The load-bearing widget JS has no executable test coverage

The widget is a template string CI never executes, so the source gate and the baked variant's size-changed notification are effectively untested. A jsdom test that `eval`s the widget script and drives `postMessage` would cover the source gate, the handshake, and the image branch in one go.

### Smaller notes

- **ChatGPT is not byte-identical.** The `csp` refactor adds `resourceDomains` to `_meta.ui.csp` for every client. The ChatGPT staging spot-check is the right gate, since a strict-validating Apps SDK rejecting an unknown `csp` key would break the working path. No test pins `resourceDomains` present for ChatGPT — the existing `toMatchObject` assertion passes either way.
- **`console.warn` URL logging is fine** — `image_url` is `{apiBase}/api/v1/image/{pub_id}/?dark_mode=...`, public and token-free.
- **`postMessage(..., "*")`** for size-changed matches the existing handshake pattern and leaks only width/height. Fine.
- **README frontmatter** (`description: >-`) is unrelated to the PR's subject. Harmless — YAML folded scalars join lines with spaces, so the single-line body round-trips identically.
- **Test comment quality is high**, and the `realPngResponse` helper correctly reflects that `fetchImageDataUrlAndDims` now parses IHDR — good catch that the old bare-signature fixture would have silently produced `undefined`.

### Verdict

The gate flip and the fetch-count invariant are well tested; the widget-side behavior is not, and the client bucketing is broader than the PR body claims. I would not merge on unit tests alone — the unchecked staging boxes are the actual gate.

Blocking, in order:
1. Determine the real UAs for Claude Code / Agent SDK / API MCP connector; keep them on the PNG path rather than deferring.
2. Staging verify on **both** claude.ai and Claude Desktop, including the >250 KB case.
3. Add executable coverage for the source gate (and the baked variant's size-changed) before this ships to a permanent URI.

noahjax (approved):
Approving — all three findings addressed on the merits.

- **Client bucketing**: you captured the real UA (`claude-code/2.1.220 (sdk-cli)`) rather than reasoning about it, and carved it out ahead of the broad match with unit coverage on observed strings. The `mcp_servers` residual is honestly documented and not resolvable without log evidence from that surface — right call to note it and move on.
- **Source gate**: `parent || top` with a warn on drop. A sibling's `event.source` is never parent or top, so the injection defense is intact.
- **Cap**: sidestepping the `resourceDomains` question by lifting to 400 KB is the better fix — the old cap's motivating failure was ChatGPT-side and `extraMeta` never ships the field there, so the reasoning holds.
- **Executable coverage**: the jsdom project is more than I asked for. Sibling forgery with the warn asserted, forged handshake, `javascript:` guard, and the 0px collapse are all real regression protection, and it is properly wired into `npm test` / `npm run typecheck` that CI runs.

Two things that remain unverifiable in code — inherent jsdom limits, not gaps in your work, but they are why the staging checks are still the gate:

1. The DOM tests dispatch synthetic `MessageEvent`s with an explicit `source` (jsdom does not populate it), so they pin the handler's logic, not real-browser cross-origin `WindowProxy` identity against claude.ai's actual delivery context.
2. jsdom never fires `load` for the `data:` URI, so the measure-then-`notifyHeight` sequence inside `image.addEventListener("load")` — the claude-ai-mcp#69 workaround, and the mechanism behind the 2026-05 clipping this PR fixes — is still unexercised. The tests confirm `src` is assigned, not that the height comes out right.

One note on the raised cap: it changes the top-end failure mode. Above 400 KB you still degrade to a link, but a ~533 KB-encoded `_meta` now tests claude.ai's message-size limit — and if that limit rejects the tool result, the failure is worse than a text link (whole result fails, not just the chart). Your blocking >250 KB staging check covers it; just worth knowing which way it can break.

Approving the code. Please land only after the three staging boxes are checked — especially Claude Desktop before the widget URI is burned in, given this file's own note that Desktop caches resource URIs beyond connector lifecycle.

PR #168, opened 2026-07-24
feat(oauth): RFC 8707 resource indicators + audience-bound tokens; per-tool securitySchemes

25eliu:
## Why

OpenAI's connector review couldn't complete the OAuth flow, and the server didn't implement RFC 8707 Resource Indicators or audience-bind its tokens. This makes the OAuth surface correct and cross-client (ChatGPT + Claude), independent of any one host's guidance.

## What changed

**Resource indicators + audience-bound tokens** (`feat(oauth)`)
- `/authorize` reads the `resource` param, rejects a value that doesn't canonicalize to this server's `${origin}/mcp` (`invalid_target`), and threads it through the state cookie → Stytch callback rebuild → consent form → auth code.
- `/token` stamps issued access + refresh tokens with `iss` (bare-origin issuer) and `aud` (the `/mcp` resource). Refresh copies the resource forward so audience survives rotation; an omitted resource defaults `aud` to `${origin}/mcp`.
- `/mcp` validation is now **tri-state** (`access.ts`): a Worker-issued access JWT that fails a binding check (type / `iss` / `aud` / scope / decrypt) returns a clean 401 instead of silently degrading to the raw-Tako-token path. `iss`/`aud` are enforced only when present, so pre-cutover tokens still work (access tokens live 15 min).
- `WWW-Authenticate` on `/mcp` 401s now carries `error`, `error_description`, `scope`, and `resource_metadata`; a dedicated challenge distinguishes a bad-binding OAuth token from a missing/malformed header.
- Protected-resource metadata keeps advertising `${origin}/mcp` as the resource (the `/mcp` URL is unchanged).

**Per-tool OAuth** — every runtime tool descriptor advertises `securitySchemes: [{ type: "oauth2", scopes: ["mcp"] }]` via `_meta` (the only descriptor field the SDK preserves); the widget `_meta` now merges rather than overwrites.

**Annotations** — `ToolAnnotations.openWorldHint` is now **required** so every tool declares all three hints explicitly. Values follow the MCP spec: web-search-style tools (`tako_search`/`answer`/`contents`/`available_data`) are `openWorldHint: true`; `tako_visualize` is `false` (closed domain — renders caller-supplied data).

**Already enforced, verified here:** PKCE S256 required (`handlers.ts:464`), exact redirect-URI matching (`:531`), 60 s single-use auth codes with `jti` replay cache (`:83`, `checkAndMarkRedeemed`).

## Tests (all green: 392 pass, typecheck clean, registry:check ok)
- e2e authorize→token: tokens carry correct `iss`/`aud`; `aud` defaults to `/mcp` when `resource` omitted; trailing-slash/query resource canonicalized; foreign resource → `invalid_target`; `resource` survives into the login state cookie.
- `/mcp` rejects wrong-audience, wrong-issuer, wrong-scope, wrong-type tokens; accepts legacy no-`iss`/`aud` during cutover.
- every tool in `tools/list` exposes `_meta.securitySchemes`.
- tool annotations pinned (open-world values + all-hints-explicit).

## NOT in this PR — reviewer login needs infra, not code

The OpenAI rejection's root cause is the sign-in step: Tako's `/login` offers only **Google SSO** and **email magic-link**, both of which need an out-of-band step (a Google account / an email inbox), so a reviewer can't complete it with static credentials. Fixing that means giving the reviewer a **username/password (no-MFA) path**, which requires enabling Stytch's Passwords product on the project + a login form and a `passwords/authenticate` handler + a provisioned reviewer account. That's a Stytch-config + product decision, not a code-only change, so it's deliberately left out rather than shipped as non-functional code. Recommend tracking it separately.

## Rollout
Access tokens are 15 min, so the no-`iss`/`aud` acceptance window self-clears almost immediately; refresh tokens (14 d) re-mint audienced access tokens on next use. No forced re-auth.

### DIFF
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -388,6 +388,16 @@ function registerTool(
     annotations: tool.annotations,
   };
 
+  // Advertise per-tool OAuth on every runtime descriptor. This only survives
+  // to the client via `_meta` — the SDK's `tools/list` serializer drops
+  // unknown top-level descriptor fields, but passes `_meta` through verbatim.
+  // Hosts (ChatGPT / Claude) read it to know the tool is reachable via the
+  // OAuth flow and requires the `mcp` scope. The widget block below MERGES
+  // into this rather than replacing it.
+  config._meta = {
+    securitySchemes: [{ type: "oauth2", scopes: ["mcp"] }],
+  };
+
   if (tool.outputSchema !== undefined) {
     // Output schemas are optional — only read tools declare them.
     // In practice every `outputSchema` we ship is `z.object(...)`,
@@ -601,6 +611,7 @@ function registerTool(
     // static URI for all three keys. The template resource is still
     // registered above for future hosts that may support it.
     config._meta = {
+      ...(config._meta as Record<string, unknown>),
       ui: { resourceUri: ui.uri },
       "ui/resourceUri": ui.uri,
       "openai/outputTemplate": ui.uri,
@@ -979,8 +990,15 @@ export async function handleMcpRequest(
   //   `tryResolveOAuthAccessToken` returns null, we forward the bearer
   //   verbatim. Backwards-compatible with every Claude Code install in
   //   the wild.
-  const oauthMappedToken = await tryResolveOAuthAccessToken(bearer, env);
-  const token = oauthMappedToken ?? bearer;
+  const origin = new URL(request.url).origin;
+  const oauth = await tryResolveOAuthAccessToken(bearer, env, origin, `${origin}/mcp`);
+  if (oauth.kind === "reject") {
+    // The bearer IS a Worker-issued JWT but failed a binding check
+    // (wrong audience/issuer/scope/type). Return a clean 401 rather than
+    // forwarding it to Django as a raw API key.
+    return oauthChallengeResponse(request, oauth.error, oauth.errorDescription);
+  }
+  const token = oauth.kind === "ok" ? oauth.takoToken : bearer;
 
   // Base ctx — `sendProgress` here is a placeholder overridden per
   // tool call inside `registerTool`'s SDK callback (where the
@@ -1086,9 +1104,30 @@ export async function handleMcpRequest(
  * doc — that is how MCP hosts (Claude.ai, ChatGPT) bootstrap an OAuth
  * flow when they have only the MCP URL and got a 401.
  */
+/**
+ * Build the RFC 6750 / RFC 9728 `WWW-Authenticate: Bearer` challenge value.
+ * Always carries `resource_metadata` (so an MCP host can bootstrap OAuth from
+ * a bare 401) and `scope`; `error` / `error_description` are added when known.
+ */
+function wwwAuthenticate(
+  origin: string,
+  error?: string,
+  errorDescription?: string,
+): string {
+  const parts = [
+    `resource_metadata="${origin}/.well-known/oauth-protected-resource"`,
+    `scope="mcp"`,
+  ];
+  if (error !== undefined) parts.unshift(`error="${error}"`);
+  if (errorDescription !== undefined) {
+    // Quote-safe: strip any double quotes from the human-readable reason.
+    parts.push(`error_description="${errorDescription.replace(/"/g, "'")}"`);
+  }
+  return `Bearer ${parts.join(", ")}`;
+}
+
 function bearerAuthResponse(request: Request, err: BearerAuthError): Response {
   const origin = new URL(request.url).origin;
-  const resourceMetadataUrl = `${origin}/.well-known/oauth-protected-resource`;
   return new Response(
     JSON.stringify({
       jsonrpc: "2.0",
@@ -1099,7 +1138,39 @@ function bearerAuthResponse(request: Request, err: BearerAuthError): Response {
       status: 401,
       headers: {
         "Content-Type": "application/json; charset=utf-8",
-        "WWW-Authenticate": `Bearer error="invalid_token", resource_metadata="${resourceMetadataUrl}"`,
+        "WWW-Authenticate": wwwAuthenticate(origin, "invalid_token"),
+      },
+    },
+  );
+}
+
+/**
+ * 401 for an OAuth access token that verified as ours but failed a binding
+ * check (audience / issuer / scope / type). Distinct from `bearerAuthResponse`
+ * (missing / malformed header) so the `error` / `error_description` reflect the
+ * specific OAuth failure and ChatGPT/Claude can decide whether to re-link.
+ */
+function oauthChallengeResponse(
+  request: Request,
+  error: string,
+  errorDescription: string,
+): Response {
+  const origin = new URL(request.url).origin;
+  return new Response(
+    JSON.stringify({
+      jsonrpc: "2.0",
+      id: null,
+      error: {
+        code: -32001,
+        message: errorDescription,
+        data: { kind: error },
+      },
+    }),
+    {
+      status: 401,
+      headers: {
+        "Content-Type": "application/json; charset=utf-8",
+        "WWW-Authenticate": wwwAuthenticate(origin, error, errorDescription),
       },
     },
   );
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -391,11 +391,13 @@ function registerTool(
   // Advertise per-tool OAuth on every runtime descriptor. This only survives
   // to the client via `_meta` — the SDK's `tools/list` serializer drops
   // unknown top-level descriptor fields, but passes `_meta` through verbatim.
-  // Hosts (ChatGPT / Claude) read it to know the tool is reachable via the
-  // OAuth flow and requires the `mcp` scope. The widget block below MERGES
-  // into this rather than replacing it.
+  // Reverse-DNS namespaced (`com.tako/…`) per the MCP `_meta` rules, which
+  // reserve unprefixed keys for the protocol and would collide with the field
+  // SEP-1488 (still Draft) will standardize at the descriptor top level. Hosts
+  // read it to know the tool is reachable via OAuth and needs the `mcp` scope.
+  // The widget block below MERGES into this rather than replacing it.
   config._meta = {
-    securitySchemes: [{ type: "oauth2", scopes: ["mcp"] }],
+    "com.tako/securitySchemes": [{ type: "oauth2", scopes: ["mcp"] }],
   };
 
   if (tool.outputSchema !== undefined) {
@@ -991,7 +993,7 @@ export async function handleMcpRequest(
   //   verbatim. Backwards-compatible with every Claude Code install in
   //   the wild.
   const origin = new URL(request.url).origin;
-  const oauth = await tryResolveOAuthAccessToken(bearer, env, origin, `${origin}/mcp`);
+  const oauth = await tryResolveOAuthAccessToken(bearer, env, origin);
   if (oauth.kind === "reject") {
     // The bearer IS a Worker-issued JWT but failed a binding check
     // (wrong audience/issuer/scope/type). Return a clean 401 rather than
@@ -1156,6 +1158,10 @@ function oauthChallengeResponse(
   errorDescription: string,
 ): Response {
   const origin = new URL(request.url).origin;
+  // RFC 6750 §3.1: insufficient_scope → 403 (the grant is wrong; a
+  // re-auth would re-present the same scope and loop). Everything else →
+  // 401 (re-authenticate).
+  const status = error === "insufficient_scope" ? 403 : 401;
   return new Response(
     JSON.stringify({
       jsonrpc: "2.0",
@@ -1167,7 +1173,7 @@ function oauthChallengeResponse(
       },
     }),
     {
-      status: 401,
+      status,
       headers: {
         "Content-Type": "application/json; charset=utf-8",
         "WWW-Authenticate": wwwAuthenticate(origin, error, errorDescription),
--- a/workers/src/oauth/access.ts
+++ b/workers/src/oauth/access.ts
@@ -16,21 +16,44 @@ import type { Env } from "../env.js";
 import { decryptAesGcm, verifyJwt } from "./jwt.js";
 import type { AccessTokenClaims } from "./types.js";
 
+/** The required scope every access token must carry to reach `/mcp`. */
+const REQUIRED_SCOPE = "mcp";
+
 /**
- * Verify an incoming bearer token as a Worker-issued OAuth access JWT.
- * Returns the decrypted Tako API token on success, `null` otherwise
- * (any failure path: missing config, wrong shape, bad signature,
- * expired, decryption failure).
+ * Result of resolving an incoming `/mcp` bearer:
+ * - `ok`: a valid Worker-issued OAuth access token → forward `takoToken`.
+ * - `reject`: the token IS a Worker-issued JWT but fails a binding check
+ *   (wrong type / issuer / audience / scope / undecryptable). The caller
+ *   MUST return a 401 — falling through to the raw-token path would send a
+ *   JWT to Django as an API key and mask the real reason with a Django 401.
+ * - `not_oauth`: not one of our access JWTs at all (raw Tako API token, a
+ *   non-JWT bearer, a bad signature, or an expired token) → the caller falls
+ *   through to the legacy raw-Tako-token path, unchanged.
+ */
+export type OAuthAccessResult =
+  | { kind: "ok"; takoToken: string }
+  | { kind: "reject"; error: string; errorDescription: string }
+  | { kind: "not_oauth" };
+
+/**
+ * Verify an incoming bearer as a Worker-issued OAuth access JWT and bind it to
+ * this resource server. `expectedIssuer` is the authorization-server issuer
+ * (the bare origin, e.g. `https://mcp.tako.com`); `expectedResource` is this
+ * server's canonical resource (the `/mcp` endpoint, e.g.
+ * `https://mcp.tako.com/mcp`). `iss` is checked against the former, `aud`
+ * against the latter.
  *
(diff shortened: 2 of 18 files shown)


workers/src/oauth/handlers.ts
```diff
@@ -1231,6 +1277,8 @@ export async function handleStytchCallback(
     authorizeUrl.searchParams.set("state", stateClaims.state);
   if (stateClaims.scope !== null)
     authorizeUrl.searchParams.set("scope", stateClaims.scope);
+  if (stateClaims.resource !== null)
```
robertabbott: **(1)** `StateCookieClaims.resource` is declared non-optional `string | null` (`types.ts:135`), so TS narrows this to `string` after the guard — but a state cookie minted by pre-PR code has no `resource` key, so at runtime it is `undefined`, and `undefined !== null` passes. I ran the exact sequence:

```
set("resource", undefined)  →  /authorize?resource=undefined
raw: "undefined"  →  canonicalizeResource  →  new URL throws  →  null  →  invalid_target
```

`handlers.ts:525` then returns that as a bare `text/plain` 400 (not `errorPage`), and this response already cleared the state cookie, so the user has no way to recover — they have to restart the connect flow from Claude.ai/ChatGPT. The exposure is every user who started login before the deploy and finishes the magic link / Google round-trip after it, i.e. up to `STATE_COOKIE_MAX_AGE_S = 10 * 60` (`cookies.ts:24`).

The PR's "No forced re-auth" rollout note holds for access and refresh tokens (both handled with `?? null`), but this is the one changed claim shape whose absence isn't. Could this be `?? null`, or `typeof stateClaims.resource === "string"` to match the defensive style already used in `access.ts:92,99`? Worth typing the field `resource?: string | null` too, so TS stops asserting a value old cookies don't have — and a test that mints a legacy state cookie and pushes it through `handleStytchCallback` would pin it.

workers/src/oauth/handlers.ts
```diff
@@ -472,6 +477,20 @@ function readAuthorizeQuery(url: URL): AuthorizeQuery | string {
...
+  // RFC 8707 resource indicator. This AS serves exactly one resource — this
+  // server's own origin — so a `resource` that canonicalizes to anything
+  // else is `invalid_target`. Omitted is allowed (older clients); the token
+  // is then audienced to this origin by default at issue time.
+  const expectedResource = `${url.origin}/mcp`;
+  const rawResource = p.get("resource");
+  let resource: string | null = null;
+  if (rawResource !== null) {
+    const canonical = canonicalizeResource(rawResource);
+    if (canonical === null || canonical !== expectedResource) {
```
robertabbott: **(2)** Exact equality against `${origin}/mcp` means `?resource=https://mcp.tako.com` is now a hard 400 — and so is `https://mcp.tako.com/`, since `resource.ts:36` strips the trailing slash down to the bare origin. Two things make that reachable rather than theoretical:

- The MCP authorization spec's *Canonical Server URI* section lists the bare origin (`https://mcp.example.com`) as a valid canonical URI and only **SHOULD**s the most specific form.
- RFC 9728 §3.3 requires the `resource` in the metadata doc to be identical to the resource identifier the well-known suffix was inserted into. This Worker serves the PRM doc **only** at the root suffix (`index.ts:82`, exact match — the §3.1 path-inserted `/.well-known/oauth-protected-resource/mcp` falls through to the 404 at `index.ts:120`) while advertising `resource: ${origin}/mcp` at `:234`. A client that derives its resource identifier from the URL it actually fetched therefore sends the bare origin, and now hard-400s where pre-PR it linked fine. The MCP TS SDK survives this only incidentally (`selectResourceURL` prefers `resourceMetadata.resource`, and `checkResourceAllowed` prefix-matches rather than checking §3.3 identity); a non-SDK client has no such cushion.

Given the PR exists to get through a connector review, was the bare-origin form considered and deliberately rejected as the stricter reading, or would you accept a resource whose canonical form is the origin (and/or also serve the path-inserted PRM URL)?

Separately on delivery: RFC 6749 §4.1.2.1 reserves the shown-to-the-user error for a missing/invalid redirect URI or client id, and requires other failures to be returned to the client via the redirect URI (`error`, `error_description`, `state`). Because this check lives inside `readAuthorizeQuery`, it runs before `clientFromClientId` (`:527`) and the `redirect_uris` membership check (`:531`) — so the handler structurally can't redirect an `invalid_target` even when it has a valid registered `redirect_uri`. Should the resource check move after client validation so the error reaches the connector as an OAuth error rather than an unstyled page?

workers/src/oauth/handlers.ts
```diff
@@ -952,6 +982,10 @@ async function handleRefreshGrant(
       user_email: claims.user_email,
       enc_tako_token: claims.enc_tako_token,
     },
+    issuer,
+    // Carry the audience forward across rotation. Legacy refresh tokens
+    // predate the claim (undefined) → default to the bare origin at issue.
+    claims.resource ?? null,
```
robertabbott: **(3)** `handleRefreshGrant` validates signature + `type === "refresh"` + `jti` replay, then copies `claims.resource` forward — but never checks the presented token's own `iss`/`aud` against this request's origin, while `issueTokens` re-stamps `iss` from the current Host (`:810`). Two consequences:

- **Dead-end loop:** a refresh token whose `resource` names hostname A, redeemed at hostname B, mints an access token with `iss` = B and `aud` = A`/mcp` — which `/mcp` at B then rejects on `aud` (`access.ts:99`). Refresh keeps returning 200 and every token it produces 401s, with no path back to a clean re-auth.
- **Audience laundering:** staging keeps `workers_dev: true` + `preview_urls: true` alongside the custom domain (`wrangler.jsonc:56-57`) on one signing key, so a token bound to the workers.dev hostname upgrades to `mcp.staging.tako.com` through a single refresh. Legacy refresh tokens (no `resource`, 14-day tail) get audienced to whatever Host they're presented on. Prod is single-origin (`:79`), so this part is staging-only.

Should the refresh grant reject a token whose `iss`/`aud` don't match this origin, rather than re-audiencing it to the arriving Host?

Also: `handleToken` never reads the `resource` form param — `params.get("resource")` appears only at `:485` in the `/authorize` path. The MCP auth spec has `resource` as a MUST on both authorization *and* token requests, and RFC 8707 §2 has the AS reject an unknown target. Today `resource=https://evil.example` at `/token` is silently ignored while the same value at `/authorize` hard-400s — worth validating symmetrically and cross-checking against the code's bound resource?

workers/src/oauth/access.ts
```diff
@@ -41,33 +64,80 @@ export async function tryResolveOAuthAccessToken(
...
+    };
+  }
+  if (typeof claims.iss === "string" && claims.iss !== expectedIssuer) {
+    return {
+      kind: "reject",
+      error: "invalid_token",
+      errorDescription: `token issuer ${claims.iss} does not match ${expectedIssuer}`,
+    };
+  }
+  if (typeof claims.aud === "string" && claims.aud !== expectedResource) {
```
robertabbott: **(4)** The MCP auth spec is a MUST here: servers "MUST reject tokens that do not include them in the audience claim or otherwise verify that they are the intended recipient." This branch keys the tolerance on claim *absence* rather than on a deploy date or a version claim, so nothing closes the window — it stays open indefinitely unless someone remembers to delete it. Since `handlers.ts:1009` sets `aud` unconditionally, being strict costs at most one 15-minute generation of access tokens (refresh re-mints audienced ones on next use), so the strict option looks genuinely cheap. Is there a follow-up to require `aud` once the cutover has passed?

Smaller point on the same guard: `typeof claims.aud === "string"` makes this broader than "absent" — RFC 7519 permits `aud` to be an array of strings, and an array skips the comparison entirely and is accepted. Unreachable while this Worker is the only minter, but should a non-string `aud` be rejected outright so a future minter can't silently bypass the check? Same question for `iss` at `:92`.

workers/src/oauth/access.ts
```diff
@@ -41,33 +64,80 @@ export async function tryResolveOAuthAccessToken(
...
-    return null;
+    return { kind: "not_oauth" };
   }
 
   const claims = await verifyJwt<AccessTokenClaims>(bearer, sign);
-  if (!claims || claims.type !== "access") return null;
-  if (typeof claims.enc_tako_token !== "string") return null;
+  // Bad signature / expired / malformed: indistinguishable from a raw token
+  // that happens to be JWT-shaped, so fall through rather than hard-reject.
+  if (!claims) return { kind: "not_oauth" };
```
robertabbott, line 95: **(5)** `verifyJwt` enforces `exp` itself (`jwt.ts:132-137`, 30s skew) and returns `null`, so an **expired** Worker-issued access token lands here as `not_oauth`. `mcp.ts:1001` then sets `token = bearer`, i.e. the expired JWT is forwarded to Django as `X-API-Key` (`django.ts:218`) — which is exactly the outcome the new tri-state docstring at `:25-28` says `reject` exists to prevent, and it's the single most frequent real-world case (every 15 minutes).

What the client actually gets: `initialize` and `tools/list` touch no Django endpoint, so the session looks healthy (200). The first `tools/call` gets a Django 401 → `DjangoUnauthorizedError` (`django.ts:364`) → caught at `mcp.ts:707` → `djangoErrorToToolResult` → `{ isError: true, … }` returned as a normal JSON-RPC result: **HTTP 200, no `WWW-Authenticate`**. Hosts trigger the refresh exchange off a 401 + challenge on the resource, so the model sees "tool failed: unauthorized" instead.

This isn't a regression — `origin/main` behaves the same — and a host honoring `expires_in` (`handlers.ts:1039`) refreshes proactively. But a signature-valid-but-expired token is trivially distinguishable from a raw API key (have `verifyJwt` report expiry separately, or parse `exp` before verifying). Was leaving expiry in the fall-through deliberate, or is it worth moving into `reject` with `error="invalid_token"` so `oauthChallengeResponse` fires and the host refreshes?

workers/src/oauth/access.ts
```diff
@@ -41,33 +64,80 @@ export async function tryResolveOAuthAccessToken(
...
+
+  // From here the signature is OURS, so any binding failure is a definite
+  // 401 rather than a fall-through.
+  if (claims.type !== "access") {
```
robertabbott, line 99: **(6)** Five of the six `reject` returns emit nothing (type here, `iss` `:92`, `aud` `:99`, scope `:106`, missing-credentials `:117`); only the decryption path logs (`:132`). `oauthChallengeResponse` (`mcp.ts:1153`) logs nothing either, and there's no Analytics Engine binding in `wrangler.jsonc` — observability is `console.*` into Workers Logs.

The repo convention is already there: `console.error` for anomalies (`:132`, `mcp.ts:1080`, `handlers.ts:667,1212`) and `console.log` for per-request facts (`mcp.ts:1037` logs the `?tools=` param). These new reject paths are the only anomaly paths that opt out, so if audience binding starts rejecting real users after deploy the only evidence is a response body that reaches ChatGPT/Claude and never us — nothing to filter on in `wrangler tail`, no counter, no separation from an ordinary Django 401. Could each reject log one `console.error` line with the mismatched `iss`/`aud` (both non-secret) so a rollout regression is greppable?

workers/src/mcp.ts
```diff
@@ -388,6 +388,16 @@ function registerTool(
...
   };
 
+  // Advertise per-tool OAuth on every runtime descriptor. This only survives
+  // to the client via `_meta` — the SDK's `tools/list` serializer drops
+  // unknown top-level descriptor fields, but passes `_meta` through verbatim.
+  // Hosts (ChatGPT / Claude) read it to know the tool is reachable via the
+  // OAuth flow and requires the `mcp` scope. The widget block below MERGES
+  // into this rather than replacing it.
+  config._meta = {
+    securitySchemes: [{ type: "oauth2", scopes: ["mcp"] }],
```
robertabbott: **(7)** Two questions about this placement.

**Location:** SEP-1488 ("securitySchemes in Tool Metadata for Mixed-Auth Servers") defines `securitySchemes` as a **top-level optional field on the tool descriptor**, not under `_meta` — and it's still Status: Draft. I checked: `securitySchemes` appears nowhere in the MCP `schema/draft/schema.ts` nor anywhere in `@modelcontextprotocol/sdk@1.29.0`'s `dist/`. So will ChatGPT's reviewer, reading the SEP-1488 field, actually find `_meta.securitySchemes`? Was this verified against a real ChatGPT read, or inferred from the SDK constraint?

**Namespace:** the MCP `_meta` rules reserve unprefixed key names for protocol-level metadata and **SHOULD** reverse-DNS anything else (`com.example/…`). An unprefixed `securitySchemes` sits in the namespace MCP reserves for itself, and collides by name with the field SEP-1488 would standardize elsewhere — worth `com.tako/securitySchemes` until the SEP lands?

For the record the comment's SDK claim checks out and was worth stating: `dist/esm/server/mcp.js:703` destructures only `{ title, description, inputSchema, outputSchema, annotations, _meta }`, and the `tools/list` handler emits only known keys plus `_meta` verbatim — unknown top-level fields are indeed dropped.

workers/src/mcp.ts
```diff
@@ -1099,7 +1138,39 @@ function bearerAuthResponse(request: Request, err: BearerAuthError): Response {
...
+      jsonrpc: "2.0",
+      id: null,
+      error: {
+        code: -32001,
+        message: errorDescription,
+        data: { kind: error },
+      },
+    }),
+    {
+      status: 401,
```
robertabbott: **(8)** This hardcodes 401 for every rejection kind, including `insufficient_scope`. RFC 6750 §3.1: "insufficient_scope: … The resource server SHOULD respond with the HTTP 403 (Forbidden) status code", and the MCP auth spec's error table agrees (`403 Forbidden — Invalid scopes or insufficient permissions`). A 401 tells the host "re-authenticate", which re-presents a token carrying the same wrong scope and loops; 403 tells it the grant itself is wrong. Should the status be derived from `oauth.error`?

(Not currently reachable — `readAuthorizeQuery` constrains scope to `SUPPORTED_SCOPES` and `:689` defaults it to `"mcp"` — so this is robustness rather than a live bug, but it's the kind of detail a connector reviewer checks.)

workers/src/tools/annotations.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    for (const tool of [
+      takoSearch,
+      takoAnswer,
+      takoContents,
+      takoAvailableData,
+      takoVisualize,
+    ]) {
+      expect(typeof tool.annotations.readOnlyHint).toBe("boolean");
+      expect(typeof tool.annotations.destructiveHint).toBe("boolean");
+      expect(typeof tool.annotations.openWorldHint).toBe("boolean");
```
robertabbott: **(9)** This block's stated guarantee is "every tool declares all three hints explicitly", but it iterates 5 of the 12 tools that declare `annotations` — missing `get_credit_balance` (`openWorldHint: false`), the three `tako_agent*`, and the three `tako_graph_*` (all `true`). The type already makes the field required, so the compiler covers explicitness; what this test could add over the type is coverage of the *values* across the full set. Would iterating the registry in `_registry.ts` rather than a hand-listed five make the assertion match the docstring — and catch a future tool that ships the wrong value?

workers/src/tools/types.ts
```diff
@@ -96,8 +96,13 @@ export interface ToolAnnotations {
...
   destructiveHint: boolean;
-  /** Tool's effects are observable outside Tako (e.g. creates a public URL). */
-  openWorldHint?: boolean;
+  /**
+   * Tool's effects are observable outside Tako (e.g. creates a public URL).
+   * Required: every tool must declare it explicitly so the read-only
+   * retrieval tools (false) are never conflated with the write/publish tools
+   * (true) in the advertised annotations.
+   */
+  openWorldHint: boolean;
```
robertabbott, line 108: **(10)** The retained doc comment now says the opposite of what the code ships. It defines the field as "Tool's effects are observable outside Tako (e.g. creates a public URL)" and the new lines assert "retrieval tools (false) … write/publish tools (true)" — but the shipped values are the reverse on every tool: the retrieval tools are `true`, and `tako_visualize`, which literally creates a public card URL, is `false`.

The values are right; the MCP schema defines `openWorldHint` as "this tool may interact with an 'open world' of external entities … if false, the tool's domain of interaction is closed" — a statement about the domain of interaction, not about externally observable effects. It's the comment that's stale. This looks worth fixing rather than leaving, because the branch history shows the comment already caused the error once: `7b12912` flipped four tools to `false` on the strength of exactly this wording before `1c2fa5f` reverted them. Replacing it with the spec's wording would stop the next flip.

workers/src/oauth/resource.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * This MCP server's canonical resource is its endpoint URL — `${origin}/mcp`
+ * (e.g. `https://mcp.tako.com/mcp`) — the value advertised as `resource` in the
+ * protected-resource metadata. Tokens issued by `/token` are audienced to that
+ * value (their `aud` claim), while `iss` is the authorization-server issuer
+ * (the bare origin, matching the RFC 8414 metadata). `/mcp` validates both.
+ */
+
+/** The authorization-server issuer identifier: the bare origin. */
+export function serverIssuer(req: Request): string {
```
robertabbott, line 15: **(11)** `serverIssuer` and `serverResource` are never imported — `handlers.ts:69` pulls only `canonicalizeResource`, and both call sites re-derive the origin inline instead (`handlers.ts:810` for the issuer, `mcp.ts:993-994` for both). Since these two helpers exist precisely to centralize that derivation, is the intent to use them at those call sites? That inline duplication is the drift vector for the two findings above — the `/mcp` validator, the `/token` issuer, and the `/authorize` expected-resource all compute the same thing independently, so they can diverge silently. If they're not wanted, dropping them keeps `resource.ts` to the one function that's actually load-bearing.

robertabbott (commented):
Findings inline. 1–3 are the load-bearing ones (numbered in the comment bodies); 4–11 are non-blocking.

robertabbott (approved):

PR #170, opened 2026-07-24
fix: align ChatGPT tool annotations for app review

25eliu:
## Summary

- preserve the canonical MCP `openWorldHint` values for Claude and other MCP clients
- translate tool annotations only for ChatGPT descriptors using OpenAI Apps review semantics
- mark ChatGPT agent kickoff tools as write operations because they enqueue workflows
- add regression coverage for default and opt-in ChatGPT tool surfaces
- add `chatgpt-app-submission.json` with app metadata, five default tools, five positive tests, and three negative tests

## Why

The MCP protocol defines `openWorldHint` by domain of interaction and explicitly treats web search as open-world. OpenAI Apps review uses the hint as a public/external-state action label. Applying either interpretation globally makes the descriptors wrong for the other client family.

The adapter keeps the MCP-spec annotations as the source of truth and changes only the descriptors served to ChatGPT-class clients. This avoids degrading other MCP clients while making the production tool listing consistent with OpenAI's submission review expectations.

## Impact

ChatGPT sees retrieval tools as read-only, non-destructive, and closed-world, while `tako_visualize` is identified as a non-destructive write that creates publicly accessible chart URLs. Other clients and the generated MCP registry retain their existing annotations.

Optional agent and graph tools receive the same client-specific treatment if enabled, but the submission JSON covers only the five tools exposed by the default production MCP URL.

## Validation

- `npm run typecheck`
- `npm test` — 399 tests passed
- `npm run registry:check` — 12 tools synchronized
- submission JSON structure, subtitle length, tool count, and test-case counts validated with `jq`
- `git diff --check`

### DIFF
--- a/chatgpt-app-submission.json
+++ b/chatgpt-app-submission.json
@@ -87,8 +87,8 @@
       "expected_output_url": null
     },
     {
-      "description": "Retrieve the underlying text from a supplied web result URL.",
-      "user_prompt": "Use Tako to fetch the page text behind https://www.bls.gov/cpi/.",
+      "description": "Retrieve the underlying page text behind a web result URL as a post-search follow-up.",
+      "user_prompt": "I got this web result URL from a Tako search earlier: https://www.bls.gov/cpi/. Use Tako to fetch the full page text behind it.",
       "file_attachment_urls": null,
       "tools_triggered": "tako_contents",
       "expected_output": "Returns readable page text from the supplied URL with enough context to use in a follow-up answer.",
--- a/chatgpt-app-submission.json
+++ b/chatgpt-app-submission.json
@@ -1,0 +1,140 @@
+{
+  "$schema": "https://developers.openai.com/apps-sdk/schemas/chatgpt-app-submission.v1.json",
+  "schema_version": 1,
+  "app_info": {
+    "display_name": "Tako",
+    "subtitle": "Search live data and charts",
+    "description": "Tako helps users find current data and metrics from proprietary datasets and the live web, get citation-backed answers, retrieve underlying result content, check available data coverage, and create charts from structured data.",
+    "category": "PRODUCTIVITY"
+  },
+  "tools": {
+    "tako_search": {
+      "annotations": {
+        "readOnlyHint": true,
+        "openWorldHint": false,
+        "destructiveHint": false
+      },
+      "justifications": {
+        "read_only_justification": "Searches Tako's data index and the live web and returns matching cards and links without modifying records or external state.",
+        "open_world_justification": "Does not publish content or change publicly visible or third-party state.",
+        "destructive_justification": "Does not delete, overwrite, revoke access, or perform irreversible actions."
+      }
+    },
+    "tako_answer": {
+      "annotations": {
+        "readOnlyHint": true,
+        "openWorldHint": false,
+        "destructiveHint": false
+      },
+      "justifications": {
+        "read_only_justification": "Retrieves source material and returns a synthesized, citation-backed answer without changing stored or external data.",
+        "open_world_justification": "Does not publish the answer or modify publicly visible or third-party systems.",
+        "destructive_justification": "Does not delete, overwrite, revoke access, or perform irreversible actions."
+      }
+    },
+    "tako_contents": {
+      "annotations": {
+        "readOnlyHint": true,
+        "openWorldHint": false,
+        "destructiveHint": false
+      },
+      "justifications": {
+        "read_only_justification": "Retrieves CSV data or page text from a supplied result URL without modifying the source.",
+        "open_world_justification": "Does not publish content or change the supplied website or any third-party system.",
+        "destructive_justification": "Does not delete, overwrite, revoke access, or perform irreversible actions."
+      }
+    },
+    "tako_available_data": {
+      "annotations": {
+        "readOnlyHint": true,
+        "openWorldHint": false,
+        "destructiveHint": false
+      },
+      "justifications": {
+        "read_only_justification": "Looks up Tako's structured-data coverage for an entity or metric and returns matching names without changing data.",
+        "open_world_justification": "Does not publish content or modify publicly visible or third-party state.",
+        "destructive_justification": "Does not delete, overwrite, revoke access, or perform irreversible actions."
+      }
+    },
+    "tako_visualize": {
+      "annotations": {
+        "readOnlyHint": false,
+        "openWorldHint": true,
+        "destructiveHint": false
+      },
+      "justifications": {
+        "read_only_justification": "Creates a persistent Tako chart card from structured data supplied by the user.",
+        "open_world_justification": "Creates a chart card with publicly accessible webpage and embed URLs.",
+        "destructive_justification": "Creates a new card but does not delete, overwrite, revoke access, or perform irreversible transactions."
+      }
+    }
+  },
+  "test_cases": [
+    {
+      "description": "Retrieve a current metric as structured data with a chart.",
+      "user_prompt": "Show me the latest US consumer price index and its recent trend as a chart.",
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_search",
+      "expected_output": "Returns a relevant CPI data card with recent values, source attribution, and an inline chart when available.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Answer a specific data question with supporting citations.",
+      "user_prompt": "What was US real GDP growth in 2025? Give me a concise answer with citations.",
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_answer",
+      "expected_output": "Returns one concise, citation-backed answer grounded in relevant data or web sources.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Retrieve the underlying text from a supplied web result URL.",
+      "user_prompt": "Use Tako to fetch the page text behind https://www.bls.gov/cpi/.",
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_contents",
+      "expected_output": "Returns readable page text from the supplied URL with enough context to use in a follow-up answer.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Check which structured metrics are available before running a priced search.",
+      "user_prompt": "What proprietary structured data does Tako have for Tesla?",
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_available_data",
+      "expected_output": "Returns whether coverage exists, matching entities, and the exact available metric names for a follow-up search.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Create a chart from data supplied directly by the user.",
+      "user_prompt": "Create a bar chart titled Regional Sales using North America: 500, Europe: 300, and Asia: 450, with units of USD millions.",
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_visualize",
+      "expected_output": "Creates and displays a regional sales bar chart and provides a link to open the resulting Tako card.",
+      "expected_output_url": null
+    }
+  ],
+  "negative_test_cases": [
+    {
+      "description": "Do not trigger for unrelated creative-writing requests.",
+      "user_prompt": "Write a short bedtime story about a friendly lighthouse.",
+      "file_attachment_urls": null,
+      "tools_triggered": null,
+      "expected_output": "The app should not be invoked because the request does not involve data discovery, data retrieval, or visualization.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Do not trigger for calendar-management requests.",
+      "user_prompt": "Schedule a meeting with my design team tomorrow at 10 AM.",
+      "file_attachment_urls": null,
+      "tools_triggered": null,
+      "expected_output": "The app should not be invoked because it cannot access or modify calendars.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Do not trigger for sending messages through another service.",
+      "user_prompt": "Email this market summary to my team.",
+      "file_attachment_urls": null,
+      "tools_triggered": null,
+      "expected_output": "The app should not be invoked because it cannot send email or messages.",
+      "expected_output_url": null
+    }
+  ]
+}
--- a/chatgpt-app-submission.json
+++ b/chatgpt-app-submission.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://developers.openai.com/apps-sdk/schemas/chatgpt-app-submission.v1.json",
+  "$schema": "https://developers.openai.com/plugins/schemas/chatgpt-app-submission.v1.json",
   "schema_version": 1,
   "app_info": {
     "display_name": "Tako",
--- a/workers/scripts/gen-registry.ts
+++ b/workers/scripts/gen-registry.ts
@@ -22,6 +22,10 @@ import { fileURLToPath, pathToFileURL } from "node:url";
 
 import { z } from "zod";
 
+import {
+  isToolOnSurface,
+  toolAnnotationsForClient,
+} from "../src/tools/_surface.js";
 import type { ToolAnnotations, ToolModule } from "../src/tools/types.js";
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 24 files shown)


chatgpt-app-submission.json
```diff
@@ -0,0 +1,140 @@
+{
+  "$schema": "https://developers.openai.com/plugins/schemas/chatgpt-app-submission.v1.json",
+  "schema_version": 1,
+  "app_info": {
+    "display_name": "Tako",
+    "subtitle": "Search live data and charts",
+    "description": "Tako helps users find current data and metrics from proprietary datasets and the live web, get citation-backed answers, retrieve underlying result content, check available data coverage, and create charts from structured data.",
+    "category": "PRODUCTIVITY"
+  },
+  "tools": {
```
robertabbott, line 10: The five annotation triples here match what `toolAnnotationsForClient` emits for `chatgpt` today — I checked all five. Nothing keeps them matching: `registry/server.json` is guarded by `registry:check` in `predeploy:production`, but this file is hand-maintained and referenced nowhere in the repo (no script, no doc, no test). A later edit to `CHATGPT_ANNOTATION_OVERRIDES` or to a tool's canonical annotations would leave the submitted app metadata claiming something production no longer serves. Should this be generated (or `--check`ed) from `TOOL_REGISTRY` + `toolAnnotationsForClient`, the way the registry already is?

workers/src/mcp.ts
```diff
@@ -133,6 +138,53 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
+  tako_agent_start: { readOnlyHint: false, openWorldHint: false },
+  tako_agent_wait: { openWorldHint: false },
+  tako_visualize: { openWorldHint: true },
+};
+
+export function toolAnnotationsForClient(
+  tool: AnyToolModule,
+  client: McpClientKind,
+): ToolAnnotations {
+  if (client !== "chatgpt") return tool.annotations;
```
robertabbott: `detectMcpClient` is a substring match on `chatgpt` / `openai`, and anything that doesn't match falls through to the canonical annotations. `index.test.ts:250` pins exactly that: a request with no UA gets `openWorldHint: true` on every retrieval tool. So if OpenAI's review tooling or directory crawler lists tools without a matching UA, the reviewer sees descriptors that contradict `chatgpt-app-submission.json` — the specific failure this PR exists to prevent. Has the real ChatGPT connector UA been confirmed from a production or staging request log, rather than the `"ChatGPT/1.0"` stand-in the tests use?

workers/src/mcp.ts
```diff
@@ -133,6 +138,53 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
+  >
+> = {
+  tako_search: { openWorldHint: false },
+  tako_answer: { openWorldHint: false },
+  tako_contents: { openWorldHint: false },
+  tako_available_data: { openWorldHint: false },
+  tako_graph_search: { openWorldHint: false },
+  tako_graph_related: { openWorldHint: false },
+  tako_graph_node: { openWorldHint: false },
+  tako_agent: { readOnlyHint: false, openWorldHint: false },
```
robertabbott: `tako_agent_start` becomes `readOnlyHint: false` because it enqueues a server-side run, while `tako_search` / `tako_answer` / `tako_contents` stay read-only — each of which debits a metered credit balance, and `tako_contents` also mints a presigned artifact. Under the Apps review reading, what's the rule that puts "enqueues a run" on the write side and "debits credits" on the read side? Worth stating next to the map so the next tool lands on the right side without re-deriving it.

workers/src/mcp.ts
```diff
@@ -133,6 +138,53 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
   return "unknown";
 }
 
+/**
+ * OpenAI's ChatGPT Apps review guidance uses `openWorldHint` as a
+ * public/external-state mutation label, which differs from the MCP protocol's
+ * domain-of-interaction meaning (where web search is the canonical open-world
+ * example). Keep each tool module's canonical MCP annotations untouched, then
```
robertabbott: The doc block frames every entry as an MCP-vs-Apps semantic difference, which is plainly right for `openWorldHint`. But `readOnlyHint` means the same thing in both — the tool does not modify its environment. If dispatching an agent run counts as a write, then `tako_agent.ts:322` and `tako_agent_start.ts:51` declaring `readOnlyHint: true` are wrong for plain MCP clients too, not just for ChatGPT. Should the canonical annotation be corrected rather than overridden per-client?

workers/src/mcp.ts
```diff
@@ -133,6 +138,53 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
+  tako_answer: { openWorldHint: false },
+  tako_contents: { openWorldHint: false },
+  tako_available_data: { openWorldHint: false },
+  tako_graph_search: { openWorldHint: false },
+  tako_graph_related: { openWorldHint: false },
+  tako_graph_node: { openWorldHint: false },
+  tako_agent: { readOnlyHint: false, openWorldHint: false },
+  tako_agent_start: { readOnlyHint: false, openWorldHint: false },
+  tako_agent_wait: { openWorldHint: false },
+  tako_visualize: { openWorldHint: true },
```
robertabbott: Every other entry narrows a hint for Apps review; this one flips `openWorldHint` to `true`. Under the Apps semantics the doc block cites, `openWorldHint: true` + `readOnlyHint: false` is the label ChatGPT reads to decide a call needs explicit user confirmation — and `tako_visualize` is default-on for ChatGPT (`CHATGPT_DEFAULT_ON_TOOL_NAMES`, mcp.ts:320), not opt-in. Was the confirmation-prompt effect on that flow checked, and should the rationale sit in the doc block next to the `tako_agent_start` note?

workers/src/mcp.ts
```diff
@@ -133,6 +138,53 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
+ * domain-of-interaction meaning (where web search is the canonical open-world
+ * example). Keep each tool module's canonical MCP annotations untouched, then
+ * adapt only the descriptors served to ChatGPT-class clients.
+ *
+ * `tako_agent_start` is also non-read-only under the Apps review meaning
+ * because it enqueues a server-side workflow. The combined `tako_agent` gets
+ * the same override for completeness even though ChatGPT receives the split
+ * start/wait pair instead.
+ */
+const CHATGPT_ANNOTATION_OVERRIDES: Readonly<
```
robertabbott: `ToolModule.descriptionByClient?: Partial<Record<McpClientKind, string>>` (`tools/types.ts:254`) is the established pattern for "this tool's descriptor differs on this host", and `registerTool` already resolves it two lines above the new call (`mcp.ts:435`). This adds a second, differently-shaped mechanism: a name-keyed map in `mcp.ts`, so a tool's ChatGPT annotations live in a different file from the tool. Would an `annotationsByClient?: Partial<Record<McpClientKind, Partial<ToolAnnotations>>>` on the module fit better — same resolution shape as descriptions, keys checked by the compiler instead of by a loop test, and each tool's two annotation sets sitting side by side?

chatgpt-app-submission.json
```diff
@@ -0,0 +1,140 @@
...
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_answer",
+      "expected_output": "Returns one concise, citation-backed answer grounded in relevant data or web sources.",
+      "expected_output_url": null
+    },
+    {
+      "description": "Retrieve the underlying text from a supplied web result URL.",
+      "user_prompt": "Use Tako to fetch the page text behind https://www.bls.gov/cpi/.",
+      "file_attachment_urls": null,
+      "tools_triggered": "tako_contents",
```
robertabbott, line 93: `tako_contents`'s description opens with "Fetch the real content behind one result URL from tako_search or tako_answer" and frames web URLs as the fallback "when tako_search / tako_answer surfaced relevant `web_results`" (`tako_contents.ts:29`, `:35`). This test case hands the model a pasted bls.gov URL with no prior search. The handler accepts any URL, but the description points the model at a search first, so ChatGPT may well call `tako_search` here and fail the declared `tools_triggered`. Should the prompt be reframed as a post-search follow-up, or the description loosened for the direct-URL case?

workers/src/mcp.ts
```diff
@@ -133,6 +138,53 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
+    string,
+    Partial<
+      Pick<
+        ToolAnnotations,
+        "readOnlyHint" | "destructiveHint" | "openWorldHint"
+      >
+    >
+  >
+> = {
+  tako_search: { openWorldHint: false },
```
robertabbott: Nit: `get_credit_balance` is the one registry tool with no entry here. Its canonical annotations already match the Apps reading (read-only, closed-world), so the behavior is right and `mcp.test.ts` passes. But in an otherwise-exhaustive 11-of-12 map the omission reads like an oversight rather than a decision. Worth an explicit entry or a one-line note?

workers/scripts/gen-registry.ts
```diff
@@ -128,6 +206,7 @@ const METADATA_PATH = resolve(REPO_ROOT, "registry", "metadata.json");
 const REGISTRY_PATH = resolve(REPO_ROOT, "registry", "server.json");
 const BARREL_PATH = resolve(TOOLS_DIR, "_registry.ts");
 const LLMS_FULL_PATH = resolve(REPO_ROOT, "llms-full.txt");
+const SUBMISSION_PATH = resolve(REPO_ROOT, "chatgpt-app-submission.json");
```
robertabbott, line 230: The guard is the right shape, but it never runs on a PR that edits only the file it protects. `chatgpt-app-submission.json` is at the repo root, and no workflow's paths filter reaches it: `workers-ci.yml` is `paths: ['workers/**', 'registry/**']`, `workers-deploy` is `workers/**`, `publish-mcp` is `server.json`, `regenerate` is `openapi/sdk.yaml`. So a PR whose only change is editing a hint value in the submission JSON gets `pr-title-lint` and nothing else — no `registry:check`, no drift error, and (per the note in `regenerate.yml`) no required status check on `main` to backstop it. That's the same failure this guard exists to prevent, just approached from the submission side rather than the tool side. Should the root file be added to the `workers-ci` paths filter, or moved under `workers/` so it's covered by the existing globs?

workers/scripts/gen-registry.ts
```diff
@@ -116,6 +120,80 @@ export function assertLlmsFullCoverage(
...
+ * descriptors. The submission file is hand-maintained (its justifications
+ * and test cases cannot be generated), so this validates instead of
+ * emitting: the declared tool set must equal ChatGPT's default tool
+ * surface, and each tool's annotation hints must equal what
+ * `toolAnnotationsForClient(tool, "chatgpt")` actually serves. Without
+ * this, an edit to a tool's annotations (canonical or `annotationsByClient`)
+ * would leave the submitted app metadata claiming something production no
+ * longer serves.
+ */
+export function assertChatgptSubmissionParity(
```
robertabbott, line 140: The two sibling guards in this file both have negative-path unit tests in `gen-registry.test.ts` — `assertAllToolsDescribed` plus six cases for `assertLlmsFullCoverage`. This one has none. I confirmed the hint-mismatch branch works by mutating `tako_search`'s ChatGPT override and watching it fail with the right message, but the `missing tool` and `extra tool` branches and the missing-`tools`-key throw are unexercised, and a guard whose only proof is that it passes on today's inputs can't distinguish working from vacuous. Worth three cases next to the existing ones?

workers/scripts/gen-registry.ts
```diff
@@ -116,6 +120,80 @@ export function assertLlmsFullCoverage(
...
+  if (submission.tools === undefined || typeof submission.tools !== "object") {
+    throw new Error(
+      'chatgpt-app-submission.json: missing top-level "tools" object',
+    );
+  }
+  const declaredTools = submission.tools;
+
+  // The submission covers the DEFAULT production MCP URL: no `?tools=`
+  // opt-ins, client detected as chatgpt.
+  const noOptIns: ReadonlySet<string> = new Set();
```
robertabbott, line 159: `isToolOnSurface` takes no tier, and the `tier === "free"` gate stayed in `mcp.ts` immediately above the call — so it's the one surface gate that did not move into `_surface.ts`, and this check is implicitly asserting the authenticated surface only. `FREE_TIER_TOOL_NAMES` is `{tako_available_data, tako_search, tako_answer}`, so an anonymous caller gets 3 of the 5 tools declared here, with `tako_contents` and `tako_visualize` — the tools behind test cases 3 and 5 — absent entirely. Should the check take a tier and assert the free surface too, or at minimum narrow the docstring from "the default production MCP URL" to "the authenticated default surface", so the assumption is visible to whoever reads a failure?

workers/src/tools/types.ts
```diff
@@ -265,6 +265,39 @@ export interface ToolModule<
...
+   *   Apps review, and `tako_visualize` (mints public chart URLs) is the
+   *   reverse.
+   * - `readOnlyHint` — nominally the same in both ("does not modify its
+   *   environment"), but the two ecosystems draw the write line
+   *   differently. The rule used here: a call is a WRITE when it creates
+   *   a durable, user-addressable resource — an agent run reachable later
+   *   via `run_id`/`thread_id`, a chart card with public URLs. Metering
+   *   side effects do NOT flip the hint: every Tako tool, including pure
+   *   reads, debits a credit balance, so counting billing as a write
+   *   would make `readOnlyHint` vacuously false everywhere. Under the MCP
```
robertabbott: The rule stated two sentences up is "a call is a WRITE when it creates a durable, user-addressable resource — an agent run reachable later via `run_id`/`thread_id`". This sentence then calls that same object "ephemeral compute whose record is bookkeeping" in order to keep the canonical `readOnlyHint: true`. Both descriptions are of one thing, and `readOnlyHint` has a single definition in MCP — `tako_agent_start` returns a `run_id` the caller is required to address later, and it accepts `thread_id` to append to a thread that persists. So the doc now records the disagreement rather than resolving it. Is this genuinely an MCP-vs-Apps semantic split, or is canonical `readOnlyHint: true` on `tako_agent` / `tako_agent_start` simply wrong for Claude as well?

workers/src/mcp.ts
```diff
@@ -140,11 +145,34 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
   if (ua.includes("claude") || ua.includes("anthropic")) return "claude";
   // ChatGPT's Apps SDK connector typically advertises `ChatGPT-User`,
-  // `openai-mcp`, or similar in UA.
-  if (ua.includes("chatgpt") || ua.includes("openai")) return "chatgpt";
+  // `openai-mcp`, or similar in UA. OpenAI's published crawler/agent UAs
+  // (`GPTBot`, `OAI-SearchBot`) contain neither substring, so match them
+  // explicitly: if any OpenAI-family tooling (directory crawler, app
+  // review harness) lists our tools, it must see the same descriptors
+  // `chatgpt-app-submission.json` declares — not the canonical MCP ones
+  // the unknown-UA fallback would serve. Residual risk: an OpenAI
```
robertabbott: Thanks for writing the residual risk down plainly — but the default still falls in the direction that costs the most. The two outcomes aren't symmetric: an OpenAI reviewer or crawler with an unmatched UA lands on `unknown`, sees canonical labels, and contradicts the submission (an app rejection), whereas serving the narrower Apps labels to a generic third-party MCP client understates `openWorldHint` on retrieval — cosmetic, and reversible. Would flipping the polarity close the hole: Apps-review labels for `unknown`, canonical MCP retained for `claude` (the one client family actually known to read the spec meaning)?

Separately, on the two added UAs: `gptbot` / `oai-searchbot` are OpenAI's web crawlers, not the Apps connector, and `McpClientKind` gates more than annotations — it selects the tool set (`isToolOnSurface`), decides whether the chart widget is registered (`mcp.ts:497`), and suppresses image content blocks (`tako_search.ts:229`, `tako_visualize.ts:232`). The rationale here is annotation-only, so the lever is wider than the stated reason.

workers/src/mcp.ts
```diff
@@ -140,11 +145,34 @@ export function detectMcpClient(userAgent: string | null): McpClientKind {
...
+  }
   return "unknown";
 }
 
+// Per-client annotation resolution lives with the tool surface config in
+// `tools/_surface.ts` (each tool declares its own `annotationsByClient`
+// next to its canonical MCP annotations — see `annotationsByClient` in
+// `tools/types.ts` for the MCP-vs-Apps-review semantics). Re-exported here
+// so existing imports from `./mcp.js` keep working.
+export { toolAnnotationsForClient };
```
robertabbott: Nit: "so existing imports from `./mcp.js` keep working" isn't accurate — `toolAnnotationsForClient` doesn't exist on `main`, so there are no pre-existing importers. `gen-registry.ts` imports it from `_surface.js` directly, and the only consumer of the `mcp.js` path is `mcp.test.ts`, added in this PR. The stated reason would also discourage a future reader from deleting it. Point the test at `_surface.js` and drop the re-export?

robertabbott (commented):
Re-review of the revisions. The four substantive prior comments got real fixes: `annotationsByClient` on the modules, surface config extracted to `_surface.ts` and shared with codegen, the write-line rule written down in `types.ts`, `get_credit_balance`'s omission made explicit, and the `tako_contents` test case reframed as a post-search follow-up.

Verified locally on `2595e59`: `registry:check` passes (12 tools); the new guard fires correctly when mutated (flipping `tako_search`'s ChatGPT `openWorldHint` yields a specific, actionable error); full suite is 462 passing; and `chatgpt-app-submission.json` validates against the real OpenAI-hosted schema at its `$schema` URL with 0 errors.

Six things below. Two are about the drift guard itself.

workers/scripts/gen-registry.ts
25eliu: Fixed in 7a0477f — added `chatgpt-app-submission.json` to the `workers-ci` paths filter (push and pull_request), rather than moving it: root placement matches the `server.json` precedent and the path is what the OpenAI submission references. Also added `llms-full.txt`, which had the identical gap — `assertLlmsFullCoverage` protects a root file no workflow triggered on. A comment above the filter now records why those two root files are listed.

workers/scripts/gen-registry.ts
25eliu: Added in 7a0477f — `gen-registry.test.ts` now has a `assertChatgptSubmissionParity` describe block next to the existing guards: missing tool, extra tool, missing top-level `tools` key, plus the hint-mismatch case (cheap to pin even though you verified it by mutation) and a case for the new free-tier-subset branch. The real-file parity assertion stays in `registry:check` — the vitest suite runs in the Workers pool, which has no filesystem, and a note in the test file says so.

workers/scripts/gen-registry.ts
25eliu: Both halves, in 7a0477f. `isToolOnSurface` now takes a `tier` (default `"authenticated"`) and the free-tier gate moved out of `mcp.ts` into `_surface.ts` as gate 1 of 4 — the registration loop passes the request tier, so the one stray surface gate is gone. The parity check passes `"authenticated"` explicitly, its docstring now says AUTHENTICATED surface and names the expected anonymous gap (`tako_contents`/`tako_visualize` absent, test cases assume an OAuth-linked connection), and a new assertion checks the free surface is a SUBSET of the declared tools — so a grown free tier can't silently serve an undeclared tool.

One thing this surfaces that code can't resolve: whether the OpenAI reviewer actually connects authenticated depends on the auth config in the ChatGPT developer portal, not this repo. If the app is submitted without OAuth linking, the reviewer lands on the anonymous 3-tool surface and declared test cases 3 and 5 fail. Flagged to Eric to confirm the portal setup.

workers/src/tools/types.ts
25eliu: You're right — it was one object described two ways, and the second description existed only to defend the canonical value. Resolved at the source in 7a0477f: `tako_agent` and `tako_agent_start` now declare canonical `readOnlyHint: false` (a queued run reachable later via `run_id`/`thread_id` is exactly the durable, user-addressable resource the rule names), their `chatgpt` overrides shrink to `openWorldHint` only — the one genuine MCP-vs-Apps split — and the `types.ts` doc now states the write line once, for every client, with no per-ecosystem carve-out. `registry/server.json` regenerated; `annotations.test.ts` pins the writer set as exactly {`tako_visualize`, `tako_agent`, `tako_agent_start`}. Cost: claude-family hosts may now confirm before agent dispatch — which is correct labeling for a tool that spends non-trivial credits.

workers/src/mcp.ts
25eliu: Polarity flipped in 7a0477f: `annotationClientFamily` in `_surface.ts` reserves canonical MCP labels for `claude` and resolves everything else — `unknown` included — to the `chatgpt` override family, for exactly the asymmetry you describe (unmatched-UA reviewer contradiction = rejection risk; understated `openWorldHint` for a generic client = cosmetic). After the canonical `readOnlyHint` fix on the agent tools, the per-client delta is `openWorldHint` only, which shrinks the blast radius of the flip. Tests updated end-to-end: no-UA `tools/list` now pins closed-world retrieval, and a new claude-UA test pins the canonical path.

On the crawler UAs: kept at the full `chatgpt` classification, but the comment now states the wider rationale instead of the annotation-only one. The polarity flip alone can't make a crawler's TOOL LIST match the submission — the `unknown` surface swaps the split agent pair for `tako_agent` and drops `tako_visualize` — so OpenAI-family tooling needs the chatgpt surface, not just chatgpt labels. The extra levers are harmless to a crawler (widget metadata is inert; image suppression trims bytes).

workers/src/mcp.ts
25eliu: Fixed — re-export dropped, `mcp.test.ts` imports `toolAnnotationsForClient` from `tools/_surface.js`, and the comment no longer claims pre-existing importers. (`mcp.ts` keeps its own import for the `registerTool` call.)

robertabbott (approved):

PR #171, opened 2026-07-26
feat: anonymous rate-limited free tier for /mcp

25eliu:
## Summary

Serves unauthenticated `/mcp` requests (no `Authorization` header) as an **anonymous free tier** instead of a 401 — the no-OAuth fallback for hosts that can't run a sign-in flow (e.g. the upcoming Centaur integration), and a zero-friction first-run path for any MCP client.

- **Fail-closed activation:** requires BOTH a `FREE_TIER_API_KEY` secret (a dedicated free-tier Tako account's key, forwarded to Django as `X-API-Key`) and the `FREE_TIER_RATE_LIMITER` binding. With either missing, behavior is byte-identical to today (401). Each env opts in independently; nothing is enabled until the secret is set.
- **Rate limiting:** 10 `tools/call`/min per client IP via Cloudflare's rate-limit binding (`unsafe.bindings`, per-env namespaces). Handshake methods (`initialize`, `tools/list`, notifications) are never metered. Over-limit → HTTP 429 with a JSON-RPC error carrying an upsell message. Limiter runtime failure fails **open** (Django-side account limits are the spend backstop); missing config fails **closed**.
- **Restricted toolset:** anonymous connections see exactly 3 tools — `tako_available_data`, `tako_search`, `tako_answer` — hidden (not registered), not erroring. The gate runs before every other registration rule, so `?tools=` opt-ins and ChatGPT default-on tools cannot widen the anonymous surface.
- **Auth semantics preserved:** a *malformed/empty* `Authorization` header still 401s (a client trying to authenticate is never silently downgraded); raw-token and OAuth JWT paths are unchanged.
- **Batch hardening (from adversarial review):** anonymous JSON-RPC batch arrays are rejected with 400 (`-32600`, `batch_not_supported`) — the SDK dispatches every batch element, so metering a batch as one hit would have let one request spend unbounded credits. Batching was removed from the MCP spec in 2025-06-18, so no modern client is affected.

New module: `workers/src/freetier.ts` (config gate, metering decision, limiter call, 429/400 responses). Wiring in `mcp.ts` (`handleMcpRequest` anonymous branch + `createMcpServer` `tier` option). Operator docs in `workers/README.md`, including a warning that enabling the free tier changes onboarding for OAuth-capable hosts (no 401 → no automatic sign-in prompt).

Design doc: `docs/superpowers/specs/2026-07-26-anonymous-free-tier-design.md` (local, gitignored).

## Test plan

- [x] 31 new tests (unit + end-to-end through `worker.fetch`): fail-closed 401, malformed-auth 401, 3-tool surface, `X-API-Key` forwarding, per-IP metering, 429 shape/message, batch rejection with zero Django spend, limiter fail-open, authenticated bypass
- [x] Full suite: 428/428 passing, `tsc --noEmit` clean
- [x] Adversarial whole-branch review (batch-bypass finding fixed + re-reviewed)
- [ ] Staging: set `wrangler secret put FREE_TIER_API_KEY --env staging` and verify anonymous connect from a real MCP client (3 tools listed, 11th call/min → 429)
- [ ] Confirm the free-tier Tako account's Django-side limits are set (spend backstop for fail-open)
- [ ] Decide production enablement (see README operator warning re: OAuth-host onboarding)

### DIFF
--- a/workers/README.md
+++ b/workers/README.md
@@ -45,16 +45,41 @@ Behavior when active:
 - Limiter runtime failures fail open (the shared account's Django-side
   limits are the spend backstop); missing configuration fails closed.
 
-Enabling on an environment:
+### Rollout checklist
 
-```bash
-# 1. Create/choose the free-tier Tako account, then:
-wrangler secret put FREE_TIER_API_KEY --env staging   # paste the API key
-wrangler secret put FREE_TIER_API_KEY --env production
-# 2. The FREE_TIER_RATE_LIMITER binding is already in wrangler.jsonc;
-#    deploy as usual. To change the limit, edit `simple.limit` there AND
-#    the FREE_TIER_LIMIT_MESSAGE copy in src/freetier.ts.
-```
+Setting the `FREE_TIER_API_KEY` secret is the on/off switch — until it is
+set on an environment, anonymous requests 401 exactly as before. There is
+nothing to configure in the Cloudflare dashboard: the rate limiter is
+config-as-code in `wrangler.jsonc` and deploys with the Worker.
+
+1. **Deploy the Worker** (normal deploy flow). The
+   `FREE_TIER_RATE_LIMITER` binding ships with it.
+2. **Create the dedicated free-tier Tako account** and generate its API
+   key from the account page. All anonymous traffic spends this one
+   account's credits.
+3. **Set Django-side limits/credits on that account.** This is the total
+   spend backstop: the Worker limiter fails *open* on limiter runtime
+   errors, so give the account a credit budget you can afford to lose.
+4. **Enable staging first:**
+   ```bash
+   wrangler secret put FREE_TIER_API_KEY --env staging   # paste the API key
+   ```
+5. **Verify on staging:** connect an MCP client to
+   `mcp.staging.tako.com/mcp` with no auth → `tools/list` shows exactly
+   the three free tools; make 11 `tools/call`s inside a minute → the
+   11th returns 429 with the upsell message.
+6. **Enable production** — after consciously accepting the operator
+   warning below:
+   ```bash
+   wrangler secret put FREE_TIER_API_KEY --env production
+   ```
+
+To change the per-IP limit later, edit `simple.limit` in **all three**
+`unsafe.bindings` blocks in `wrangler.jsonc` (dev/staging/production)
+AND the `FREE_TIER_LIMIT_MESSAGE` copy in `src/freetier.ts`, then
+redeploy. Counting is per-Cloudflare-colo and approximate (an IP whose
+traffic hits two colos can see roughly 2× the limit) — acceptable for
+abuse protection, not billing.
 
 **Operator warning:** OAuth-capable hosts (claude.ai and similar) decide
 whether to run their OAuth sign-in flow based on getting a 401 from the
--- a/workers/README.md
+++ b/workers/README.md
@@ -20,26 +20,47 @@ npm run dev
 
 `/mcp` requests with **no** `Authorization` header are served as a
 rate-limited free tier instead of a 401 — but only in environments that
-opt in. Two bindings gate it (fail-closed: with either missing, anonymous
+opt in. Three bindings gate it (fail-closed: with any missing, anonymous
 requests 401 exactly as before):
 
 | Binding | Kind | Purpose |
 |---|---|---|
-| `FREE_TIER_API_KEY` | secret | Tako API key of the dedicated free-tier account, forwarded to Django as `X-API-Key` |
-| `FREE_TIER_RATE_LIMITER` | `unsafe.bindings` ratelimit in `wrangler.jsonc` | 10 metered requests / 60 s per client IP |
+| `FREE_TIER_API_KEY` | secret | Tako API key of the dedicated free-tier account, forwarded to Django as `X-API-Key` (trimmed, so a piped `wrangler secret put` newline can't break it) |
+| `FREE_TIER_RATE_LIMITER` | `unsafe.bindings` ratelimit in `wrangler.jsonc` | per-IP fairness bucket: 10 free-tool `tools/call`s / 60 s |
+| `FREE_TIER_GLOBAL_RATE_LIMITER` | `unsafe.bindings` ratelimit in `wrangler.jsonc` | global ceiling: 1000 anonymous requests / 60 s across ALL callers |
+
+The global ceiling is the actual spend/volume bound: per-IP keying means
+little for hosted MCP hosts (claude.ai, ChatGPT, and similar egress from
+a handful of platform IPs), and it also caps the otherwise-unmetered
+handshake methods. The per-IP bucket is the fairness layer on top.
 
 Behavior when active:
 
 - Anonymous connections see exactly three tools: `tako_available_data`,
   `tako_search`, `tako_answer`. Everything else is hidden.
-- Only `tools/call` is metered (per `CF-Connecting-IP`); `initialize` /
-  `tools/list` are always free. Over-limit calls get an HTTP 429 with an
-  upsell message pointing at https://trytako.com/account/. Each metered
-  request logs one line: `[free-tier] ip=<ip> allowed|limited`.
+- Every anonymous request counts against the global ceiling; the per-IP
+  bucket counts only `tools/call`s naming one of the three free tools
+  (the only requests that spend Tako credits). A `tools/call` for a
+  hidden tool returns "tool not found" without burning per-IP quota;
+  `initialize` / `tools/list` never burn it. IPv4 clients are keyed by
+  address, IPv6 by /64 prefix.
+- An over-limit free-tool call returns **HTTP 200 with a JSON-RPC tool
+  result** (`isError: true`) carrying the upsell message pointing at
+  https://trytako.com/account/ — deliberately not a 429, which MCP SDK
+  clients surface as a transport error the model never reads. The global
+  ceiling (which trips before the body is parsed) returns a plain 429.
+- If the shared account itself runs out of Tako credits (Django 402),
+  the tool result carries the same style of upsell copy instead of the
+  raw billing error.
+- Each per-IP-metered request logs one line:
+  `[free-tier] ip=<key> allowed|limited`.
 - JSON-RPC batch requests (array bodies) are rejected outright with an
   HTTP 400 — never metered, never forwarded to Django. Batching was
   removed from the MCP spec in 2025-06-18, and metering a batch as a
   single limiter hit would let it stand in for unlimited `tools/call`s.
+- Anonymous request bodies are capped at 128 KiB (HTTP 413 past that) —
+  the body peek is new unauthenticated surface and its read is bounded,
+  so a lying `Content-Length` doesn't help.
 - A malformed `Authorization` header still 401s — only a fully absent
   header selects the free tier.
 - Limiter runtime failures fail open (the shared account's Django-side
@@ -49,11 +70,12 @@ Behavior when active:
 
 Setting the `FREE_TIER_API_KEY` secret is the on/off switch — until it is
 set on an environment, anonymous requests 401 exactly as before. There is
-nothing to configure in the Cloudflare dashboard: the rate limiter is
-config-as-code in `wrangler.jsonc` and deploys with the Worker.
+nothing to configure in the Cloudflare dashboard: both rate limiters are
+config-as-code in `wrangler.jsonc` and deploy with the Worker.
 
 1. **Deploy the Worker** (normal deploy flow). The
-   `FREE_TIER_RATE_LIMITER` binding ships with it.
+   `FREE_TIER_RATE_LIMITER` and `FREE_TIER_GLOBAL_RATE_LIMITER` bindings
+   ship with it.
 2. **Create the dedicated free-tier Tako account** and generate its API
    key from the account page. All anonymous traffic spends this one
    account's credits.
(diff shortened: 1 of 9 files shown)


workers/src/freetier.ts
```diff
@@ -0,0 +1,200 @@
...
+  "https://trytako.com/account/ for higher limits.";
+
+/**
+ * HTTP 429 for an over-limit anonymous request. JSON-RPC envelope with
+ * `id: null` (the limiter runs before the SDK parses the request id),
+ * `code: -32000` (implementation-defined server-error range; distinct
+ * from `-32001`, which auth failures use), and a `Retry-After` matching
+ * the limiter window.
+ */
+export function freeTierLimitResponse(): Response {
```
robertabbott: The upsell message is the whole point of this response, but I do not think it reaches the model. On a non-2xx POST the SDK client throws `StreamableHTTPError(429, "Error POSTing to endpoint: {…json…}")` at the transport layer (`streamableHttp.js:364`) — the pending `tools/call` rejects rather than resolving, and `id: null` means the response cannot be matched to the request anyway. So the carefully worded copy arrives, if at all, as a stringified JSON blob inside a transport error, not as something the model can read and relay.

Since `tools/call` is the only metered method, could this instead be HTTP 200 with a JSON-RPC *result* echoing the request `id`, `isError: true`, and the upsell as text content — the same shape `djangoErrorToToolResult` already produces? The tests here assert the Worker's response bytes, and the staging check ("11th call → 429") passes either way, so neither would catch this. Same question for `freeTierBatchResponse`.

workers/src/freetier.ts
```diff
@@ -0,0 +1,200 @@
...
+  let body: unknown;
+  try {
+    body = await request.clone().json();
+  } catch {
+    return "allowed";
+  }
+  if (Array.isArray(body)) return "batch";
+  if (!isMeteredJsonRpcBody(body)) return "allowed";
+
+  const key = request.headers.get("cf-connecting-ip") ?? "unknown";
```
robertabbott: Is `CF-Connecting-IP` the right key for the clients this is actually for? For a locally-run client (Claude Code, Claude Desktop) it is the user's IP and works as intended. But for a hosted host — claude.ai, ChatGPT, and I would guess Centaur — the MCP fetch happens server-side, so this is the platform's egress IP: the entire platform's anonymous traffic shares one 10/min bucket. That is simultaneously a self-DoS for legitimate users and zero per-user abuse protection.

And for direct IPv6 clients, one `/64` yields effectively unlimited distinct keys.

If the target is a hosted integration, would a second bucket on a constant key (a global ceiling, set well above expected legitimate volume) be the control that actually bounds spend, with per-IP kept as the fairness layer?

workers/README.md
```diff
@@ -15,3 +15,78 @@ npm test
...
+  `tools/list` are always free. Over-limit calls get an HTTP 429 with an
+  upsell message pointing at https://trytako.com/account/. Each metered
+  request logs one line: `[free-tier] ip=<ip> allowed|limited`.
+- JSON-RPC batch requests (array bodies) are rejected outright with an
+  HTTP 400 — never metered, never forwarded to Django. Batching was
+  removed from the MCP spec in 2025-06-18, and metering a batch as a
+  single limiter hit would let it stand in for unlimited `tools/call`s.
+- A malformed `Authorization` header still 401s — only a fully absent
+  header selects the free tier.
+- Limiter runtime failures fail open (the shared account's Django-side
```
robertabbott, line 76: Fail-open is a defensible call, but it makes the Django-side cap the only thing standing between a limiter outage and unbounded spend on the shared account — and the only signal that the Worker is failing open is a `console.error` in Workers Logs. No counter, no Logpush metric, no alarm.

How would you find out the free tier had been failing open for an hour? Given the shortcut, this feels like the one piece of observability worth having before the secret is set in production.

workers/README.md
```diff
@@ -15,3 +15,78 @@ npm test
...
+set on an environment, anonymous requests 401 exactly as before. There is
+nothing to configure in the Cloudflare dashboard: the rate limiter is
+config-as-code in `wrangler.jsonc` and deploys with the Worker.
+
+1. **Deploy the Worker** (normal deploy flow). The
+   `FREE_TIER_RATE_LIMITER` binding ships with it.
+2. **Create the dedicated free-tier Tako account** and generate its API
+   key from the account page. All anonymous traffic spends this one
+   account's credits.
+3. **Set Django-side limits/credits on that account.** This is the total
```
robertabbott: Rate-limit exhaustion gets a hand-written upsell, but the shared account running out of credits gets whatever Django's billing error says, spliced into model-visible text by `djangoErrorToToolResult`. That is the more likely steady-state failure of the two, and — since fail-open explicitly leans on this cap — the case where behavior matters most.

Is it worth mapping the credit-exhaustion status to the same upsell copy, so it does not surface to anonymous users as a raw Django billing message?

workers/src/env.ts
```diff
@@ -89,6 +103,22 @@ export interface Env {
...
   OPENAI_APPS_CHALLENGE_TOKEN?: string;
+  /**
+   * Tako API key of the dedicated free-tier account, forwarded to Django
+   * as `X-API-Key` for anonymous (no `Authorization` header) `/mcp`
+   * requests. A Worker secret (`wrangler secret put FREE_TIER_API_KEY
+   * --env <env>`). Optional and fail-closed: when unset (or when
+   * `FREE_TIER_RATE_LIMITER` is unbound), anonymous requests get the
+   * same 401 as before the free tier existed. See `freetier.ts`.
+   */
+  FREE_TIER_API_KEY?: string;
```
robertabbott, line 114: Not arguing for server-side scoping here — but since this is a full-privilege Tako API key and the three-tool restriction is Worker-side registration filtering rather than an authorization boundary, the key is the whole blast radius. Two things I could not find documented:

- The kill switch. `wrangler secret delete FREE_TIER_API_KEY --env production` reverts to the pre-PR 401 with no code change — is that worth stating in the README as the rollback, next to the on switch?
- Rotation. If the key needs rotating, is `wrangler secret put` over the top enough, or does the old key need explicit revocation on the Django side too?

workers/src/freetier.ts
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Gate the free tier on configuration: BOTH the shared API key and the
+ * rate-limit binding must be present, else `null` (fail-closed — callers
+ * return the same 401 the Worker served before the free tier existed).
+ * The limiter is shape-checked (not just truthy) so a misconfigured
+ * binding fails here, at the gate, instead of throwing mid-request.
+ */
+export function resolveFreeTierConfig(env: Env): FreeTierConfig | null {
+  const apiKey = env.FREE_TIER_API_KEY;
```
robertabbott: Should this `.trim()` the secret? `wrangler secret put` fed from a pipe or file picks up a trailing newline, which passes the `length === 0` gate and then 401s every anonymous request against Django — indistinguishable from a wrong key, and with no distinct log line. Cheap to rule out here at the gate, which is where the rest of the shape checking already lives.

workers/src/freetier.ts
```diff
@@ -0,0 +1,200 @@
...
+ * shared account's Tako credits. Handshake and discovery methods
+ * (`initialize`, `tools/list`, notifications, pings) must stay unmetered
+ * or clients would burn quota (or get 429s) just connecting. Array bodies
+ * never reach here — `checkFreeTierRateLimit` rejects batches before the
+ * metering decision — so no recursion into batch elements is needed.
+ * Anything unparseable / non-object is unmetered: it can never reach
+ * Django, and the SDK will reject it with a proper JSON-RPC error on its
+ * own.
+ */
+export function isMeteredJsonRpcBody(body: unknown): boolean {
```
robertabbott, line 123: Two things this method-only check lets through:

1. Metering fires before the tool is known to exist. An anonymous `tools/call` naming a hidden tool (`tako_agent`, `get_credit_balance`) counts against the bucket and then returns "tool not found" — so a confused client can burn its whole minute without ever spending a Tako credit. Would gating on `params.name ∈ FREE_TIER_TOOL_NAMES` be better, or is counting the attempt deliberate?
2. Unmetered methods are unbounded. `initialize` / `tools/list` staying free is right, but they now need no credential at all and `/mcp` has no WAF in front of it. Worth a second, much higher bucket on those?

workers/src/freetier.ts
```diff
@@ -0,0 +1,200 @@
...
+    body = await request.clone().json();
+  } catch {
+    return "allowed";
+  }
+  if (Array.isArray(body)) return "batch";
+  if (!isMeteredJsonRpcBody(body)) return "allowed";
+
+  const key = request.headers.get("cf-connecting-ip") ?? "unknown";
+  try {
+    const { success } = await limiter.limit({ key });
```
robertabbott, line 305: Metering is cost-blind: `tako_answer` (130 s timeout, LLM synthesis) and `tako_available_data` (15 s lookup) both cost one unit. So the real spend rate against the shared account is 10 `tako_answer` calls per minute per IP, not 10 cheap lookups. Is a weight per tool, or a tighter separate bucket for `tako_answer`, worth it here — or is the Django cap meant to absorb that?

workers/src/freetier.ts
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * Limiter failure fails OPEN (see module header for why).
+ */
+export async function checkFreeTierRateLimit(
+  request: Request,
+  limiter: RateLimit,
+): Promise<"allowed" | "limited" | "batch"> {
+  let body: unknown;
+  try {
+    body = await request.clone().json();
```
robertabbott: This parses an unbounded, unauthenticated body — and then the transport parses it again, so the anonymous path now does two full parses plus the clone's buffering. Before this PR an anonymous request 401'd before any body was read, so this is new surface reachable with no credential. `django.ts` caps error-body reads at 8 KiB (`ERROR_BODY_MAX_CHARS`) for the same reason. Would a `Content-Length` check before the clone be worth adding?

workers/src/mcp.ts
```diff
@@ -971,36 +995,66 @@ export async function handleMcpRequest(
   request: Request,
   env: Env,
 ): Promise<Response> {
-  // Gate the whole endpoint behind Bearer auth. If the header is missing /
-  // malformed / empty, return a uniform 401 before invoking the SDK.
-  let bearer: string;
+  // Gate the endpoint behind Bearer auth — with one carve-out. A request
```
robertabbott, line 1019: The `handleMcpRequest` docstring just above (lines 984-988) is now false: "A missing / malformed / empty `Authorization` header short-circuits here with a uniform JSON-RPC 401 response — the SDK never processes unauthenticated traffic. `initialize` requires auth too." Missing is exactly the case that no longer 401s. Worth updating, since it is the contract doc a future reader will trust.

workers/wrangler.jsonc
```diff
@@ -32,6 +32,22 @@
...
+  // Free-tier rate limiter (10 metered requests / 60 s / IP). Cloudflare's
+  // rate-limiting API is configured via `unsafe.bindings`. Keep the
+  // `limit` value in sync with FREE_TIER_LIMIT_MESSAGE in
+  // `src/freetier.ts` — the upsell message states the number.
+  // NB: the binding alone does NOT enable the free tier; the
+  // FREE_TIER_API_KEY secret must also be set (fail-closed).
+  "unsafe": {
+    "bindings": [
+      {
+        "name": "FREE_TIER_RATE_LIMITER",
```
robertabbott, line 51: The limit `10` now lives in four places — three `unsafe.bindings` blocks and `FREE_TIER_LIMIT_MESSAGE` — with a README note to keep them in sync by hand. Would a test asserting `FREE_TIER_LIMIT_MESSAGE` contains the value parsed out of `wrangler.jsonc` catch the drift, so the upsell can never advertise a number the limiter does not enforce?

robertabbott (commented):
Reviewed as a deliberate edge-only shortcut rather than a full server-side free tier, so I have skipped the "put this in Django instead" argument. The plumbing is clean: fail-closed config gate, missing-vs-malformed preserved, both authenticated paths byte-identical, and the tier gate sits ahead of every other registration rule. I checked the leak paths and they hold — no `console.*` prints the token, `DjangoError` messages carry only path/method/status, and `get_credit_balance` is gated with a test. The batch finding was real, not theater: the SDK server maps every element of an array body through dispatch (`webStandardStreamableHttp.js:408`).

Given the shortcut, the Worker is the only control, so my comments concentrate on the three things that decide whether it is safe to leave running: does the 429 actually reach the user, does the per-IP key limit anything for the clients you are targeting, and how do you notice or stop runaway spend.

workers/src/freetier.ts
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * Limiter failure (either bucket) fails OPEN — see module header. The
+ * "failing open" error line is the runbook grep target.
+ */
+export async function checkFreeTierRateLimit(
+  request: Request,
+  config: FreeTierConfig,
+): Promise<FreeTierMeterResult> {
+  try {
+    const { success } = await config.globalLimiter.limit({ key: "global" });
```
robertabbott: A constant key makes the per-colo property the README notes further down ("an IP whose traffic hits two colos can see roughly 2x the limit") much more consequential than it is for the per-IP bucket: `"global"` gets an independent 1000/min counter in every colo the anonymous traffic lands in, so the enforced number is 1000 x (colos reached). Hosted hosts egressing from many regions — the stated target — maximize that fan-out. Cloudflare's ratelimit binding has no global mode, so this is not fixable in the Worker.

The good news is that it does not need to be. Every anonymous request authenticates as the one `FREE_TIER_API_KEY` user, and Django already counts per user in Redis (`django_redis.cache.RedisCache` in `production.py`, which is where DRF's `SimpleRateThrottle` counts): `/api/v3/search/` heads its stack with `_SEARCH_USER` at 720/min per user and `/api/v1/answer/` with `_DRF_USER` at 720/min, both in `app/backend/api/throttling/policy.py`. Those are genuinely global, and tighter than the 1000/min this bucket advertises.

So would it be more accurate for the module docstring and README to describe this bucket as per-colo burst shaping, and name Django's per-user throttle on the free-tier account as the platform-wide ceiling? Worth confirming `_GRAPH_TIER` covers `tako_available_data`'s `/api/beta/graph/*` calls the same way. And the still-open cost question from the earlier thread lands here too: whatever the real ceiling turns out to be, neither bucket weighs the tool, so it is that many `tako_answer` calls.

workers/README.md
```diff
@@ -15,3 +15,131 @@ npm test
...
+nothing to configure in the Cloudflare dashboard: both rate limiters are
+config-as-code in `wrangler.jsonc` and deploy with the Worker.
+
+1. **Deploy the Worker** (normal deploy flow). The
+   `FREE_TIER_RATE_LIMITER` and `FREE_TIER_GLOBAL_RATE_LIMITER` bindings
+   ship with it.
+2. **Create the dedicated free-tier Tako account** and generate its API
+   key from the account page. All anonymous traffic spends this one
+   account's credits.
+3. **Set Django-side limits/credits on that account.** This is the total
```
robertabbott: The credit throttles for these endpoints carry `bypassed_by=_CREDIT_BYPASS` in `app/backend/api/throttling/policy.py`, which reads "anonymous; MPP; ENTERPRISE plan; **PAYG billing mode (metered in USD instead)**" — `_ANSWER_CREDIT`, `_V3_CREDIT` and `_CONTENTS_CREDIT` all share it. So on a PAYG account the credit ceiling never rejects: spend meters in USD and never 402s, which would also make the new Django-402 mapping in `mcp.ts` dead code.

What does survive PAYG is the rate side — `_DRF_USER` and `_SEARCH_USER` are rate throttles with no bypass entry — so the account is bounded in requests/min but not in total dollars.

Since fail-open leans entirely on this step, is it worth naming the specific mechanism here (the per-user rate throttle, and the plan or billing mode required for a credit ceiling to apply at all) rather than "limits/credits"? Reads like the one item to nail down before the production secret goes in, rather than anything blocking the merge.

workers/src/freetier.ts
```diff
@@ -0,0 +1,476 @@
...
+    },
+  );
+}
+
+/**
+ * The global-ceiling message. Keep the number in sync with the
+ * `FREE_TIER_GLOBAL_RATE_LIMITER` bindings in `wrangler.jsonc` (asserted
+ * by the same drift test as `FREE_TIER_LIMIT_MESSAGE`).
+ */
+export const FREE_TIER_GLOBAL_LIMIT_MESSAGE =
```
robertabbott, line 387: This docstring says to keep the number in sync and that the same drift test asserts it, but `FREE_TIER_GLOBAL_LIMIT_MESSAGE` states no number ("at capacity right now"), and the global half of the drift test only asserts the three env bindings agree with each other — it never compares against the message. Should the docstring drop the sync claim, or is the missing number the thing to fix, so the global message becomes checkable the way the per-IP one is?

workers/src/mcp.ts
```diff
@@ -706,6 +733,18 @@ function registerTool(
...
         if (err instanceof DjangoError) {
+          // Free tier: the shared account exhausting its Tako credits is
+          // the tier's expected steady-state failure (the Django-side cap
+          // is the fail-open spend backstop). Surface it as upsell copy,
+          // not the raw billing error — which would read as a bug to an
+          // anonymous user who has no account to top up.
+          if (
+            callCtx.tier === "free" &&
+            err instanceof DjangoHttpError &&
+            (err.status === 402 || err.body.includes("PAYMENT_REQUIRED"))
```
robertabbott: `PaymentRequiredError` in `app/backend/knowledge/api/ga/v1/types.py` is served with 402 and `error_type: "PAYMENT_REQUIRED"`, so `err.status === 402` already covers credit exhaustion on its own. The `||` arm matches the token anywhere in the body at any status, so anything that echoes caller-supplied text back in a 4xx body could turn a validation error into "the free tier is out of shared capacity". What does the body arm cover that the status check does not?

workers/src/freetier.ts
```diff
@@ -0,0 +1,476 @@
...
+
+/**
+ * HTTP 429 for a request over the GLOBAL anonymous ceiling. This fires
+ * before the body is read, so no request id is available and the tripping
+ * request may not even be a `tools/call` — a plain 429 (which the SDK
+ * client surfaces as a transport error) is the honest answer here, unlike
+ * the per-IP case above. It should be rare: the ceiling is sized well
+ * above expected legitimate volume.
+ */
+export function freeTierGlobalLimitResponse(): Response {
```
robertabbott: This is the shape the per-IP path was just changed to avoid — an SDK client throws on the non-2xx and the model never reads the message. The ordering is what forces it (the ceiling trips before the body is parsed), but the body read is now bounded and the `Content-Length` precheck is cheap, so size-check -> peek -> global -> per-IP would let a global-limited `tools/call` return the same readable tool result. Is "should be rare" enough to keep the unreadable shape here?

robertabbott (approved):
Delta review of `7afc6ea` against my last pass at `a704b25`. Ten of eleven prior comments are resolved; the cost-weighting one (metering `tako_answer` and `tako_available_data` at one unit each) is the only open thread, and I have folded a note about it into the first comment below rather than reopening it. Verified on the branch: `tsc --noEmit` clean, 453/453 tests pass. I also re-checked that the three-tool surface holds under the new tier gate (all three names are in `TOOL_REGISTRY`, none is optional or ChatGPT-gated) and that `tier` reaches `callCtx` through the spread.

Approving because the PR is inert on merge — without the secret `resolveFreeTierConfig` returns `null` and anonymous requests 401 exactly as today, and nothing on the authenticated path moves. The two substantive comments below are preconditions for setting the production secret, not for merging; they fit the rollout checkboxes that are already unchecked in the description.

PR #174, opened 2026-07-27
fix(plugin): stop empty-search retry loops + make the agent tool truly optional

25eliu:
## Summary

Fixes the Claude Code plugin behaving badly in the field: models were burning endless priced `tako_search` calls on zero-result responses, and the bundled skills pushed the opt-in agent tool.

### 1. Zero-result searches now carry a recovery protocol (`guidance` field)
A miss previously returned a bare `{"cards": [], "web_results": []}` — models answer that with billed reword-and-retry loops that never converge. `buildSearchOutput` now attaches a `guidance` string on **any zero-card response**, tailored to what was searched:

- **Zero cards, web hits** (the common default-source miss): work from the web results / `tako_contents`, don't re-search for a chart.
- **Both empty, data searched**: check `tako_available_data` (free) → at most ONE pinned retry (data-only callers are told that retry may add `web`) → stop.
- **Both empty, web-only**: web-shaped recovery, no node-pinning advice.

The field is declared in `searchOutputShape` (verified: zod v4's JSON-schema emission has no `additionalProperties: false`, so no client validation break; `tako_answer`/`tako_visualize` unaffected).

### 2. Skills no longer recommend `tako_agent`
It's an opt-in tool; cohort/ranking asks now steer to parallel narrow `tako_search` calls. Each skill also gained a HARD STOP rule on empty results (max 2 priced searches per entity+metric question). README's embedded skill copies are synced.

### 3. Plugin no longer force-enables the agent tools
`.claude-plugin/plugin.json` now points at the plain endpoint (was `?tools=agent`). Agent stays a per-connection opt-in via `?tools=`.

## Review
Ran /review with adversarial + specialist subagents; all findings fixed in the second commit — most notably the guidance originally required cards AND web both empty, which would have silently skipped the main retry-loop case. Stale docblocks (`tako_search.ts` header, `mcp.ts` throw-on-empty claim) corrected.

**Known transition note:** users who previously *copy-pasted* skill text from the README (rather than installing the plugin) keep stale copies recommending `tako_agent`; with agent no longer force-enabled they'll get a method-not-found on that one call and recover. Plugin users update atomically.

**Known transition note (update-path auth):** plugin users who had supplied an API key via the old `userConfig` land on the anonymous free tier after updating — the server serves unauthenticated requests as the free tier rather than a 401, so nothing errors, but the toolset shrinks to the three core tools at 10 req/min. Degradation is graceful (search/answer/available_data keep working) and the README plugin section now carries an update note directing them to `/mcp` → **Authenticate** (or the Connect button on Claude.ai) to restore full access.

## Test plan
- [x] `npm run typecheck` (all 4 tsconfig projects)
- [x] `npm run test` — 462 + 8 widget tests, including new unit tests for all four guidance variants, handler-level assertions, and an MCP SDK round-trip test pinning `structuredContent.guidance`
- [x] `npm run registry:check`
- [ ] After deploy: verify a zero-card search against prod returns `guidance`
- [ ] Reinstall the plugin and confirm the agent tools no longer register


jfassio (commented):
Ran a super-review (blast-radius, schema/widget verification, doc-sync check across the 4 hand-maintained HARD STOP copies, mental trace of all 5 guidance branches). Everything checked out except two things worth a look — neither blocking, CI's green and the rest of the diff is solid.

.claude-plugin/plugin.json
```diff
@@ -18,22 +18,10 @@
...
-  },
   "mcpServers": {
     "tako": {
       "type": "http",
-      "url": "https://mcp.tako.com/mcp?tools=agent",
-      "headers": {
-        "Authorization": "Bearer ${user_config.tako_api_key}"
-      }
+      "url": "https://mcp.tako.com/mcp"
     }
```
jfassio, line 25: The old manifest forced Bearer auth via `userConfig.tako_api_key` interpolated into a header; this drops to a bare URL with no headers. `freetier.ts` shows an unauthenticated request is served as the anonymous free tier (10 req/min, only `tako_search`/`tako_answer`/`tako_available_data`) *instead of* a 401 — so does an existing plugin user who already supplied their API key silently downgrade to the free tier on update, with no error or prompt, since nothing in the new manifest references their stored key? The PR's "Known transition note" covers README copy-paste users but not this case — is that an acceptable transition, or should update-path users get a heads-up (e.g. a note to run `/mcp` → Authenticate after updating)?

workers/src/tools/_search_results.ts
```diff
@@ -373,6 +377,59 @@ export type SearchOutput = {
...
 
+// Which sources a search actually hit, for tailoring zero-result guidance.
+// "tako" is a legacy alias for "data" (see tako_search's sources enum).
+// undefined (an older/other caller not threading sources) is treated as
+// the default both-sources search.
+type SearchedSources = readonly string[] | undefined;
+const searchedData = (s: SearchedSources): boolean =>
+  s === undefined || s.includes("data") || s.includes("tako");
+const searchedWeb = (s: SearchedSources): boolean =>
+  s === undefined || s.includes("web");
```
jfassio: `searchedData`/`searchedWeb` treat `sources === undefined` as "both" (per the doc comment, for "an older/other caller not threading sources"). `tako_search`'s zod schema defaults `sources` to `["data","web"]`, so it's never actually `undefined` from the only current caller — is this branch deliberate future-proofing for a caller that doesn't exist yet, or dead code worth dropping?

.claude-plugin/plugin.json
25eliu: Confirmed — your read is right: with the free-tier bindings live, an absent `Authorization` header is served as the anonymous free tier instead of a 401 (`mcp.ts`'s tier gate), so an updating keyed user silently drops to the three core tools at 10 req/min with no prompt. We're treating that as acceptable-with-heads-up: the silent free-tier landing is the intended new-user onboarding (free tier first, auth as the upgrade), and the downgrade is graceful rather than breaking — but update-path users now get the heads-up you suggested. Added in aded35b: a README plugin-section note ("Updating from an earlier plugin version?" → run `/mcp` → Authenticate, or Connect on Claude.ai), and the PR description's Known transition note now covers this case too.

workers/src/tools/_search_results.ts
25eliu: Dead code — dropped in aded35b. Verified `tako_search` is the only production caller and its zod schema always supplies `sources` (defaulted to `["data","web"]`), so `searchedSources` is now a required `readonly string[]` and the `undefined`-means-both branch is gone. Any future caller now has to thread sources explicitly (typecheck-enforced) rather than getting a silent both-sources assumption.

jfassio (approved):
Both review threads look correctly addressed:

- `.claude-plugin/plugin.json` silent free-tier downgrade: mitigated with a README callout for updating users (aded35b) — the right ceiling given a static manifest can't detect a prior key.
- `_search_results.ts` dead `undefined`-sources branch: confirmed `tako_search` is the only caller, dropped in aded35b, `searchedSources` now required and typecheck-enforced.

Verified independently (not just trusting the commit messages) — checked the diff line-by-line and confirmed no other tool file calls `buildSearchOutput`. Guidance-gating logic (keyed on `cards.length === 0`, not both-empty) and test coverage across the four guidance variants look solid. No blocking issues.

PR #175, opened 2026-07-27
fix(free-tier): declare rate limiters under ratelimits, not unsafe.bindings

robertabbott:
## Summary

**The anonymous free-tier per-IP rate limit does not enforce.** Found while running #171's own staging verification step.

Measured against deployed staging (secret set, bindings present):

- **15** metered anonymous `tools/call` requests in **2 seconds** against a `10 requests/60s` bucket — **all admitted**
- Across the session: **49** requests, **49 × `[free-tier] ip=<key> allowed`**, **0 × `limited`**, **0 × `failing open`**
- `wrangler versions view` reported the binding correctly the whole time: `env.FREE_TIER_RATE_LIMITER (10 requests/60s)  Rate Limit`

### Cause

#171 declared both limiters under `unsafe.bindings` with `type: "ratelimit"`. That key is a raw passthrough to the Cloudflare API — the deploy is accepted, the binding renders with the right numbers, and `limit()` resolves without throwing. It just never counts, so every call returns `{success: true}`. A silent no-op.

Cloudflare documents a first-class `ratelimits` key, and wrangler 4.87 carries schema support for it on both `RawConfig` and `RawEnvironment` (`simple.period` enum `[10, 60]`).

### Change

Move all three blocks — top-level (`tako-mcp-dev`), `env.staging`, `env.production` — from `unsafe.bindings` to `ratelimits`, dropping the now-redundant `type` field:

```jsonc
// before                                    // after
"unsafe": {                                  "ratelimits": [
  "bindings": [                                { "name": "FREE_TIER_RATE_LIMITER",
    { "name": "FREE_TIER_RATE_LIMITER",          "namespace_id": "1001",
      "type": "ratelimit",                       "simple": { "limit": 10, "period": 60 } },
      "namespace_id": "1001",                  ...
      "simple": { ... } },                   ]
```

`namespace_id` values are unchanged (1000/1010 dev, 1001/1011 staging, 1002/1012 prod), so the same namespaces are reused — the old ones never accumulated state.

**Also lowers the per-colo ceiling from 1000 to 120 requests/60s** per request from @robertabbott. Worth stating plainly in case it matters downstream: the binding has **no per-worker granularity** — 120 is enforced per Cloudflare colo, so the effective platform-wide number is `120 × colos reached`. It is also 2 qps, not 3. If a true global qps bound is wanted, the Worker binding cannot express it; that lives in Django's per-user throttling on the free-tier account (or a Durable Object).

### Scope / blast radius

No source change. `resolveFreeTierConfig` resolves bindings **by name**, which is unchanged, and `limit({key})` is the same runtime API.

The metered branch in `handleMcpRequest` is reachable only when the `Authorization` header is **fully absent**:

```js
if (bearer === null && freeTier !== null) {      // ← only limiter call site
    const meterResult = await checkFreeTierRateLimit(request, freeTier);
} else if (bearer !== null) {                     // raw API key or OAuth JWT
    tier = "authenticated";
}
```

So a paid user's own API key or OAuth token never touches either bucket — including the constant-key global one. A malformed or empty header still 401s and is never downgraded to the free tier.

Side effect: the `▲ WARNING "unsafe" fields are experimental` line that printed on every wrangler invocation in this repo is gone.

## Test plan

- [x] `npm run typecheck` — exit 0
- [x] `npm test` — 456 passed (workers project). The `test/widget` project fails locally on node 22.3.0 with a `require()` of ES Module error in jsdom's `html-encoding-sniffer`; **confirmed pre-existing on clean main** with the same node, and CI runs node 22.23.x where `main` passes.
- [x] `wrangler deploy --dry-run --env staging` and `--env production` — both resolve `FREE_TIER_RATE_LIMITER (10 requests/60s)` and `FREE_TIER_GLOBAL_RATE_LIMITER (120 requests/60s)` as `Rate Limit`, no `unsafe` warning
- [ ] **Live burst against deployed staging** — 15 sequential metered calls, expecting 1–10 admitted and 11+ returning `Free tier limit reached (10 requests/min)`. This is the only check that can actually catch the bug, and it cannot run pre-merge.

**No unit test can cover this class of defect** — the suite injects fake limiters and passes whether or not the real binding counts. That is why #171 shipped green with an inert limiter. The README now states the verification requirement at the point where limits get changed.

## Lines changed

- logic: 0
- config: 119 (`wrangler.jsonc`, 3 blocks + comments)
- docs: 39 (`README.md` 22, `env.ts` comments 17)
- tests: 1 (comment only)

## Note

Staging currently serves the anonymous free tier with **no working per-IP cap** (left up deliberately). Bounded by the free-tier account's $24.41 balance and Django's 720/min per-user throttle. **Production should not be enabled until this is merged, deployed, and the live burst passes** — its config had the identical defect.

Production deploy is still manual: release-triggered deploys fail on a missing Workers Routes:Edit permission on the `tako.com` zone (7 consecutive failures since v0.8.2). Tracked separately.

### DIFF
--- a/workers/README.md
+++ b/workers/README.md
@@ -26,12 +26,19 @@ requests 401 exactly as before):
 | Binding | Kind | Purpose |
 |---|---|---|
 | `FREE_TIER_API_KEY` | secret | Tako API key of the dedicated free-tier account, forwarded to Django as `X-API-Key` (trimmed, so a piped `wrangler secret put` newline can't break it) |
-| `FREE_TIER_RATE_LIMITER` | `unsafe.bindings` ratelimit in `wrangler.jsonc` | per-IP fairness bucket: 10 free-tool `tools/call`s / 60 s |
-| `FREE_TIER_GLOBAL_RATE_LIMITER` | `unsafe.bindings` ratelimit in `wrangler.jsonc` | per-colo burst shaping: 1000 anonymous requests / 60 s / colo, all callers |
+| `FREE_TIER_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-IP fairness bucket: 10 free-tool `tools/call`s / 60 s |
+| `FREE_TIER_GLOBAL_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-colo burst shaping: 120 anonymous requests / 60 s / colo, all callers |
+
+Declare both under the first-class `ratelimits` key. **Never under
+`unsafe.bindings`** — that path is a raw passthrough: the API accepts it,
+`wrangler versions view` renders the limits correctly, and `limit()`
+resolves without throwing, but it never counts, so every call returns
+`{success: true}`. No unit test catches it (the suite injects fake
+limiters); the only proof is a live burst against a deployed Worker.
 
 Neither Worker bucket is the platform-wide spend bound. Cloudflare's
 ratelimit binding counts per colo (no global mode), so the constant-key
-bucket enforces `1000/min × colos reached` — it exists to shape per-colo
+bucket enforces `120/min × colos reached` — it exists to shape per-colo
 floods, including the otherwise-unmetered handshake methods. Per-IP
 keying means little for hosted MCP hosts (claude.ai, ChatGPT, and
 similar egress from a handful of platform IPs); it's a fairness layer.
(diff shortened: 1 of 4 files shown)


25eliu (approved):
Verified locally on the branch: typecheck clean across all 4 tsconfig projects; 456/456 workers tests pass (the vitest workers pool loads wrangler.jsonc directly, so the new `ratelimits` key parses in the runtime config path too); wrangler 4.87.0's config schema carries `ratelimits` at top level and per-env; and `wrangler deploy --dry-run` for both staging and production resolves both bindings as `Rate Limit` at 10/60s and 120/60s with the `unsafe` warning gone. The drift test extracts limits by binding name from the raw JSONC, so it's location-agnostic and still guards message↔config sync.

No source change, bindings resolved by name, metered branch unchanged — blast radius is config-only. LGTM.

One reminder for whoever deploys: as the PR itself notes, the post-deploy live burst against staging (expect 11+ of 15 metered calls rejected) is the only check that actually proves the limiter counts — don't skip it, and don't enable production's free tier until it passes.

PR #176, opened 2026-07-27
fix(anonymous-tier): stop advertising an unenforceable rate limit; credits are the bound

robertabbott:
## Summary

The anonymous `/mcp` tier advertised **"10 requests/min"** per IP. Live measurement against deployed staging proved that is not enforceable, so this branch stops claiming it and documents what the limiter actually does.

**This branch changes no limiter value.** It originally halved the per-colo bucket; review established that value was never measured and that lowering it rejects legitimate handshakes, so it was reverted (see the thread on `wrangler.jsonc`). Behaviour is unchanged; copy, tests, and docs are what move.

### What the measurements showed

| Per-IP config | Cold first burst (of 200) | Sustained, converged | Slow drip, 20 requests at 3 s |
|---|---|---|---|
| `limit: 10` | 104–118 admitted | ~53–64 per 200 | **20/20 admitted, 0 limited** |
| `limit: 1` | ~118 admitted | ~19–25 per 200 | 13/20 admitted, 7 limited |

One IP had **292 requests admitted in ~16 seconds** against a 10/min limit.

Three properties follow: the configured limit controls *sustained* admission but not burst admission; cold start leaks ~115 requests **regardless of the limit**, so a caller who pauses one period gets a fresh leak; and normal-paced traffic is never limited at all — which is the path a real client takes, making the advertised number false in the common case.

Cloudflare documents the binding as *"permissive, eventually consistent, and intentionally designed to not be used as an accurate accounting system"*, with counters *"cached on the same machine and updated asynchronously"*.

### Changes

- **`wrangler.jsonc`** — no value changed. Comments corrected: they previously told operators to keep the limit in sync with the user-facing message, which is now the opposite of what the drift test enforces. The per-colo block now records that its value is unmeasured and that it gates `initialize`/`tools/list`, so lowering it rejects handshakes.
- **Copy** — all five user-facing strings drop rate figures, drop the word "free" (there is no default free tier, so promising a "free API key" is inaccurate), and point at `https://tako.com/account/` instead of the `trytako.com` redirect. Wording is now "anonymous access". The fifth string was an HTTP 413 body reading *"exceeds the free-tier limit"* — found during review, previously unowned.
- **Tests** — the drift test previously asserted the advertised number **matched** config, actively protecting a false promise. It is inverted: it now fails if any message states a rate, says "free", or uses the old host. Two new assertions guard `period` staying 60 and the six `namespace_id`s staying distinct (a duplicate would silently merge the per-IP and per-colo counters).
- **`README.md`** — records the measurements so nobody rediscovers them, plus the two dead ends: `unsafe.bindings` was tried and changed nothing, and a shorter `period` admits more. Rollout is reframed around monitoring credits with **no prescribed budget figure**.

### Two factual corrections to the README

Both cost real debugging time during rollout:

1. It claimed PAYG **"never 402s — unbounded dollars"**. False — `backend/subscriptions/decorators.py:249-261` returns 402 at `balance_cents <= 0`. The real hazard is **auto-reload with a card on file**, which refills the cap.
2. `GET /api/v1/credit_balance/` — and therefore the `get_credit_balance` MCP tool — reads the **legacy Metronome/Redis ledger**, not `ApiCreditAccount.balance_cents`. It reported `$0.00` for an account holding `$24.41`. The doc now names `GET /api/v1/billing/api_credits/`, which returns the authoritative balance plus `has_saved_card` and `auto_reload.enabled`.

## Disclosure — the structured fields now match the vague prose

The copy was made deliberately vague, but the machine-readable `kind` fields ship to callers and restated what the prose withheld. Two fixes:

- **`free_tier_credits_exhausted` → `capacity`.** That kind named the internal mechanism and told an anonymous caller — who has no account here — that a **shared** account had run out of **credits**. Probing it would then reveal how depleted that account is.
- **`global_rate_limited` — kept distinct.** Collapsing it into `rate_limited` was tried and reverted in review: the two messages describe different situations, so the topology reads off `error.message` regardless. The collapse broke a client-visible contract for no gain. Hiding the topology would require the messages to converge first, which costs the caller the difference between "slow down" and "come back later".

A guard test covers the kinds this module emits: it fails if any names the internal mechanism (`free_tier`) or the cause of the shortage (`credit`, `billing`, `payment`, `exhaust`, `depleted`, `quota`). It deliberately does NOT ban `shared` — the prose says the capacity is shared on purpose, because that tells the caller the exhaustion is not their own doing. Mutation-verified.

**Deliberately not fixed here, and larger.** `djangoErrorToToolResult` still attaches the raw upstream response body, plus `path`, `status`, and `timeoutMs`, to `_meta["tako/error"]` for every Django error except the mapped 402. On the anonymous path that leaks DRF throttle text from the shared account — e.g. *"Request was throttled. Expected available in 41 seconds."* — which discloses that callers share an account and hands a prober its saturation. Closing it needs a tier-aware whitelist rather than pass-everything, which is a behavioural change to a shared helper and belongs in its own reviewed PR.

## Isolation — paying users are unaffected

No control flow changed; verified against the diff rather than asserted. Both `.limit()` call sites are reachable only through the `bearer === null && freeTier !== null` branch, so a request bearing a raw Tako API key or an OAuth JWT never touches either bucket and never gets the restricted 3-tool surface. `tier = "free"` is assigned at exactly one line; the surface gate cannot restrict anything without it, and the default is `"authenticated"`. A malformed or empty `Authorization` header still 401s and is never downgraded.

## Review round

Seven comments, all verified against source before acting. Five fixed in `128b9f7`, two reverted in `32cd0a4`:

- **README claimed "the hazard is auto-reload, not the billing mode" — backwards.** `is_metered_request` returns true for an API-key request only when `billing_mode == PAYG`. On `CONTRACT` (`credit-exempt, externally billed`) the decorator returns the view before `balance_cents` is read, and `is_mpp` short-circuits identically, so **a CONTRACT or MPP account has no credit bound at all**. Step 3 now requires PAYG as the first of three conditions. It also restores the floor this branch had dropped: the per-user rate throttles apply on every billing mode.
- **`ApiCreditBalanceView` returns a hardcoded `balance_cents: 0` with `billing_mode: null`** when the user has no `enterprise_account` — indistinguishable from depleted, which is the confusion the switch away from the legacy endpoint was meant to end. Step 3 now says to read `billing_mode` first and treat `null` as "not wired up".
- **The guard banned `shared` while the prose says "out of shared capacity".** The prose was right; the ban was wrong.
- **The `period`/`namespace_id` scans were unanchored** over the whole file, so a later `"period"` key would fail the suite with a message about admitting more traffic. Both are now scoped to the `FREE_TIER_*` blocks and re-mutation-tested.
- **The guard's title overclaimed.** It covers only this module's kinds; `djangoErrorKind` also reaches anonymous callers. Title narrowed, uncovered surface named.

## Test plan

- [x] `npm run typecheck` — exit 0
- [x] `npm test` — **472 passed**. `test/widget/widget-dom.test.ts` fails locally on node 22.3.0 (`require()` of ES Module in jsdom's `html-encoding-sniffer`); confirmed identical on clean `origin/main`, passes in CI on node 22.23.x
- [x] `wrangler deploy --dry-run` both envs — report `FREE_TIER_RATE_LIMITER (10 requests/60s)` and `FREE_TIER_GLOBAL_RATE_LIMITER (60 requests/60s)`
- [x] Drift test mutation-tested five ways (old string restored, rate figure injected, "free" reintroduced, host reverted, upsell URL dropped) — all five fail correctly, so the test genuinely binds
- [ ] **Post-merge on staging:** confirm the deployed bindings still read `10` and `120`, that anonymous `tools/list` still returns exactly three tools, and that no limit message contains a rate figure. There is no longer a behavioural claim to prove — no limiter value changed — so this is a regression check, not a measurement.

## Lines changed

- logic: 0
- config: 25 (`wrangler.jsonc` — comments only, no value changed)
- tests: 148 (`freetier.test.ts`)
- source strings/comments: 78 (`freetier.ts` — string constants, one error `kind`, doc comments)
- docs: 152 (`README.md`)
- repo hygiene: 4 (`.gitignore`)

## Note

`.gitignore` gains `.worktrees/`, which the repo already used for local worktrees while leaving it untracked — a stray broad `git add` could have committed an entire nested checkout. Separable commit (`0a1eed0`); drop it if you would rather keep it independent.

**Production is deliberately untouched.** Its limiter has the same permissive behaviour and its account balance is unverified. Enabling it stays a separate decision.

### DIFF
--- a/workers/README.md
+++ b/workers/README.md
@@ -16,10 +16,10 @@ npm run typecheck
 npm run dev
 ```
 
-## Anonymous free tier
+## Anonymous access
 
 `/mcp` requests with **no** `Authorization` header are served as a
-rate-limited free tier instead of a 401 — but only in environments that
+rate-limited anonymous tier instead of a 401 — but only in environments that
 opt in. Three bindings gate it (fail-closed: with any missing, anonymous
 requests 401 exactly as before):
 
--- a/workers/README.md
+++ b/workers/README.md
@@ -115,7 +115,7 @@ Behavior when active:
   the body peek is new unauthenticated surface and its read is bounded,
   so a lying `Content-Length` doesn't help.
 - A malformed `Authorization` header still 401s — only a fully absent
-  header selects the free tier.
+  header selects the anonymous tier.
 - Limiter runtime failures fail open (the shared account's Django-side
   limits are the spend backstop); missing configuration fails closed.
 
@@ -150,7 +150,11 @@ config-as-code in `wrangler.jsonc` and deploy with the Worker.
    MCP tool — reads the legacy Metronome/Redis ledger via
    `BillingServiceSingleton.get_remaining_credit_balance`, NOT
    `ApiCreditAccount.balance_cents`. It reported $0.00 for an account that
-   actually held $24.41.
+   actually held $24.41. To check this mechanically instead of by eye, call
+   `GET /api/v1/billing/api_credits/` (`ApiCreditBalanceView`,
+   `backend/subscriptions/api_credit_views.py`). It returns the authoritative
+   `balance_cents`, plus `has_saved_card` and `auto_reload.enabled` — the
+   three facts this step asks you to confirm.
 4. **Enable staging first:**
    ```bash
    wrangler secret put FREE_TIER_API_KEY --env staging   # paste the API key
@@ -211,7 +215,7 @@ for abuse protection, not billing.
 
 **Operator warning:** OAuth-capable hosts (claude.ai and similar) decide
 whether to run their OAuth sign-in flow based on getting a 401 from the
-*first* request to `/mcp`. With the free tier active, that first request
+*first* request to `/mcp`. With anonymous access active, that first request
 succeeds anonymously instead — so any newly added connector on such a
 host starts out running against the shared free-tier account (3 tools,
 shared quota) rather than prompting the user to sign in for their own
--- a/workers/README.md
+++ b/workers/README.md
@@ -145,7 +145,7 @@ config-as-code in `wrangler.jsonc` and deploy with the Worker.
    more as needed with `add_api_credit --email <account> --amount <n>` or
    the `add-api-credit.yaml` Action.
 
-   **Read the balance from the account console, not the API.**
+   **Do not read the balance from the legacy `credit_balance` endpoint.**
    `GET /api/v1/credit_balance/` — and therefore the `get_credit_balance`
    MCP tool — reads the legacy Metronome/Redis ledger via
    `BillingServiceSingleton.get_remaining_credit_balance`, NOT
--- a/workers/README.md
+++ b/workers/README.md
@@ -27,7 +27,7 @@ requests 401 exactly as before):
 |---|---|---|
 | `FREE_TIER_API_KEY` | secret | Tako API key of the dedicated free-tier account, forwarded to Django as `X-API-Key` (trimmed, so a piped `wrangler secret put` newline can't break it) |
 | `FREE_TIER_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-IP fairness bucket: 10 free-tool `tools/call`s / 60 s |
-| `FREE_TIER_GLOBAL_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-colo burst shaping: 60 anonymous requests / 60 s / colo, all callers |
+| `FREE_TIER_GLOBAL_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-colo burst shaping: 120 anonymous requests / 60 s / colo, all callers |
 
 Declare both under the first-class `ratelimits` key. **Never under
 `unsafe.bindings`** — that path is a raw passthrough: the API accepts it,
@@ -38,7 +38,7 @@ limiters); the only proof is a live burst against a deployed Worker.
 
 Neither Worker bucket is the platform-wide spend bound. Cloudflare's
 ratelimit binding counts per colo (no global mode), so the constant-key
-bucket enforces `60/min × colos reached` — it exists to shape per-colo
+bucket enforces `120/min × colos reached` — it exists to shape per-colo
 floods, including the otherwise-unmetered handshake methods. Per-IP
 keying means little for hosted MCP hosts (claude.ai, ChatGPT, and
 similar egress from a handful of platform IPs); it's a fairness layer.
@@ -81,6 +81,24 @@ Do NOT reach for `unsafe.bindings` or a shorter `period` to fix this.
 `unsafe.bindings` was tried and changed nothing. A shorter period resets the
 counter more often, and because the cold-start leak recurs per period, it
 admits MORE traffic.
+
+**The per-colo bucket has NOT been measured.** Every row above varies the
+per-IP bucket. Sustained bursts showed per-colo rejections rising while
+per-IP rejections fell, so the per-colo bucket is the one that clamps under
+load — but no measurement supports any particular value for it, and 120 is
+not tuned. Lowering it was tried in this branch and reverted for that
+reason.
+
+Before changing it, note what it gates. `checkFreeTierRateLimit` hits the
+per-colo bucket **before** the batch check and before the metering decision,
+so `initialize` and `tools/list` count against it even though they spend
+nothing. A connector opening a new session costs roughly three requests, so
+the ceiling divided by three is the rough budget of new anonymous sessions
+per minute per colo. Past it, a handshake gets a plain 429 — not a 401, so
+an OAuth-capable host shows no sign-in prompt, and not a tool result, so the
+model reads nothing. It just looks broken. Any retune therefore needs a
+measurement of **sessions per minute per colo until handshakes fail**, not
+just admitted-call counts, which cannot see a rejected handshake at all.
 
 Per-call cost is bounded by the tool schemas: the anonymous toolset cannot
 select the expensive `deep` tier. `tako_answer` does not expose `effort` and
--- a/workers/README.md
+++ b/workers/README.md
@@ -134,16 +134,37 @@ config-as-code in `wrangler.jsonc` and deploy with the Worker.
    account's credits. Create it in the SAME environment the Worker points
    at: a production key returns 401 against `staging.tako.com` and vice
    versa, which is indistinguishable from a wrong key.
-3. **Fund the account and leave no card on file.** The pre-paid credit
-   balance is the abuse bound: the Worker limiters only dampen the burn
-   rate, so exhaustion is the real backstop. `@meters_api_credits`
-   pre-gates each request and returns 402 when `balance_cents <= 0` (see
-   `backend/subscriptions/decorators.py`). The hazard is **auto-reload**,
-   not the billing mode: `maybe_auto_reload` refills the balance when
-   auto-reload is enabled AND a card is attached, which would turn the cap
-   into a soft one. Keep auto-reload off. Monitor the balance and allocate
-   more as needed with `add_api_credit --email <account> --amount <n>` or
-   the `add-api-credit.yaml` Action.
+3. **Put the account on PAYG, fund it, and leave no card on file.** The
+   pre-paid credit balance is the abuse bound: the Worker limiters only
+   dampen the burn rate, so exhaustion is the real backstop. Three
+   conditions must all hold, and the first is easy to miss:
+
+   - **`billing_mode` must be `PAYG`.** The 402 gate runs only for metered
+     requests, and `is_metered_request` (`backend/subscriptions/api_metering.py`)
+     returns true for an API-key request **only** when
+     `ent.billing_mode == BillingMode.PAYG`. On a `CONTRACT` org
+     (`credit-exempt, externally billed`) `@meters_api_credits` returns the
+     view before `balance_cents` is ever read: no 402, and the funded
+     balance is never drawn down. `is_metered_request` also short-circuits
+     on `request.is_mpp`. A CONTRACT or MPP account therefore has **no
+     credit bound at all**.
+   - **The balance must be funded.** `@meters_api_credits` pre-gates each
+     request and returns 402 when `balance_cents <= 0` (see
+     `backend/subscriptions/decorators.py`).
+   - **Auto-reload must be off, with no card attached.**
+     `maybe_auto_reload` refills the balance when auto-reload is enabled
+     AND a card is on file, which turns the cap into a soft one.
+
+   Monitor the balance and allocate more as needed with
+   `add_api_credit --email <account> --amount <n>` or the
+   `add-api-credit.yaml` Action.
+
+   **The floor that holds either way.** Even when the credit gate is
+   bypassed, Django's per-user *rate* throttles still apply to the shared
+   account on every billing mode: `_SEARCH_USER` and `_DRF_USER` at
+   720/min, `_GRAPH_TIER` at 180/min plus 10,000/day
+   (`backend/api/throttling/policy.py`). Those bound request volume, not
+   spend.
 
    **Do not read the balance from the legacy `credit_balance` endpoint.**
    `GET /api/v1/credit_balance/` — and therefore the `get_credit_balance`
@@ -152,9 +173,16 @@ config-as-code in `wrangler.jsonc` and deploy with the Worker.
    `ApiCreditAccount.balance_cents`. It reported $0.00 for an account that
    actually held $24.41. To check this mechanically instead of by eye, call
    `GET /api/v1/billing/api_credits/` (`ApiCreditBalanceView`,
-   `backend/subscriptions/api_credit_views.py`). It returns the authoritative
-   `balance_cents`, plus `has_saved_card` and `auto_reload.enabled` — the
-   three facts this step asks you to confirm.
+   `backend/subscriptions/api_credit_views.py`). It returns
+   `balance_cents`, `billing_mode`, `has_saved_card`, and
+   `auto_reload.enabled` — every fact this step asks you to confirm.
+
+   **Read `billing_mode` before you trust the balance.** That view returns
+   a hardcoded `balance_cents: 0` with `billing_mode: null` when the user
+   has no `enterprise_account` at all, which is indistinguishable from a
+   genuinely depleted account. Treat `billing_mode: null` as "not wired
+   up", not as "out of credits", and treat anything other than `PAYG` as
+   "no credit bound".
 4. **Enable staging first:**
    ```bash
    wrangler secret put FREE_TIER_API_KEY --env staging   # paste the API key
--- a/workers/README.md
+++ b/workers/README.md
@@ -27,7 +27,7 @@ requests 401 exactly as before):
 |---|---|---|
 | `FREE_TIER_API_KEY` | secret | Tako API key of the dedicated free-tier account, forwarded to Django as `X-API-Key` (trimmed, so a piped `wrangler secret put` newline can't break it) |
 | `FREE_TIER_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-IP fairness bucket: 10 free-tool `tools/call`s / 60 s |
-| `FREE_TIER_GLOBAL_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-colo burst shaping: 120 anonymous requests / 60 s / colo, all callers |
+| `FREE_TIER_GLOBAL_RATE_LIMITER` | `ratelimits` entry in `wrangler.jsonc` | per-colo burst shaping: 60 anonymous requests / 60 s / colo, all callers |
 
 Declare both under the first-class `ratelimits` key. **Never under
 `unsafe.bindings`** — that path is a raw passthrough: the API accepts it,
@@ -38,7 +38,7 @@ limiters); the only proof is a live burst against a deployed Worker.
 
 Neither Worker bucket is the platform-wide spend bound. Cloudflare's
 ratelimit binding counts per colo (no global mode), so the constant-key
-bucket enforces `120/min × colos reached` — it exists to shape per-colo
+bucket enforces `60/min × colos reached` — it exists to shape per-colo
 floods, including the otherwise-unmetered handshake methods. Per-IP
 keying means little for hosted MCP hosts (claude.ai, ChatGPT, and
 similar egress from a handful of platform IPs); it's a fairness layer.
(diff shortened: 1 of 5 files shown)


workers/README.md
```diff
@@ -93,27 +129,43 @@ config-as-code in `wrangler.jsonc` and deploy with the Worker.
...
+   its API key from the account page. All anonymous traffic spends this one
+   account's credits. Create it in the SAME environment the Worker points
+   at: a production key returns 401 against `staging.tako.com` and vice
+   versa, which is indistinguishable from a wrong key.
+3. **Fund the account and leave no card on file.** The pre-paid credit
+   balance is the abuse bound: the Worker limiters only dampen the burn
+   rate, so exhaustion is the real backstop. `@meters_api_credits`
+   pre-gates each request and returns 402 when `balance_cents <= 0` (see
+   `backend/subscriptions/decorators.py`). The hazard is **auto-reload**,
+   not the billing mode: `maybe_auto_reload` refills the balance when
```
robertabbott: Step 3 states **"The hazard is auto-reload, not the billing mode"** and cites `@meters_api_credits` returning 402 at `balance_cents <= 0`. That gate is guarded by `is_metered_request` (`app/backend/subscriptions/api_metering.py:102`), which for an API-key request returns `ent.billing_mode == BillingMode.PAYG`. `BillingMode.CONTRACT` is annotated "credit-exempt, externally billed" (`app/backend/monolith/models.py:788`), so on a CONTRACT org the decorator returns `view_func(...)` before `balance_cents` is ever read — no 402, and the funded balance is never drawn down. `is_metered_request` also short-circuits on `request.is_mpp`.

The text this PR replaces carried exactly that warning ("credit throttles are **bypassed** for PAYG billing … and for Enterprise/MPP accounts. So the free-tier account must be a standard credit-capped plan"), added in `c762db0` as a round-2 review fix. It also carried the one bound that survives either way — "the per-user *rate* throttles (720/min search+answer, 180/min graph) apply on every billing mode and are the floor either way" — which this PR drops.

Given the PR's thesis is that credits are the bound: should step 3 require the account be on PAYG (and say how to confirm it — `ApiCreditBalanceView` already returns `billing_mode`), and keep the rate-throttle floor as the statement that holds when the credit gate is bypassed?

workers/src/freetier.ts
```diff
@@ -406,8 +409,12 @@ export function freeTierGlobalLimitResponse(
         id: null,
         error: {
           code: -32000,
+          // Deliberately the SAME kind the per-IP bucket reports. Two
+          // distinguishable kinds would let a caller map which bucket it
+          // tripped, and so infer the limiter topology by probing. The
+          // message text still tells the caller what to do.
           message: FREE_TIER_GLOBAL_LIMIT_MESSAGE,
-          data: { kind: "global_rate_limited" },
+          data: { kind: "rate_limited" },
```
robertabbott: The comment here, and the test at `freetier.test.ts:640` ("Both limiter buckets must report the SAME kind **so neither is distinguishable from the outside**"), say the goal is that a caller cannot tell which bucket it tripped. But the two responses still carry different message text in the same JSON body: `FREE_TIER_LIMIT_MESSAGE` (line 327) is "Rate limit reached for anonymous access. Try again in a minute…" and `FREE_TIER_GLOBAL_LIMIT_MESSAGE` (line 391) is "Anonymous access is at capacity right now. Try again shortly…". A prober reads `error.message` as easily as `error.data.kind`, so the topology stays fully mappable and the `kind` change costs a client-visible contract break for no gain against the stated threat.

Should the two messages converge too (they'd then need to say the same actionable thing), or should the goal be restated as something the change actually delivers — and the test comment corrected either way?

workers/wrangler.jsonc
```diff
@@ -59,7 +61,7 @@
     {
       "name": "FREE_TIER_GLOBAL_RATE_LIMITER",
       "namespace_id": "1010",
-      "simple": { "limit": 120, "period": 60 }
+      "simple": { "limit": 60, "period": 60 }
```
robertabbott: Every anonymous request hits this bucket, including `initialize` and `tools/list` — `checkFreeTierRateLimit` calls `globalLimiter.limit({ key: "global" })` at `freetier.ts:219`, before the batch check and the metering decision. Per the README, a non-`tools/call` over the ceiling gets a plain 429, which an OAuth-capable host reads as neither a 401 (no sign-in prompt) nor a readable tool error — it just looks broken. At 60/min/colo a connector doing roughly initialize + tools/list + tools/call per new session gets ~20 new sessions/min/colo before legitimate handshakes start failing.

The measurement table covers the **per-IP** bucket at `limit: 10` vs `limit: 1`; nothing in it varies the per-colo bucket, and the test plan defers the 60 check to post-merge. Was the availability cost to legitimate anonymous sessions at 60 considered, or is the reasoning that a rejected handshake is an acceptable trade for the halved burn rate?

workers/README.md
```diff
@@ -93,27 +129,43 @@ config-as-code in `wrangler.jsonc` and deploy with the Worker.
...
+   the `add-api-credit.yaml` Action.
+
+   **Do not read the balance from the legacy `credit_balance` endpoint.**
+   `GET /api/v1/credit_balance/` — and therefore the `get_credit_balance`
+   MCP tool — reads the legacy Metronome/Redis ledger via
+   `BillingServiceSingleton.get_remaining_credit_balance`, NOT
+   `ApiCreditAccount.balance_cents`. It reported $0.00 for an account that
+   actually held $24.41. To check this mechanically instead of by eye, call
+   `GET /api/v1/billing/api_credits/` (`ApiCreditBalanceView`,
+   `backend/subscriptions/api_credit_views.py`). It returns the authoritative
```
robertabbott: The step switches to `GET /api/v1/billing/api_credits/` because the legacy endpoint reported `$0.00` for an account holding `$24.41`. But `ApiCreditBalanceView.get` returns a hardcoded `{"balance_cents": 0, "has_saved_card": false, …}` when `_enterprise_or_none(request)` is None (`app/backend/subscriptions/api_credit_views.py:50-54`), so an account with no `enterprise_account` reads as `0` here too — indistinguishable from genuinely depleted, which is the confusion the switch is meant to end. Should the step say to treat `billing_mode: null` in that response as "not wired up" rather than "out of credits"?

workers/src/freetier.ts
```diff
@@ -496,24 +513,30 @@ export function freeTierBatchResponse(): Response {
...
+ * burn-rate dampener rather than a bound. Exhaustion is therefore an expected
+ * steady state, and the operator tops the account up. Without this mapping the
+ * caller would see Django's raw billing error spliced into the tool result by
+ * `djangoErrorToToolResult`, which reads as a bug rather than an upsell.
  */
 export const FREE_TIER_CREDITS_MESSAGE =
-  "The Tako free tier is temporarily out of shared capacity. Get a free " +
-  "API key at https://trytako.com/account/ to keep going with your own " +
-  "account.";
+  "Anonymous access is temporarily out of shared capacity. Get an API " +
```
robertabbott, line 529: The doc comment at line 535 says the `kind` must not restate "what the prose withholds: … that the tier runs on one **shared**, credit-funded account", and the guard test bans `/shared/` from every kind. This message reads "Anonymous access is temporarily out of **shared** capacity" — so the prose discloses the sharing, and only the machine-readable field is scrubbed for it. Is "shared" load-bearing here (it does explain why capacity is not the caller's fault), and if so should the guard's rationale drop `shared` from the banned list so the two agree?

workers/src/freetier.test.ts
```diff
@@ -561,21 +574,107 @@ describe("wrangler.jsonc ↔ message drift", () => {
...
 
   it("global binding limits agree across all 3 envs", () => {
     const limits = bindingLimits("FREE_TIER_GLOBAL_RATE_LIMITER");
     expect(limits).toHaveLength(3);
     expect(new Set(limits).size).toBe(1);
   });
+
+  it("every bucket's period is 60 — a shorter period admits MORE traffic, not less (see README 'Measured behaviour')", () => {
+    const periods = allPeriods();
+    expect(periods).toHaveLength(6); // 3 envs x (per-IP + global)
```
robertabbott, line 610: `allPeriods()` and `allNamespaceIds()` run unanchored global regexes over the whole `wrangler.jsonc`, then assert `toHaveLength(6)`. Any future `"period"` key elsewhere in the file (observability, logpush, an analytics binding) makes this fail with "a shorter period admits MORE traffic" — a message that has nothing to do with the actual edit. Would scoping the scan to the `ratelimits` arrays, or matching `"name": "FREE_TIER_..."` blocks the way `bindingLimits` already does, keep the failure legible?

workers/src/freetier.test.ts
```diff
@@ -561,21 +574,107 @@ describe("wrangler.jsonc ↔ message drift", () => {
...
+    expect(periods.every((p) => p === 60)).toBe(true);
+  });
+
+  it("all six namespace_id values are distinct — reusing one would merge the per-IP and per-colo counters", () => {
+    const ids = allNamespaceIds();
+    expect(ids).toHaveLength(6); // 3 envs x (per-IP + global)
+    expect(new Set(ids).size).toBe(6);
+  });
+
+  it("no client-visible error kind discloses how the tier is built", async () => {
```
robertabbott: The title says "no client-visible error kind discloses how the tier is built" and `toHaveLength(5)` pins it to this module's kinds. `djangoErrorToToolResult` also emits a client-visible `_meta["tako/error"].kind` on the anonymous path via `djangoErrorKind` (`mcp.ts:1051`) — `unauthorized`, `timeout`, `not_found`, `bad_request`, `response_parse`, `http`, `unknown`. None trips the regex today, so there's no live gap, but the guard wouldn't catch a regression there. Worth narrowing the title to the freetier kinds, or extending the list?

workers/README.md
robertabbott: Verified and you are right on both counts. `is_metered_request` gates on `ent.billing_mode == BillingMode.PAYG`, and `BillingMode.CONTRACT` is annotated `credit-exempt, externally billed`, so on CONTRACT the decorator returns the view before `balance_cents` is read — no 402, balance never drawn. `is_mpp` short-circuits the same way. So a CONTRACT or MPP account has **no credit bound at all**, and "the hazard is auto-reload, not the billing mode" had it backwards: billing mode is the primary gate and auto-reload is secondary.

Worse, I had read `is_metered_request` earlier in this work and still wrote the opposite into the doc.

Fixed in 128b9f7. Step 3 now requires PAYG as the first of three conditions, explains why CONTRACT/MPP silently removes the bound, and says to confirm it from `billing_mode` in the `ApiCreditBalanceView` response. It also restores the rate-throttle floor this branch had dropped — 720/min `_SEARCH_USER` and `_DRF_USER`, 180/min plus 10,000/day `_GRAPH_TIER` — stated as bounding request volume rather than spend.

workers/README.md
robertabbott: Confirmed — `ApiCreditBalanceView.get` returns a hardcoded `{"balance_cents": 0, "has_saved_card": false, "billing_mode": null, ...}` when `_enterprise_or_none(request)` is None, so an unwired account is byte-identical to a depleted one on the field I told the operator to read.

Fixed in 128b9f7, using the discriminator already in that response: read `billing_mode` first, treat `null` as "not wired up" rather than "out of credits", and treat anything other than `PAYG` as "no credit bound". That also folds into the PAYG requirement from your other comment.

workers/src/freetier.ts
robertabbott: The prose is the part that is right, so I removed the ban rather than the word. Saying the capacity is shared tells the caller the exhaustion is not their own doing, which stops them hunting a fault in their own setup — that is worth disclosing.

What neither the prose nor the kind should say is that the shortage is *spent credit*, since that hands a prober a gauge for how depleted the account is. Fixed in 128b9f7: the guard now bans `credit`, `billing`, and the internal naming, and no longer bans `shared`. The doc comment on `freeTierCreditsToolResult` now states that distinction instead of claiming the prose withholds the sharing.

workers/src/freetier.test.ts
robertabbott: Agreed — and the failure message would have been actively misleading, since it asserts a causal claim about traffic that a stray `"period"` key has nothing to do with.

Fixed in 128b9f7. Both helpers are now scoped to the `FREE_TIER_*` blocks the way `bindingLimits` already was: `"name":\s*"FREE_TIER_[A-Z_]*"[\s\S]*?"period":\s*(\d+)`. Re-verified by mutation that each still fails when violated, and that both still find exactly six.

workers/src/freetier.test.ts
robertabbott: Right that the title overclaims. I narrowed it rather than extending the list, because extending it would imply the guard covers the Django path when the real exposure there is not the `kind` at all.

Fixed in 128b9f7: the test is now "no freetier error kind ...", with a SCOPE note naming the `djangoErrorKind` values it does not cover, recording that none of them disclose anything today, and pointing at the larger surface — the raw upstream body `djangoErrorToToolResult` attaches alongside them, which is tracked separately and is the thing that actually leaks on the anonymous path.

workers/src/freetier.ts
robertabbott: You are right that the topology stays mappable, and I had noticed the same thing after pushing — but I shipped a comment and a test asserting "so neither is distinguishable from the outside", which is false. That claim is the defect regardless of which way we resolve it.

On the trade you describe: the `kind` collapse buys nothing on its own, and it does break a client-visible contract. The threat it was aimed at is real though — knowing whether you tripped a per-IP or a per-colo bucket tells an attacker whether rotating IPs helps — so a half-measure is the one outcome with no upside.

Two coherent options, and I do not want to pick unilaterally because one of them costs user-visible behaviour:

**A. Converge the messages, keep the collapse.** One message for both buckets, saying the one actionable thing ("try again shortly"). Topology genuinely stops being mappable. Cost: an anonymous caller can no longer tell "I am going too fast" from "the service is busy", which are different situations with different retry strategies.

**B. Revert the collapse, fix the claim.** `global_rate_limited` comes back, no contract break, and the comment plus test say what is true — the kinds are descriptive and the topology is visible. Cost: the hardening does not happen.

I lean B. The distinction between the two messages has real user value, the per-colo bucket is per-colo anyway so its behaviour is already fuzzy from outside, and B is honest rather than half-hardened. But A is defensible if you want the topology hidden and are willing to spend the message distinction to get it.

Note this is independent of the `free_tier_credits_exhausted` to `capacity` change, which stands on its own: that kind named the mechanism and the cause while its prose deliberately did not.

workers/wrangler.jsonc
robertabbott: It was not considered, and you have found the weakest part of this PR.

Two things you are pointing at are both true. The global bucket is hit at `freetier.ts:219` before the batch check and before `isMeteredJsonRpcBody`, so `initialize` and `tools/list` count against it. And a non-`tools/call` over the ceiling gets a plain 429 with no readable tool result, which on an OAuth-capable host is neither a sign-in prompt nor a legible error. Your arithmetic holds: at three requests per new session, 60/min/colo is about 20 new anonymous sessions per minute per colo before legitimate handshakes start failing.

The deeper problem is that the measurement table does not support the change. Every row in it varies the **per-IP** bucket at 10 versus 1. Nothing in it varies the per-colo bucket, and the only reason I picked 60 was the observation that per-colo rejections rose across sustained bursts while per-IP rejections fell — which shows that bucket was the one clamping, not that halving it was safe or sufficient. The whole thesis of this PR is that we should not state limits we have not measured, and 120 to 60 is exactly that.

So I think the retune should come out of this PR:

**Revert the per-colo bucket to 120 here.** What remains is fully supported — the copy is honest, the drift test is inverted, the docs record real measurements, and the disclosure fixes stand. Then measure the per-colo bucket properly as its own exercise, including the handshake cost you raise (sessions per minute per colo until handshakes 429, not just admitted-call counts), and retune with data in a follow-up.

The alternative is to measure it now on staging before merging, which I can do with the existing harness. That is slower but keeps the retune in one PR.

Either way the availability question deserves its own measurement, because admitted-burst counts do not capture it at all: a handshake rejection is invisible in the metric I used.

robertabbott: Reverted to 120 in 32cd0a4 — all three blocks, plus the README table and the `120/min × colos reached` line.

The README now records what this exchange established, so the next person does not repeat it: that the per-colo bucket is **unmeasured**, that every row of the table varies only the per-IP bucket, that the bucket is hit before the batch check and before the metering decision so `initialize` and `tools/list` count against it, and that a retune needs **sessions per minute per colo until handshakes fail** rather than admitted-call counts — which cannot see a rejected handshake at all. The `wrangler.jsonc` comment says the same at the point where someone would edit the value.

So this PR now changes no limiter value. What is left is the part that was actually supported: the copy stops claiming a rate, the drift test is inverted, the docs record real measurements and the two billing corrections, and the disclosure fix stands.

workers/src/freetier.ts
robertabbott: Reverted in 32cd0a4. `global_rate_limited` is back, so no contract break.

The comment and the test no longer claim the topology is hidden. Both now say why the collapse was wrong: the two messages describe different situations, so a caller reads the topology off `error.message` regardless, and hiding it in one field while leaving it in the other was the one outcome with no upside. Converging the messages is recorded as the prerequisite if that ever becomes worth doing, along with what it costs — the caller loses the difference between "slow down" and "come back later".

The guard keeps its real job: no kind may name the internal mechanism or the cause of the shortage. It no longer bans `global`, and the `kinds[0] === kinds[1]` assertion is gone. Re-verified by mutation that restoring `free_tier_credits_exhausted` still fails it.

`capacity` stays, since that one was independently justified rather than part of the collapse.

25eliu (approved):
Reviewed the full diff plus the prior review round. Verified independently:

- **No logic change** — `freetier.ts` moves only string constants, one error `kind`, and doc comments; `wrangler.jsonc` hunks touch comment lines only (10/120/60 values unchanged).
- **`free_tier_credits_exhausted` → `capacity`** is safe repo-side: the old kind is referenced nowhere else in this codebase, and the disclosure rationale is sound.
- **New `FREE_TIER_TOO_LARGE_MESSAGE`** is defined after `MAX_FREE_TIER_BODY_BYTES`, so no TDZ issue; stating the exactly-enforced byte cap while banning rate figures is the right distinction.
- **Drift-test inversion** is the correct move given the measurements — a test that pinned the advertised number to config was protecting a false claim. The `period`/`namespace_id` guards are properly scoped to the `FREE_TIER_*` blocks and CI proves the regex ordering assumptions hold.
- The deferred `djangoErrorToToolResult` body-leak fix belongs in its own PR as stated.

CI green (lint + typecheck-and-test). LGTM.

PR #179, opened 2026-07-28
feat(tools): break the punt-and-retry loop — dense first responses, deterministic verdicts, discovery handles

25eliu:
> Includes the two skills-correction commits also open as #178 (this branch builds on them). Merging this PR makes #178 redundant — close it, or merge #178 first and this PR will show only the three feature commits.

## Why

Agent trace analysis of 22 real runs: 18 cascaded through 3–6 tool waves ($0.14–$0.38/run) instead of finishing in 2 steps ($0.05). 78% of input tokens were spent in waves 3+, 60% on re-sent history. Three root causes: wave-0 answers **tease** instead of deliver, discovery runs **last** instead of first, and soft misses invite **rephrase-retry loops**.

## What

**Dense first responses (the trigger)**
- `tako_answer`: new `include_contents` (default **true**) + `preview_rows` (default 20 = the free inline allowance, max 250 held for a future backend row-count knob) — cited data cards inline their most-recent rows with the prose. Previously the handler pinned `include_contents: false` on the wire *and* stripped any rows that rode along, guaranteeing the teaser.
- `tako_search`: new `preview_rows` knob capping the inline preview per call (default 20 — the server-side free inline allowance — direction-aware most-recent rows; the real preview win is 5 → 20).
- Card payload layout (both tools): data-first key order — `content` before descriptive metadata, rows before descriptors — so size-capped clients truncate boilerplate, not data; identical per-card source/methodology paragraphs (≥120 chars) deduped to a pointer after first occurrence.

**Deterministic verdicts (the retry loop)**
- `tako_answer`: zero data cards while the data source was searched → machine-checkable `guidance` field ("NOT in Tako's data index for this phrasing — don't rephrase-retry; one `tako_available_data` check or pivot to web").

**Discovery-first (the wrong order)**
- `tako_available_data`: the coverage drill now **paginates** graph/related (100/page, up to 150 names, hard 4-page ceiling, no-progress guards, partial-keep on later-page failures; spec-legal `relation: null` reads as zero coverage, not an outage). New `coverage_filter` input threads the server-side `q` substring filter across the full coverage. New `next_call` output = the follow-up `tako_search` (query + pinned `node_ids`) prewritten — discovery→fetch in one step.

**One-wave page fetches**
- `tako_contents`: new `query` input (MCP-layer only, stripped from the wire body) — deterministic passage extraction from the fetched web text: phrase-first then per-term matching with an even per-term match budget, ±1.5k-char merged windows, max 8 passages; a miss returns an explicit NOT FOUND verdict + page head instead of silence.

**Docs**: README tool table + financial/macro skill notes updated to match.

## Review

Two independent adversarial review passes ran over the batch; all confirmed findings fixed and regression-tested, including: `relation: null` misclassified as an outage (HIGH), pagination round-trip ceiling, infinite-loop guards, and per-term match starvation in passage extraction (HIGH — a common early term could silently crowd the answer-bearing rare term out while reporting `matched: true`).

## Test plan

- [x] `npm run typecheck` — clean across all 4 tsconfig projects
- [x] `npx vitest run` — 509/509 (68 new/updated tests: pagination edges, ceiling boundaries, partial-failure paths, guidance gating incl. legacy `"tako"` alias and web-only asks, key ordering, dedupe immutability, preview_rows bounds/inertness, passage extraction incl. the starvation regression)
- [ ] Post-deploy: verify on staging that a default `tako_answer` for a covered metric returns inline rows on cited cards, and `tako_available_data` for Capital One surfaces "Net charges-off…" in `coverage.names`


robertabbott (commented):
Four issue-level findings inline (two of them false claims already published to `registry/server.json` / README / `llms-full.txt`) plus two suggestions. Each was reproduced against the branch before posting; the passage-extraction one is reproducible with a spacing tweak to the existing regression test.

workers/src/tools/_search_results.ts
```diff
@@ -127,11 +127,16 @@ export const webResultSchema = z
...
-// separate tako_contents call.
-export const INLINE_PREVIEW_ROW_CAP = 5;
+// keeps the model-facing peek bounded. 50 rows covers the common asks
+// outright (a decade of quarterly data, ~two months of daily prices) so the
+// agent usually needs no follow-up export; the full, priced export is always
+// a separate tako_contents call. The agent can adjust the cap per call via
+// tako_search's `preview_rows` (up to MAX_PREVIEW_ROWS — the backend ships
+// the series either way; this only governs how much reaches the model).
+export const INLINE_PREVIEW_ROW_CAP = 50;
+export const MAX_PREVIEW_ROWS = 250;
```
robertabbott, line 142: The backend caps the search/answer inline preview at `tako_inline_cap_for(user_info)` — `CSV_FREE_ROWS = 20`, or `CSV_INLINE_ROW_CAP = 1000` only for `INLINE_CONTENTS_LEGACY_CAP_COMPANIES = {you.com, youdotcom}` (`auto_contents.py:253`, `content_types.py:53`). `DataSourceSettings` exposes no row-count knob, so the worker can only cap *down* what already arrived. That makes `preview_rows` inert at any value >= 20 for every non-you.com caller, and the published claims — registry `"default 50, max 250"`, README `"50 by default, preview_rows up to 250"`, and this tool description's `"Raise it toward 250 when you know you need the longer series inline"` — describe rows that never ship. The comment on line 137 ("the backend ships the series either way; this only governs how much reaches the model") looks like the inverted premise.

The real win here is 5 -> 20 rows, which is genuine. Is the plan to raise the backend cap in a paired PR, or should the ceiling and the docs come down to 20 first?

Related: `tako_search.test.ts:490` and `tako_answer.test.ts:423` mock 300-row datasets, so the tests cannot catch this.

workers/src/tools/_passages.ts
```diff
@@ -0,0 +1,150 @@
...
+        head,
+      ].join("\n"),
+      matched: false,
+      truncated: text.length > head.length,
+    };
+  }
+
+  const windows = mergeWindows(
+    hits.map((i): [number, number] => [Math.max(0, i - WINDOW), Math.min(text.length, i + WINDOW)]),
+  ).slice(0, MAX_PASSAGES);
```
robertabbott: The per-term budget fixes starvation at the *match* level, but `windows.slice(0, MAX_PASSAGES)` then keeps the first 8 windows in **document order** — so a common early term can still crowd the rare term's window out.

Reproduced against this implementation: 12 `occupancy` hits spaced 4000 chars apart (wider than the 3000-char merge span, so they do not merge) followed by one late `RevPAR` hit, query `"RevPAR occupancy 2031"` — the header reads `13 match(es) for terms of "RevPAR occupancy 2031" — 8 passage(s)`, `matched: true`, and `$142.11` is **absent**.

The regression test at `_passages.test.ts:72` passes only because its 2400-char spacing is *inside* the merge span, collapsing all 100 common hits into a single window — so the position slice is never exercised there.

Should the slice be per-term round-robin (one window per term before a second from any), or should the header name which terms actually made it into the returned passages?

workers/src/tools/tako_answer.ts
```diff
@@ -160,19 +193,35 @@ const takoAnswer = {
...
     return {
       ...parsed.data,
-      cards: parsed.data.cards.map((c) => slimCard(c, null)),
+      cards,
       web_results: parsed.data.web_results.map(slimWebResult),
+      // Deterministic data-coverage verdict: the data source was searched and
+      // grounded NOTHING. Emitted regardless of how confident the prose
+      // sounds — rephrasing occasionally shakes a series loose, which teaches
+      // agents to retry forever; this converts that into one pivot.
+      ...(searchedData(input.sources) && cards.length === 0
```
robertabbott, line 238: This gates only on `searchedData(sources) && cards.length === 0`, with no branch on whether the answer succeeded or whether web grounded it. With the default `sources: ["data","web"]`, a question like `tako_answer("who won the 2026 Super Bowl")` returns a correct, fully web-cited answer *and* a verdict reading "ZERO curated data cards ground this answer — treat the metric as NOT in Tako's data index for this phrasing... Do NOT rephrase-and-retry".

`buildZeroResultGuidance` in `_search_results.ts:513` does branch on `hasWebResults` and softens the wording accordingly. Should this gate mirror that, or at least skip when `web_results` is non-empty and the prose is not a miss?

The test at `tako_answer.test.ts:450` only covers `sources: ["web"]`, which exits on the source check rather than exercising the data+web-answered-by-web case.

workers/src/tools/tako_available_data.ts
```diff
@@ -64,6 +73,9 @@ const inputSchema = z.object({
   label: z.enum(NER_LABELS).optional().describe(
     "NER label to prefer (boost, not a filter). Supply when you can categorize the term (company→ORG, place→GPE, person→PERSON, ...).",
   ),
+  coverage_filter: z.string().min(1).optional().describe(
+    'Hunting one specific metric? Case-insensitive substring filter on the coverage names + aliases, applied server-side across the FULL coverage (e.g. q="Capital One", coverage_filter="charge-off"). Use when the unfiltered list came back truncated without the metric you expected. Omit for the general coverage overview.',
```
robertabbott: `coverage_filter` is documented here (and in the registry and `llms-full.txt`) as a case-insensitive **substring** filter on the coverage names + aliases. That holds in prod today, but `graph_related_os_name_filter` is `enabled_staging: true` / `enabled_dev: true` (`app/fixtures/feature_flags.json` pk 113), and with it on, `q` matching becomes tokenized-AND rather than `CONTAINS`: `_matches_name_tokens` requires every query token as a whole token, with `_NAME_TOKEN_RE = [a-z0-9]+`.

So `coverage_filter="charge-off"` tokenizes to `{charge, off}` while `"Net charges-off"` tokenizes to `{net, charges, off}` — `charge` is absent, so it does **not** match. That is this PR's flagship documented example, and its own unchecked post-deploy item ("verify on staging that `tako_available_data` for Capital One surfaces 'Net charges-off...'") runs against exactly the flag-on semantics. `service.py:96` also notes the prod flip is planned, so the substring wording gets stale there too.

Should the description say "matches whole words" and the example become `coverage_filter="charges-off"`?

workers/src/tools/_search_results.ts
```diff
@@ -280,31 +285,137 @@ export function slimCardContent(
...
+ * ~1k-char description), which crowds the actual data out of any client-side
+ * result-size budget. The first occurrence always survives verbatim, so no
+ * information is lost. Immutable — untouched cards are returned as-is.
+ */
+export function dedupeCardBoilerplate(cards: TakoCard[]): TakoCard[] {
+  const seen = new Map<string, string>(); // text → path of its first occurrence
+  const fold = (value: unknown, path: string): unknown => {
+    if (typeof value !== "string" || value.length < DEDUPE_MIN_CHARS) return value;
+    const first = seen.get(value);
+    if (first !== undefined) return `[identical to ${first}]`;
```
robertabbott, line 382: `sources[].source_description` and `methodologies[].methodology_description` are the per-card provenance text, and commit 3dbff37 on this branch documents S&P Global / FactSet / Visible Alpha / CoinMarketCap / SimilarWeb as licensed protected sources. After this fold, cards 2..N carry `[identical to cards[0].sources[0].source_description]` in place of the attribution paragraph.

Nothing is lost within one payload, so this is fine for a model reading the whole response. But a client that renders a single card in isolation (picks card 3 to display) now shows the placeholder where the source attribution was. Is per-card attribution ever surfaced independently of the full response — and if so, should protected-source descriptions be exempt from the dedupe?

workers/src/tools/_available_data.ts
```diff
@@ -176,40 +189,62 @@ function labelSuffix(m: CoverageMatch): string {
...
+ * a handle for a name just reported as data-less would steer the agent into a
+ * priced search that misses. This is what makes discovery-first a one-step
+ * pipeline: confirm coverage → run next_call verbatim → series in hand.
+ */
+export function buildNextCall(matches: CoverageMatch[]): NextCall | null {
   const m = matches.find((x) => !x.unavailable && x.coverage.names.length > 0);
   if (!m) return null;
-  const first = m.coverage.names[0];
-  return m.coverage.kind === "entities" ? `${first} ${m.name}` : `${m.name} ${first}`;
+  const first = m.coverage.names[0] as string;
```
robertabbott, line 245: `next_call` picks `coverage.names[0]` — after the `orderMetricNames` partition, that is whichever metric the backend's popularity order happened to put first, with no relation to what the caller was after (`input.q` names the entity, not the metric). `buildSummary` then tells the model to run the ready-made `next_call` verbatim.

For `q="Capital One"` with no `coverage_filter`, that spends a priced search on an arbitrary metric — the opposite of this PR's thesis. It is well-targeted *with* a `coverage_filter`. Would gating `next_call` on a `coverage_filter` being present (or on a short/single-name coverage list) keep the handle useful without inviting an off-target priced call?

robertabbott (approved):

PR #40, opened 2026-04-22
TAKO-2596: add GitHub Actions CI workflow for workers/

jfassio:
Phase 0 of the [Tako MCP](https://linear.app/trytako/project/tako-mcp-635ac07fae22) Cloudflare Workers port. Closes [TAKO-2596](https://linear.app/trytako/issue/TAKO-2596).

**Stacked on #39 (TAKO-2595).** Targets the 2595 branch to keep the diff scoped to the two CI files. GitHub will retarget this to `main` automatically once #39 merges.

## Summary

- `.github/workflows/workers-ci.yml` — triggers on `pull_request` with `paths: ['workers/**']` + `workflow_dispatch`. Steps: checkout → setup-node (LTS, npm cache keyed on `workers/package-lock.json`) → `npm ci` → `npm run typecheck` → `npm test`. Job-level `defaults.run.working-directory: workers`. Concurrency with `cancel-in-progress: true`. Least-privilege `permissions: contents: read`.
- `workers/README.md` — 15-line stub with workflow status badge, Linear project link, and a Local dev section.
- Python-only PRs (touching `src/tako_mcp/**`) do NOT trigger this workflow.

## Test plan

- [x] YAML parses cleanly
- [x] `npm run typecheck` + `npm test` still pass locally after these additions (2595's scaffold untouched)
- [ ] First PR that touches `workers/**` (this one, since it modifies `workers/README.md`) should trigger the workflow — verify on the Actions tab once this opens
- [ ] Confirm a Python-only PR does NOT trigger the workflow (can't fully verify without a second test PR — `paths: ['workers/**']` is documented GitHub Actions behavior)
- [ ] Runs in under 3 minutes on empty test suite — expected (local pipeline runs in ~8s); observe on first run

## Review-time decisions to confirm

1. **Node pinning.** Used `node-version: 'lts/*'` (auto-tracks Node LTS). Pin to `'22'` if you want reproducibility.
2. **No `push: main` trigger.** Post-merge runs on `main` won't happen — badge reflects PR runs only. Add `push: { branches: [main], paths: ['workers/**'] }` if you want main health on the badge.
3. **Concurrency group.** `workers-ci-\${{ github.workflow }}-\${{ github.ref }}` is slightly redundant (workflow name is already `workers-ci`). Cosmetic only — safe to simplify.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #181, opened 2026-07-28
fix(widget): probe iframe capability instead of sniffing window.openai

25eliu:
## Summary

The static chart widget decided interactive-iframe vs static-PNG by **host identity**: `window.openai` defined → iframe (ChatGPT), otherwise PNG. That permanently pins Claude to the static image. claude.ai's sandbox currently hardcodes `frame-src 'self' blob: data:` and ignores declared `csp.frameDomains` ([anthropics/claude-ai-mcp#40](https://github.com/anthropics/claude-ai-mcp/issues/40), open, pending their MCP Apps security review) — but once Anthropic ships that fix, the host sniff would *still* route Claude to the PNG until we redeployed.

This PR replaces the sniff with **capability detection**:

- **ChatGPT** (`window.openai` present): commits to the interactive iframe immediately — behavior unchanged.
- **Every other host** (claude.ai, VS Code, Goose, …): paints the PNG first (today's guaranteed baseline), then probes the embed iframe hidden in the background:
  - **Success** = iframe `load` event → swap the iframe in over the PNG, at the PNG's rendered height (claude.ai sizes its outer container once per [anthropics/claude-ai-mcp#69](https://github.com/anthropics/claude-ai-mcp/issues/69); a taller swap would clip).
  - **Failure** = `frame-src` `securitypolicyviolation` (fires ~instantly on blocking hosts, so today's Claude UX is unchanged) or 8 s timeout → stay on the PNG and unload the probe frame so a blocked-content tile can never surface.

Net effect: the interactive chart lights up on any host the moment its CSP allows it — **no Tako redeploy, no widget-URI change** (Claude desktop caches widget URIs beyond connector lifecycle, so the URI must stay stable).

Also updates the now-stale comments in `mcp.ts` / `tako_search.ts` that described the old host-sniff behavior.

## Test plan

- [x] 5 new jsdom-executed widget tests (`test/widget/widget-dom.test.ts`): probe starts hidden behind the image; upgrade on load; stay-on-image on CSP violation (and a late `load` cannot swap after settling); ChatGPT path skips the probe; no probe without `embed_url`
- [x] Full suite: 533 + 13 tests pass, all 4 typecheck projects pass
- [ ] Staging smoke: chart renders unchanged on claude.ai (PNG baseline, probe fails fast via CSP violation)
- [ ] Staging smoke: chart still interactive on ChatGPT (committed iframe path untouched)

### DIFF
--- a/workers/src/tools/_chart_widget.ts
+++ b/workers/src/tools/_chart_widget.ts
@@ -359,6 +359,27 @@ const WIDGET_HTML = `<!doctype html>
   // embed page, not arbitrary cross-frame senders.
   var embedOrigin = null;
 
+  // Pin the embed/probe iframe to an explicit pixel height. Hosts vary in
+  // which of the three they read (inline \`style.height\`, the
+  // \`min-height\` floor, and the \`height\` attribute), so set all three
+  // together. One helper so the next height gotcha is a one-line fix
+  // instead of the four separate inlined call sites this replaces.
+  function setFrameHeight(n) {
+    frame.style.height = n + "px";
+    frame.style.minHeight = n + "px";
+    frame.setAttribute("height", String(n));
+  }
+
+  // Record the embed iframe's origin so the \`tako-embed-height\` resize
+  // handler will honor messages from it. Only arm this once the embed is
+  // the VISIBLE surface (an immediate iframe render, or a probe that has
+  // upgraded) — arming it while a probe frame is still hidden behind the
+  // PNG would let a background embed resize the widget under a chart the
+  // user is looking at.
+  function armEmbedOrigin(url) {
+    try { embedOrigin = new URL(url).origin; } catch (e) { embedOrigin = null; }
+  }
+
   // Pick the rendering mode by CAPABILITY, not host identity.
   //
   // ChatGPT's Apps SDK runtime exposes \`window.openai\`; its outer
@@ -394,6 +415,11 @@ const WIDGET_HTML = `<!doctype html>
   // tool-result delivery racing the flag can't double-assign
   // \`frame.src\`.
   var probeStarted = false;
+  // Flips true once the probe has swapped the interactive iframe in over
+  // the PNG. The image \`load\`/\`error\` handlers check it and no-op after
+  // an upgrade so a stray duplicate image event can't re-reveal the PNG
+  // on top of the live iframe.
+  var probeUpgraded = false;
 
   // Try to load \`url\` in the (still hidden) chart iframe and swap it
   // in over the already-rendered PNG if it genuinely loads.
@@ -417,6 +443,15 @@ const WIDGET_HTML = `<!doctype html>
     probeStarted = true;
     var settled = false;
     var timer = null;
+    // Flips true the instant we navigate the probe frame to \`url\`, and
+    // back to false before \`fail()\` sends it to about:blank. \`onLoad\`
+    // gates on this rather than \`frame.src === url\`: \`frame.src\` reads
+    // back the browser-NORMALIZED URL (\`https://tako.com:443\` collapses
+    // to \`https://tako.com\`, the host lowercases), so a raw-string
+    // compare against \`url\` can silently never match — stranding every
+    // render on the 8 s timeout on any host whose \`embed_url\` origin
+    // isn't already normalized.
+    var probeNavigated = false;
     function cleanup() {
       settled = true;
       if (timer !== null) clearTimeout(timer);
@@ -427,24 +462,28 @@ const WIDGET_HTML = `<!doctype html>
       if (settled) return;
       cleanup();
       embedOrigin = null;
+      probeNavigated = false;
       frame.src = "about:blank";
       log("iframe probe failed, staying on image", { reason: reason });
     }
     function succeed() {
       if (settled) return;
       cleanup();
+      probeUpgraded = true;
       // Swap at the image's CURRENT rendered height, not the tool's
       // requested height: claude.ai sizes its outer container once
       // (anthropics/claude-ai-mcp#69) from the PNG's footprint, so
-      // growing the frame past it would clip. A same-height swap is
-      // seamless; the \`tako-embed-height\` handshake below can still
-      // resize later on hosts that honor size-changed.
+      // growing the frame past it would clip. The probe only starts
+      // AFTER the image has loaded (see the image \`load\` handler), so
+      // \`getBoundingClientRect().height\` here is the real laid-out
+      // height — not 0 falling through to \`fallbackHeight\`.
       var rectH = image.getBoundingClientRect().height;
       var offsetH = image.offsetHeight;
       var h = Math.round(rectH || offsetH || 0) || fallbackHeight;
-      frame.style.height = h + "px";
-      frame.style.minHeight = h + "px";
-      frame.setAttribute("height", String(h));
+      setFrameHeight(h);
+      // Arm the embed-height handshake only NOW: the embed has become the
+      // visible surface, so honoring its resize messages is finally safe.
+      armEmbedOrigin(url);
       imageLink.classList.add("hidden");
       placeholder.classList.add("hidden");
       frame.classList.remove("hidden");
@@ -459,14 +498,27 @@ const WIDGET_HTML = `<!doctype html>
       }
     }
     function onLoad() {
-      // \`fail()\` navigates the frame to about:blank, which fires its
-      // own \`load\` — only count a load of the probe URL itself.
-      if (frame.src === url) succeed();
+      // \`fail()\` navigates the frame to about:blank, which fires its own
+      // \`load\`; \`probeNavigated\` is false by then so we don't count it.
+      //
+      // KNOWN GAP (bare-load success signal): \`load\` fires for ANY
+      // document the embed endpoint returns — including a Tako 404/5xx
+      // error page. On a host whose CSP allows \`frame-src\`, an embed-page
+      // outage would swap the known-good PNG for an error page. Accepted
+      // for now: the only positive signal from the cross-origin embed is
+      // the \`tako-embed-height\` message, which the Tako web app does not
+      // emit yet (see the handler comment near the bottom). Once it does,
+      // gate \`succeed()\` on that message instead of the bare \`load\`.
+      if (probeNavigated) succeed();
     }
     document.addEventListener("securitypolicyviolation", onViolation);
     frame.addEventListener("load", onLoad);
-    try { embedOrigin = new URL(url).origin; } catch (e) { embedOrigin = null; }
+    // Deliberately do NOT arm \`embedOrigin\` here — the probe frame is
+    // still hidden behind the PNG, so honoring its \`tako-embed-height\`
+    // messages would resize the widget under a chart the user is looking
+    // at. \`succeed()\` arms it after the swap.
     frame.src = url;
+    probeNavigated = true;
     timer = setTimeout(function () { fail("timeout"); }, 8000);
     log("iframe probe started", { src: url });
   }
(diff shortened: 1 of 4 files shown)


workers/src/tools/_chart_widget.ts
```diff
@@ -359,28 +359,118 @@ const WIDGET_HTML = `<!doctype html>
...
+      log("iframe probe succeeded, upgraded to interactive", { height: h });
+    }
+    function onViolation(event) {
+      var directive =
+        event && (event.effectiveDirective || event.violatedDirective);
+      if (typeof directive === "string" && directive.indexOf("frame-src") === 0) {
+        fail("csp:" + directive);
+      }
+    }
+    function onLoad() {
```
robertabbott, line 500: The comment above says `load` "only fires when a document actually loaded", which is true, but it fires for *any* document the host let through — including Tako's own 404 or a 5xx error page from `/embed/{pub_id}/`. The PNG is proof only that `/api/v1/image/{pub_id}/` rendered (it's fetched server-side and validated down to the IHDR chunk in `fetchImageDataUrlAndDims`); it says nothing about the embed page. So on a host whose CSP allows `frame-src`, an embed-page outage swaps a known-good chart PNG out for an error page. Confirmed under the branch's jsdom harness: any `load` on the probe frame hides the image link and reveals the iframe, with no check on what loaded.

Should the success gate require a positive signal from the embed page itself (the `tako-embed-height` message the origin-gated handler at line 920 already accepts) rather than the bare `load` event?

workers/src/tools/_chart_widget.ts
```diff
@@ -359,28 +359,118 @@ const WIDGET_HTML = `<!doctype html>
...
+      cleanup();
+      // Swap at the image's CURRENT rendered height, not the tool's
+      // requested height: claude.ai sizes its outer container once
+      // (anthropics/claude-ai-mcp#69) from the PNG's footprint, so
+      // growing the frame past it would clip. A same-height swap is
+      // seamless; the \`tako-embed-height\` handshake below can still
+      // resize later on hosts that honor size-changed.
+      var rectH = image.getBoundingClientRect().height;
+      var offsetH = image.offsetHeight;
+      var h = Math.round(rectH || offsetH || 0) || fallbackHeight;
```
robertabbott, line 482: Nothing sequences the probe against the image. `render()` assigns `image.src` and then starts the probe in the same tick, so if the frame's `load` lands first, two things follow — both reproduced under the branch's own jsdom harness:

- `image`'s `load` handler (line 626) unconditionally runs `imageLink.classList.remove("hidden")`, so the PNG comes back *on top of* the already-visible iframe — measured `frameHidden=false, linkHidden=false`, i.e. the chart rendered twice.
- `rectH` and `offsetH` are both 0 at that point, so `h` falls through to `fallbackHeight` — measured 720 px for a `height: 720` payload, exactly the "grew past the PNG's footprint, claude.ai clips it" case the comment two lines up warns against.

Reachable without an exotic host: when `fetchImageDataUrlAndDims` bails (PNG over `MAX_INLINE_DATA_URL_BYTES`, fetch timeout, non-PNG content type) the widget falls back to the cross-origin `image_url`, whose network fetch races the embed page's load.

Should the probe only be allowed to settle after the image has laid out, or should the image `load` / `error` handlers no-op once the probe has upgraded?

workers/src/tools/_chart_widget.ts
```diff
@@ -359,28 +359,118 @@ const WIDGET_HTML = `<!doctype html>
...
+      var directive =
+        event && (event.effectiveDirective || event.violatedDirective);
+      if (typeof directive === "string" && directive.indexOf("frame-src") === 0) {
+        fail("csp:" + directive);
+      }
+    }
+    function onLoad() {
+      // \`fail()\` navigates the frame to about:blank, which fires its
+      // own \`load\` — only count a load of the probe URL itself.
+      if (frame.src === url) succeed();
```
robertabbott: `frame.src` reads back the *resolved* URL while `url` is the raw string from `structuredContent`, and `validatePublicOrigin` in `env.ts` returns its input verbatim (it only rejects trailing slashes and non-http schemes). So a `PUBLIC_BASE_URL` of `https://tako.com:443` or `https://TAKO.com` passes validation and produces an `embed_url` that never equals `frame.src` — verified in jsdom: `https://tako.com:443/embed/a/?dark_mode=auto` reads back as `https://tako.com/embed/a/?dark_mode=auto`, `equal: false`. The failure is silent: `succeed()` never fires, every render waits out the 8 s timeout and stays on the PNG.

Would capturing `frame.src` after assignment (or a `probeUrlLoaded` flag flipped before the `about:blank` navigation) be more robust than comparing against the raw string?

workers/src/tools/_chart_widget.ts
```diff
@@ -359,28 +359,118 @@ const WIDGET_HTML = `<!doctype html>
...
+    }
+    document.addEventListener("securitypolicyviolation", onViolation);
+    frame.addEventListener("load", onLoad);
+    try { embedOrigin = new URL(url).origin; } catch (e) { embedOrigin = null; }
```
robertabbott: Before this change the image path never set `embedOrigin`, so the `tako-embed-height` branch at line 920 was unreachable there. Now it is armed for the whole probe window. The comment at line 869 says the Tako web app does not emit that message yet — but the day it does, a hidden probe frame will resize itself and call `notifyHeight(n)`, telling the host the embed page's height while the user is looking at the PNG. Given widget bundles are effectively permanent once shipped (see the `APP_UI_RESOURCE_URI` docstring), should `embedOrigin` only be set in `succeed()`, after the swap actually happens?

workers/src/tools/_chart_widget.ts
```diff
@@ -359,28 +359,118 @@ const WIDGET_HTML = `<!doctype html>
...
+      // (anthropics/claude-ai-mcp#69) from the PNG's footprint, so
+      // growing the frame past it would clip. A same-height swap is
+      // seamless; the \`tako-embed-height\` handshake below can still
+      // resize later on hosts that honor size-changed.
+      var rectH = image.getBoundingClientRect().height;
+      var offsetH = image.offsetHeight;
+      var h = Math.round(rectH || offsetH || 0) || fallbackHeight;
+      frame.style.height = h + "px";
+      frame.style.minHeight = h + "px";
+      frame.setAttribute("height", String(h));
```
robertabbott: `frame.style.height` / `style.minHeight` / `setAttribute("height", …)` now appears at lines 445-447, 590-592, 706-708, and 924-926, and `try { embedOrigin = new URL(url).origin } catch` at 468, 589, and 705. Would a `setFrameHeight(n)` (plus an `armEmbedOrigin(url)`) inside the IIFE be worth it, so the next height gotcha gets fixed in one place instead of four?

workers/src/tools/_chart_widget.ts
```diff
@@ -593,6 +683,13 @@ const WIDGET_HTML = `<!doctype html>
       // Triggers the load event above. Set last so the listener is
       // attached first.
       image.src = imageSrc;
+      // Capability probe: the PNG above is the guaranteed baseline; if
+      // this host's CSP lets the embed iframe load (it doesn't on
+      // claude.ai until anthropics/claude-ai-mcp#40 is fixed), swap in
+      // the interactive chart once it does.
+      if (validEmbed) {
+        probeInteractiveIframe(url, h);
```
robertabbott: The capability signal lives in the browser and the `extraMeta` gate lives in the worker, so on every non-ChatGPT host we keep paying the full chart-render fetch (up to `PNG_FETCH_TIMEOUT_MS`, 8 s) for a PNG that gets thrown away seconds later once the probe upgrades — plus a visible PNG-to-iframe reflow on each render. Is that the accepted steady state after Anthropic ships #40, or is the plan to re-gate `extraMeta` on a client identifier once claude.ai is known-good?

workers/src/tools/_chart_widget.ts
25eliu: Fixed in 1144e5f. The probe now starts from **inside the image `load`/`error` handlers** instead of at `render()` time, so the image is always laid out before the probe can `succeed()`. That closes both failure modes you reproduced:

- **Double render:** the image `load` handler has already run by the time the probe can swap, so it can't re-reveal the PNG on top of the iframe. Added a `probeUpgraded` guard at the top of the `load` handler as belt-and-suspenders against a stray duplicate load.
- **Wrong height:** `succeed()`'s `getBoundingClientRect().height` is now measured against the laid-out image, not 0-falling-through-to-`fallbackHeight`.

New jsdom tests: probe not armed until the PNG loads, and no re-reveal on a duplicate image load after upgrade.

workers/src/tools/_chart_widget.ts
25eliu: Fixed in 1144e5f. Replaced the `frame.src === url` compare in `onLoad` with a `probeNavigated` boolean, flipped true right after `frame.src = url` and back to false before the `about:blank` navigation in `fail()` (your suggested flag). It no longer depends on `frame.src` reading back byte-identical to the raw `url`, so `:443`/mixed-case origins that normalize on read-back can't silently strand the render on the 8 s timeout.

workers/src/tools/_chart_widget.ts
25eliu: Fixed in 1144e5f. `embedOrigin` is no longer armed at probe-start — it's armed inside `succeed()`, after the swap. During the probe window it stays `null`, so the `tako-embed-height` handler drops resize messages from the still-hidden probe frame. Extracted this into an `armEmbedOrigin(url)` helper and added a test asserting an embed-height message is ignored while the probe is hidden and honored only after the upgrade.

workers/src/tools/_chart_widget.ts
25eliu: Done in 1144e5f. Extracted `setFrameHeight(n)` (sets `style.height` + `style.minHeight` + the `height` attribute together) and `armEmbedOrigin(url)` into the IIFE. All four height sites (succeed, useIframe, no-image fallback, the `tako-embed-height` handler) and the three `embedOrigin` try/catch blocks now go through them — so the next height gotcha is a one-line fix, and the C3/C4 fixes landed cleanly in one place.

workers/src/tools/_chart_widget.ts
25eliu: Good catch — real gap. I've **documented it rather than gated on it** for now, because the only positive signal from the cross-origin embed is the `tako-embed-height` message, and the Tako web app doesn't emit it yet (the handler at the bottom is armed for when it ships). Gating `succeed()` on it today would disable the interactive upgrade entirely — including on claude.ai once #40 lands, which is the whole point of the probe.

So in 1144e5f the `onLoad` comment now spells out the bare-load risk (a 404/5xx embed page still fires `load`, so on a `frame-src`-allowing host an embed outage swaps the good PNG for an error page) and the plan: gate on `tako-embed-height` instead of the bare `load` once the embed page emits it. Worth noting the exposure is narrow today — the only `frame-src`-allowing host (ChatGPT) takes the immediate-iframe path, not the probe; the probe only runs on hosts that currently block frames.

workers/src/tools/_chart_widget.ts
25eliu: Accepted steady state, not a planned re-gate. The capability signal is deliberately client-side (in the widget), and re-gating `extraMeta` on a client identifier is exactly the host-identity sniffing this PR removes — it'd put us back to 'keep Claude on the PNG even after Anthropic ships #40.' The PNG is the always-works baseline every non-`window.openai` host renders; the probe upgrades opportunistically on top of it with no redeploy.

The cost you're describing is real (full `PNG_FETCH_TIMEOUT_MS` fetch + a PNG→iframe reflow once #40 lands on claude.ai), but the fetch is already skipped for ChatGPT via the `extraMeta` gate, and skipping it for claude.ai would require it to first be *known-good* interactive — which it isn't until #40. Once it is, optimizing the fetch away is a clean follow-up, but I'd keep that a deliberate future call rather than bake a client-id gate in now. No code change here.

robertabbott (approved):

PR #183, opened 2026-07-28
feat(mcp): ChatGPT OAuth tool discovery — top-level securitySchemes + auth challenges

25eliu:
## Problem

ChatGPT never showed the sign-in / link-account UI for the Tako connector, and the anonymous tool surface (3 tools) contradicted the 5-tool app submission. Two root causes, both verified against OpenAI's docs (developers.openai.com/apps-sdk/build/auth) and live production:

1. **ChatGPT reads a top-level `securitySchemes` array on each tool descriptor** — but the MCP SDK (verified on 1.29.0 and latest 1.30.0) drops unknown top-level fields in `registerTool` and rebuilds `tools/list` descriptors from a fixed field list. Only `_meta["com.tako/securitySchemes"]` survived to the wire, which ChatGPT ignores.
2. **Nothing ever triggered the auth UI**: anonymous requests succeed on the free tier (no HTTP 401), `tako_contents`/`tako_visualize` were hidden from the anonymous listing, and no tool call ever returned the `_meta["mcp/www_authenticate"]` challenge OpenAI documents as required.

## Fix (all gated on `client === "chatgpt"` — other clients see zero change)

- **`workers/src/tools/_security.ts`** (new): per-tool scheme map derived from `FREE_TIER_TOOL_NAMES` (`noauth`+`oauth2` for search/answer/available_data, `oauth2`-only for contents/visualize), the shared `WWW-Authenticate` formatter, the auth-challenge tool result, and a pure `tools/list` transform.
- **Post-serialization injection**: `withChatGptToolSecuritySchemes` in `mcp.ts` rewrites the buffered `tools/list` JSON for ChatGPT UAs only (the SDK physically can't carry the field; `enableJsonResponse: true` guarantees the response is fully buffered).
- **Anonymous discoverability with a hard execution gate**: anonymous ChatGPT connections now LIST all five submitted tools (required for the link-account UI), but `tako_contents`/`tako_visualize` are blocked at dispatch with an `_meta["mcp/www_authenticate"]` challenge — they never execute on the shared free-tier account. Invariant: User-Agent can widen the anonymous *listing* (to the fixed submitted set), nothing can widen anonymous *execution*.
- **Reauth on stale tokens**: Django 401s on authenticated ChatGPT connections carry the `invalid_token` challenge so ChatGPT offers re-linking. Free tier is deliberately excluded (there the rejected credential is the shared key — "your session expired" would be false and would mask a key outage).
- HTTP `WWW-Authenticate` behavior is byte-identical; OAuth validation untouched; `chatgpt-app-submission.json` untouched; parity check (`registry:check`) still enforces the anonymous surface ⊆ submitted tools.

## Review

Reviewed pre-push by a structured critical pass + 4 specialist subagents + an adversarial subagent that ran live probes against the real worker (anonymous execution attempts with valid args, UA spoofing, `?tools=` widening, batch smuggling — all blocked with zero backend calls). Review fixes landed as the third commit: free-tier 401 challenge gating, adapter failure logging, `requestOrigin` decoupled from `iconsBaseUrl`, header-injection hardening, one-source tier threading, immutability fix, docstring corrections.

## Tests

- 570/570 pass (`npm run typecheck`, `npm test`, `npm run registry:check`), 13 added in review.
- New coverage: ChatGPT 5-tool listing with per-tool top-level `securitySchemes` (e2e HTTP, authenticated AND anonymous); both gated tools blocked anonymously with schema-valid args and zero fetches; authenticated execution without challenge; free-tier 401 unchallenged; adapter passthrough for SSE/malformed JSON/non-list bodies; header-injection guards; transform edge cases; visualize stays `openWorldHint: true` / retrieval read-only for ChatGPT.

## Remaining manual verification

- [ ] Full ChatGPT OAuth login → token exchange → authenticated `tools/list` → live `tako_contents` / `tako_visualize` calls
- [ ] Confirm ChatGPT renders the sign-in prompt from the anonymous challenge
- [ ] Confirm the real ChatGPT connector User-Agent against `wrangler tail` (tests use the `"ChatGPT/1.0"` stand-in; `detectMcpClient` matches chatgpt/openai/gptbot/oai-searchbot substrings)
- [ ] Note in CHANGELOG: `_meta["com.tako/securitySchemes"]` values are now per-tool (free tools also declare `noauth`)


workers/src/mcp.ts
```diff
@@ -762,11 +806,31 @@ function registerTool(
...
+      // `tier` is stamped from the registration-time value — the SAME one
+      // that drove `isToolOnSurface` and the dispatch gate below — so the
+      // per-call context can never disagree with the surface decision
+      // (previously `ctx.tier` was a second, caller-supplied source of
+      // truth that had to be kept in sync by hand).
       const callCtx: ToolContext = {
         ...ctx,
         sendProgress,
         client: options.client,
+        tier: options.tier,
```
robertabbott, line 882: This makes `options.tier` the single authority and stamps it into `callCtx.tier`, overriding `ctx.tier` — but `createMcpServer` still resolves it with `const tier = options.tier ?? "authenticated"` (line 337), while `ctx.tier` is populated by `handleMcpRequest` and documented in `types.ts:97`, and is now silently discarded.

Since this PR moves the execution boundary off the listing and onto the dispatch gate below, the gate's sole input defaults to the *permissive* value. Concretely: `createMcpServer({token: SHARED_KEY, tier: "free", client: "chatgpt", ...}, {client: "chatgpt"})` — tier passed only via `ctx` — registers all five tools, passes the gate, and runs `tool.handler` against `ctx.token`. Not reachable today, since `handleMcpRequest` is the only call site and sets both.

Should the default derive from `ctx.tier` (fail closed), or assert the two agree, rather than defaulting to `"authenticated"`? Collapsing to one source removed the redundancy that would have caught a future second call site, which matters more now that this gate is the only execution barrier.

workers/src/tools/_security.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Security schemes for a tool, derived from the free-tier surface: tools
+ * anonymous connections may EXECUTE (`FREE_TIER_TOOL_NAMES`) advertise
+ * `noauth` + `oauth2` (anonymous works, linking unlocks the caller's own
+ * account); everything else is `oauth2`-only. Deriving from the same set
+ * that gates execution keeps the advertised schemes and the enforced
+ * behavior from drifting — metadata is never what AUTHORIZES a call, the
+ * server-side tier/token checks are.
+ */
+export function securitySchemesForTool(name: string): SecurityScheme[] {
```
robertabbott: This derives the advertised schemes from a static name set with no tier or env input. Two consequences — one I confirmed, one open.

**Confirmed:** `workers/README.md:226` documents `wrangler secret delete FREE_TIER_API_KEY --env production` as a kill switch that "instantly reverts the environment to the pre-free-tier behavior (anonymous → 401), no code change or redeploy needed." After this PR that rollback is partial: `resolveFreeTierConfig` returns null and anonymous requests hard-401 at `mcp.ts:1181`, but the descriptors keep telling ChatGPT `noauth` works. Same mismatch on `wrangler dev` and any deployed `tako-mcp-dev` where the secret is unset. `createMcpServer` already has `ctx.env` — should this take `resolveFreeTierConfig(env) !== null` so the kill switch stays a complete kill switch?

**Open question I could not settle:** `noauth` is also advertised on *authenticated* connections. The file's doc comment says deriving from `FREE_TIER_TOOL_NAMES` keeps the advertised schemes and the enforced behavior from drifting, but the two do disagree here. If ChatGPT acts on per-tool `noauth` by preferring an unauthenticated call for those three tools, a linked user's `tako_search`/`tako_answer` traffic lands on the shared `FREE_TIER_API_KEY` and its shared throttles instead of their own account. Is that worth threading tier in so `noauth` only appears on the anonymous listing? Your "authenticated `tools/list` → live calls" manual-verification item would catch it if it is real.

workers/src/mcp.ts
```diff
@@ -422,16 +463,19 @@ function registerTool(
...
-  // read it to know the tool is reachable via OAuth and needs the `mcp` scope.
+  // SEP-1488 (still Draft) will standardize at the descriptor top level.
+  // ChatGPT reads the TOP-LEVEL `securitySchemes` field instead — injected
+  // post-serialization by `withChatGptToolSecuritySchemes` in
+  // `handleMcpRequest`, since the SDK can't carry it — from the same
+  // `securitySchemesForTool` source, so the two never disagree.
   // The widget block below MERGES into this rather than replacing it.
   config._meta = {
-    "com.tako/securitySchemes": [{ type: "oauth2", scopes: ["mcp"] }],
+    "com.tako/securitySchemes": securitySchemesForTool(tool.name),
```
robertabbott: The description says the fix is "all gated on `client === \"chatgpt\"` — other clients see zero change", but this line is unconditional. claude.ai, Claude Code, Cursor and the unknown long tail now receive `[{type:"noauth"},{type:"oauth2",scopes:["mcp"]}]` on `tako_search` / `tako_answer` / `tako_available_data` where they previously received `[{type:"oauth2",scopes:["mcp"]}]`, including on fully authenticated connections. The last checklist item half-acknowledges this.

The comment above still says hosts read this key "to know the tool is reachable via OAuth and needs the `mcp` scope", which no longer describes the `noauth` entry. Was widening this to non-ChatGPT hosts intended? If only ChatGPT needs it, why not scope the value on `options.client` the way the top-level injection at line 1303 is, and keep every other client on the previous constant?

robertabbott (commented):
Reviewed the auth-boundary move specifically: since listing is no longer the security boundary, I went looking for an anonymous-execution bypass and did not find one. `tier` is single-sourced from `handleMcpRequest`, the dispatch gate sits above `tool.handler` in the only callback that reaches it, `resources/read` on the widget URI carries no credential, and `?tools=` widening plus UA spoofing plus batch smuggling are all blocked. Mutation-tested the tests too: deleting the dispatch gate fails 3 tests, and dropping the `client === "chatgpt"` guard in `isToolOnSurface` fails 3 more, so the `expect(fetchMock).not.toHaveBeenCalled()` assertions are load-bearing rather than incidental.

Inline questions cover the gate's default input, the tier/env-independence of the advertised `noauth`, and the parity guard.

One finding has no diff line to attach to — the docs this change invalidates:

`workers/README.md:210` is the **rollout verification step**: "connect an MCP client to `mcp.staging.tako.com/mcp` with no auth. Expect `tools/list` to show exactly the three tools." That is now User-Agent-dependent — a curl with `-A ChatGPT/1.0` shows five — so the step an operator runs to decide the rollout succeeded can now read as a failure, or mask a real one. Should it say which UA to send and what count to expect for each?

Same staleness in the surrounding prose, none of it covered by `registry:check`:

- `workers/README.md:109` — "Anonymous connections see exactly three tools: `tako_available_data`, `tako_search`, `tako_answer`. Everything else is hidden."
- `workers/README.md:114` — "a `tools/call` for a hidden tool returns \"tool not found\""; on ChatGPT those two tools are no longer hidden and now return `authRequiredToolResult`.
- `workers/README.md:266` — the operator warning still assumes OAuth hosts decide sign-in from a 401, which is exactly what this PR routes around for ChatGPT. The resulting asymmetry (ChatGPT gets a per-tool sign-in affordance, claude.ai still lands silently on the anonymous tier) is the kind of thing the next reader cannot recover from the code.
- `README.md:33`, `:50`, `:257`, `llms.txt:14`, `llms-full.txt:16` — all still describe the anonymous surface as three tools.

workers/scripts/gen-registry.ts
```diff
@@ -146,10 +146,12 @@ export function assertLlmsFullCoverage(
...
  *
  * The anonymous free-tier surface is asserted separately (and more
- * weakly): it must be a SUBSET of the declared tools. The reverse gap is
- * expected — `tako_contents` and `tako_visualize` are absent for
- * anonymous connections — which is why the submission's test cases
- * assume an OAuth-linked connection.
+ * weakly): it must be a SUBSET of the declared tools. Since the ChatGPT
+ * link-account flow requires auth-only tools to stay listed
+ * (`CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES` in `tools/_surface.ts`),
+ * the anonymous ChatGPT listing now matches the declared five — but
```
robertabbott: This now states as fact that "the anonymous ChatGPT listing now matches the declared five", but the check below is still one-directional: `isToolOnSurface(name, "chatgpt", ∅, "free") ⇒ declaredNames.has(name)`. The anonymous listing *shrinking* is never an error, so deleting a name from `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES` — the exact regression this PR exists to prevent — still passes `registry:check`.

Should this become an equality assertion (free ChatGPT surface == declared tools) so the guard enforces what the comment claims?

workers/src/mcp.ts
```diff
@@ -1198,7 +1294,14 @@ export async function handleMcpRequest(
...
       const response = await transport.handleRequest(request);
       await logSdkValidationRejections(requestForLogging, response);
-      return response;
+      // ChatGPT-only compatibility adapter: rewrite the buffered
+      // `tools/list` response to carry the top-level `securitySchemes`
+      // field its Apps SDK reads (the MCP SDK cannot serialize unknown
+      // descriptor fields — see `tools/_security.ts`). Every other
+      // client gets the SDK's response untouched.
+      return client === "chatgpt"
+        ? await withChatGptToolSecuritySchemes(response)
```
robertabbott: Both halves of this fix — the injection here and the anonymous widening in `isToolOnSurface` — hinge on `detectMcpClient`'s User-Agent substring match, and the checklist item "confirm the real ChatGPT connector User-Agent against `wrangler tail`" is still open. If the production connector UA misses all four substrings (`chatgpt`, `openai`, `gptbot`, `oai-searchbot`) it resolves to `unknown`, neither the injection nor the anonymous listing fires, and the deploy silently reverts to precisely the behavior this PR fixes.

The only log line added is `securitySchemes injection failed` in the catch below, which fires on a parse error and never on a UA miss. Worth logging the resolved client on the ChatGPT `tools/list` path so a UA change is visible in `wrangler tail` rather than showing up as "the sign-in prompt stopped appearing again"?

workers/src/mcp.ts
25eliu: Fixed in d7cb2ae — both suggestions, cheaply: `createMcpServer` now resolves `options.tier ?? ctx.tier ?? "authenticated"`, so a call site that declares the tier only on `ToolContext` engages the gate instead of silently getting the permissive default; and if both are provided and disagree it throws (`createMcpServer: options.tier (...) disagrees with ctx.tier (...)`) rather than picking a side — the redundancy you noted is back, as a loud assertion instead of two hand-synced reads. Regression test added: a server built with `ctx.tier: "free"` only (no `options.tier`) still blocks `tako_contents` with the challenge and zero Django calls.

workers/src/tools/_security.ts
25eliu: Both halves addressed in d7cb2ae by making the derivation per-connection: `securitySchemesForTool(name, {client, tier})` advertises `noauth` only on an ANONYMOUS ChatGPT listing for the three free tools.

**Kill switch (confirmed issue):** complete again, with no env plumbing needed — with `FREE_TIER_API_KEY` deleted, anonymous requests 401 before any listing exists, so `tier: "free"` (and with it `noauth`) can never be served. Verified on `wrangler dev` (which has no secret): anonymous + ChatGPT UA → 401 + `WWW-Authenticate`, no descriptors.

**Open question (authenticated `noauth`):** resolved in the direction you leaned — authenticated listings are now `oauth2`-only for every tool, so ChatGPT is never told a linked user's calls could go anonymous onto the shared key. Rationale: `noauth` only matters pre-link, and pre-link connections are exactly the anonymous ones, so scoping it there loses nothing. e2e tests assert both listings (authenticated all-`oauth2` in index.test.ts, anonymous `noauth`+`oauth2` on the free three in freetier.test.ts).

workers/src/mcp.ts
25eliu: Not intended — fixed in d7cb2ae. The same per-connection scoping that resolves the thread above restores "zero change for other clients" as a true statement: `_meta["com.tako/securitySchemes"]` is now derived with `{client, tier}`, so claude.ai / Claude Code / Cursor / unknown clients get exactly the previous constant `[{type:"oauth2",scopes:["mcp"]}]` on every tool and every tier; only anonymous ChatGPT listings see `noauth`. The stale comment above the assignment is updated too. index.test.ts pins the non-ChatGPT constant again.

workers/scripts/gen-registry.ts
25eliu: Fixed in d7cb2ae — the check is now an equality assertion: the anonymous ChatGPT surface must equal the declared tools, with a distinct error message per direction (`…is declared in the submission but missing from the anonymous free-tier ChatGPT surface (link-account UI needs it listed pre-auth)`). Mutation-tested: temporarily deleting `tako_contents` from `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES` fails `registry:check` naming the tool; restoring passes. The stale "asserted as a SUBSET" doc comment is updated to match.

workers/src/mcp.ts
25eliu: Fixed in d7cb2ae — `withChatGptToolSecuritySchemes` now logs `[mcp] tools/list securitySchemes injected client=chatgpt tier=<tier>` whenever the transform actually fires. That makes it a positive rollout signal: present in `wrangler tail` = UA resolved to chatgpt AND injection ran; absent while sign-in misbehaves = points straight at `detectMcpClient`. The rollout-verification step in workers/README.md now tells the operator to look for this exact line. (Confirming the real connector UA in production logs stays on the PR's manual checklist — this makes that check greppable.)

robertabbott (approved):

workers/src/tools/_surface.ts
```diff
@@ -61,13 +61,39 @@ export const CHATGPT_DEFAULT_ON_TOOL_NAMES: ReadonlySet<string> = new Set([
...
+ * listing can never surface the sign-in UI, so the two submitted tools
+ * that need a linked account must still be listed before authentication
+ * (developers.openai.com/apps-sdk/build/auth). They are listed, NOT
+ * runnable: `registerTool` in `mcp.ts` gates the call at dispatch time
+ * and answers with an `_meta["mcp/www_authenticate"]` challenge instead
+ * of executing on the shared free-tier account. Every other client keeps
+ * the original policy (hidden entirely — a listed tool that errors on
+ * call wastes agent turns on hosts with no linking UI to show).
+ *
+ * Must stay a subset of the tools `chatgpt-app-submission.json` declares —
```
robertabbott: This still says the set "Must stay a subset of the tools `chatgpt-app-submission.json` declares". After d7cb2ae the guard is an equality in both directions: I deleted `tako_contents` from this set and `registry:check` failed with `tool "tako_contents" is declared in the submission but missing from the anonymous free-tier ChatGPT surface`.

The `gen-registry.ts` docstring was updated for that; this one — the doc whoever edits this set actually reads — still describes shrinking as allowed, which is exactly the action the new guard rejects. Should it say that this set plus `FREE_TIER_TOOL_NAMES` must EQUAL the declared tools?

workers/src/mcp.ts
```diff
@@ -213,9 +232,13 @@ export function createMcpServer(
...
      * Connection tier. `"free"` (anonymous, no Authorization header)
-     * restricts the registered toolset to `FREE_TIER_TOOL_NAMES`.
-     * Omitted → `"authenticated"`, the full surface — what every
-     * existing caller and test gets.
+     * restricts the EXECUTABLE toolset to `FREE_TIER_TOOL_NAMES` — on
+     * ChatGPT the auth-required submitted tools stay LISTED for the
+     * link-account UI but are blocked at dispatch (see
+     * `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES` in `tools/_surface.ts`
+     * and the free-tier gate in `registerTool`). Omitted →
+     * `"authenticated"`, the full surface — what every existing caller
```
robertabbott: `Omitted → "authenticated", the full surface — what every existing caller and test gets` no longer holds after the resolution change below: an omitted option now falls back to `ctx.tier` first, and the new regression test in `mcp.test.ts` depends on precisely that (a ctx-only `"free"` engages the dispatch gate).

Same staleness in `workers/src/tools/types.ts:93`, which says `ctx.tier` is "Currently consumed only by the Django error mapping in `mcp.ts`" — it is now an input to the only execution barrier for auth-required tools. Should both state the resolution order, given the `types.ts` doc currently tells a reader that `ctx.tier` only affects error copy?

workers/scripts/gen-registry.ts
```diff
@@ -190,16 +193,29 @@ export function assertChatgptSubmissionParity(
...
+      problems.push(
+        `tool "${name}" is on the anonymous free-tier ChatGPT surface but not declared in the submission`,
+      );
+    }
+  }
+  for (const name of declaredNames) {
+    if (!freeChatgptSurface.has(name)) {
       problems.push(
-        `tool "${t.name}" is on the anonymous free-tier ChatGPT surface but not declared in the submission`,
+        `tool "${name}" is declared in the submission but missing from the anonymous free-tier ChatGPT surface (link-account UI needs it listed pre-auth)`,
```
robertabbott: When I triggered this branch by shrinking `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES`, the thrown error ended with the footer at line 243: `Update chatgpt-app-submission.json so each tool's annotations equal toolAnnotationsForClient(tool, "chatgpt") for the default ChatGPT surface`. For this direction the fix is restoring the name in `_surface.ts`, not editing the submission JSON — the remediation line points at the wrong file for the failure that just fired.

Related transitive consequence, which I could not find stated anywhere: combined with the pre-existing `declared == authenticated ChatGPT surface` assertion above, this equality makes the anonymous and authenticated ChatGPT surfaces necessarily identical. A future ChatGPT tool that *should* stay hidden pre-auth cannot exist without loosening this check, and the message would steer whoever hits it into listing it anonymously instead. Should the footer name the file per failure direction, and the docstring record the surfaces-are-identical consequence?

robertabbott (commented):
Delta re-review of d7cb2ae against the 878d7f14 review. All five prior inline findings plus the doc staleness in the review body are resolved; the three below are documentation accuracy only, nothing blocking.

Verified independently rather than taking the commit message for it:

- `typecheck` clean, `npx vitest run src` 566 passed, `registry:check` ok (12 tools).
- Mutation-tested all four new guards; every one is load-bearing. Reverting the tier fallback to `options.tier ?? "authenticated"` fails the new ctx-tier-only gate test. Deleting `tako_contents` from `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES` fails `registry:check` naming the tool. Dropping `tier === "free"` from `securitySchemesForTool` fails 3 tests (unit plus both authenticated e2e listings); dropping `client === "chatgpt"` fails the non-ChatGPT constant.
- The kill-switch claim holds on its own reading: `tier = "free"` is set only where `freeTier !== null`, so with `FREE_TIER_API_KEY` deleted, `noauth` cannot be served at any layer.
- The new disagreement throw is unreachable in production — both `createMcpServer` tier inputs come from the single `tier` binding in `handleMcpRequest`.

workers/src/tools/_surface.ts
25eliu: Fixed in 49bdf51 — the docstring now states the real invariant: this set plus `FREE_TIER_TOOL_NAMES` must EQUAL the declared submission tools, both directions enforced, with removal explicitly called out as a `registry:check` failure (dropping a submitted tool's link-account affordance).

workers/src/mcp.ts
25eliu: Fixed in 49bdf51, both spots: the `createMcpServer` option doc now states the resolution order (`options.tier ?? ctx.tier ?? "authenticated"`) and why the ctx fallback is deliberate fail-closed behavior; `types.ts` reorders `ctx.tier`'s doc to lead with its role as a SECURITY INPUT to the surface + dispatch gate (including the disagreement throw), demoting the Django error-copy use to second — and notes that inside a tool call, handlers see the registration-time tier stamped over the field.

workers/scripts/gen-registry.ts
25eliu: Both parts fixed in 49bdf51.

**Remediation direction:** the shrunken-surface problem line now carries its own fix (`— fix by restoring the name in CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES / FREE_TIER_TOOL_NAMES (workers/src/tools/_surface.ts, workers/src/freetier.ts), NOT by editing the submission`), and the footer is direction-neutral. Re-verified by re-running your mutation: deleting `tako_contents` now produces an error naming `_surface.ts`.

**Surfaces-are-identical consequence:** now recorded in the `assertChatgptSubmissionParity` docstring as deliberate — every submitted ChatGPT tool is visible pre-auth because the link-account UI needs it — together with the escape hatch for a future hidden-pre-auth tool (relax the anonymous side back to ⊆ plus an explicit intentionally-hidden allowlist), so whoever hits the error is steered away from silencing it by listing the tool anonymously.

PR #187, opened 2026-07-29
fix: correct the research skills against live results; batch contents URLs; fix the structuredContent envelope

25eliu:
Grounded in ~60 live tool calls against production, plus a routing-framing pass and two rounds of review-driven follow-ups (§7-§9, then §10-§13 below addressing a full line-by-line review). Rebased onto current `main` after #182/#183 landed.

---

## 1. `fix(skills)` — correct the three research skills against live results

**The `sources` default was backwards.** The skills told agents to pass `["data"]`, calling web results "clutter". Verified: `"OpenAI annualized revenue"` and `"Nvidia next earnings date"` return zero usable data cards while the web results carry the answer, and web never degraded card selection on a well-formed query.

**The flagship comparison example does not work.** `"Intel vs Nvidia revenue"` reproducibly (3/3) returns two single-entity cards, not the two-series card the skill promised. `"Coca-Cola vs PepsiCo revenue"` returns a real one.

**The web-traffic brand-query warning had the failure backwards.** `"Netflix traffic"` returns Netflix *subscriber* cards on data — plausible numbers that are not traffic — and CDN/BGP articles on web, not the documented zero cards.

**The macro skill instructed citing from a fixed "FRED / OECD / BIS" list.** Live cards also come from BLS, IMF, World Bank, Census and Polymarket.

Per-skill anecdotes are replaced by one "Pick the right card" checklist (Overview cards outranking charts, segment variants crowding out company-wide metrics, level-vs-rate, projections, entity ambiguity), and each skill documents coverage the old text talked agents out of: private companies, stock quotes, crypto, demographics, app MAU, long-tail domains. README copies are regenerated from `skills/` and verified byte-for-byte.

## 2. `fix(render)` — render freshness and free-tier relevance

Two facts were silently dropped from **every real card**. `freshness` never rendered because the guard was `typeof rec.data_freshness === "string"` but the backend sends `{"data_as_of": …}` — verified against a captured production card. It is the only reliable actual-vs-projection and stale-vintage signal, so two of the skill checks above were unrunnable without it. `relevance` never rendered for unentitled callers, who carry the coarse string rather than the gated numeric score.

## 3. `fix(available_data)` — remove `coverage_filter`

**Correction (review):** this was originally written up as "never released." That is wrong. `coverage_filter` shipped in **v0.15.0** (published 2026-07-29T01:47Z) and appears in that release's `registry/server.json`, the discovery card `publish-mcp.yml` pushes to the public MCP registry; staging has served it since #179 merged. The v0.15.0 *production* deploy failed, so prod likely never served it, and Zod strips unknown keys, so a caller still pinned to the old schema gets a silently unfiltered result rather than a 400 — small blast radius, but it is a released input being removed, and per AGENTS.md that wants a **minor** bump pre-1.0, not the patch a `fix:` title implies.

It cost more calls than it saved: after four clean "no coverage matching" verdicts the model re-called unfiltered anyway (4 × ~8k chars), because a bare negative gave it nothing to trust. Removed rather than patched; the truncation note now steers callers to treat an unseen name as **unconfirmed** rather than absent.

⚠️ **Known gap accepted with this (narrowed below, see §9):** the filter was the only way to reach past the coverage cap. Centene reports `250+` metrics with `truncated: true`, and no shown name contains "membership" — so `Centene Total membership - Medicare` (which exists, and `tako_search` finds) was no longer discoverable via `tako_available_data`. Deliberate call, recorded here; §9 raises the cap.

**Correction (review):** this note originally said "50 shown." That number was wrong. Measured against prod (anonymous `tako_available_data q=Centene`, v0.15.0, `MAX_COVERAGE_NAMES = 200`): `ent::centene_corporation::f41593cd` returns **~200 names** (201 comma-separated fields, at least two of which are fragments of metric names that themselves contain commas — `Net PP&E, Excl. Operating Leases` and `EBT, Incl. Unusual Items (Normalized)`). Exactly the cap, not 50. The "membership" finding holds: still absent from all ~200.

## 4. `feat(contents)` — batch up to 10 URLs per call

One URL per call meant eight filings cost eight round trips, and every extra round trip re-sends the whole conversation as input tokens — quadratic in the transcript, not just latency. The backend still takes one URL per request, so the batch fans out to N concurrent subrequests. Same bill, far fewer waves.

`urls` (1–10) is the documented input; `url` stays accepted as a legacy alias — not enforced as mutually exclusive with `urls` (if a caller sends both, `urls` wins silently rather than 400-ing a request we can serve). Output is `{ results[], cost }` with **positional** entries — `results[i]` describes `urls[i]` including failures, which carry `error` instead of a payload, because compacting them out would silently re-map every later index onto the wrong URL. `allSettled` contains partial failure so one gated 403 no longer discards the pages that resolved; when *every* URL fails the first error is re-thrown so the single-URL path keeps its existing self-correcting envelope.

## 5. `fix(mcp)` — carry the payload in structuredContent

A client reading `structuredContent` — the spec-natural choice when a tool advertises an `outputSchema` — got `{request_id, usage, guidance}` and silently performed worse than one reading the text block. Prod ships cards there today, so this regresses on the next deploy.

**Inverted rather than mirrored:** `structuredContent` carries the full payload; the text channel becomes a readable index (titles, headline descriptions, facts line, one-line row pointer) instead of a second copy. Mirroring both would restore the doubling the markdown work removed. Advertised shapes now declare `cards`/`web_results`/`answer`, still loosely typed to avoid reintroducing the wire-drift failure `_search_results.ts` documents.

**Which half is a restore and which is new (review):** v0.14.0's `mcp.ts` set `result.structuredContent = output` and `tako_search` had no `renderText`, so every release up to v0.14.0 shipped the full payload in *both* channels. Refilling `structuredContent` is therefore a **return to released behavior** — no `structuredContent`-reading client can regress. Making the text channel **pointer-only is the new bet**: no release has ever served it, and `mcp.ts:1002` names the long-tail hosts (Cursor, Windsurf, Gemini CLI, LibreChat) whose forwarding behavior is unverified here. Claude and ChatGPT are covered; one smoke check on a non-Claude host would retire the rest cheaply and is the recommended follow-up.

Side effect: dropping snippets from the text channel strictly strengthens upstream-content isolation — a snippet forging Tako's own section framing is no longer echoed there, so that test asserts absence rather than correct fencing.

## 6. `refactor(routing)` — make tako_answer the entry point, tako_search the specialist

The tools read as peers, and peers get misrouted. Under current payloads retrieval structurally cannot finish a question: a card's rows are usually license-gated (captions without values) and a web result is a snippet (a pointer without facts), so every search obligates a follow-up wave, and each wave re-sends the whole conversation. `tako_answer` is the only tool whose single priced response can terminate the loop.

Framing only, no gating, nothing removed: `tako_answer` now leads with "START HERE for any question that wants a value"; `tako_search` is reframed around what it's genuinely best at (breadth fan-out, a chart card when the embed is the deliverable, harvesting node ids/urls) and states plainly that it locates data and does not carry values; `tako_available_data` is positioned as the coverage question, explicitly not a warm-up before every lookup. The three bundled skills, their README copies, and `llms-full.txt` follow the same routing. A follow-up commit cut the resulting prose's em-dash overuse (87 → 35 across the three skills) — no meaning changed, README copies regenerated and verified byte-for-byte.

## 7. `fix(render)` — tako_contents was still doubling the payload

Code review caught that commit 4's batching work regressed the exact thing commit 5 fixed. Search/answer cards correctly moved rows into `structuredContent` with a one-line pointer left in the text; `tako_contents` put the payload in `structuredContent` too, but **also** kept fencing the full page text into the text channel. Every call shipped the payload twice — up to 10 urls × a 100k-char inline cap — on any host that counts both channels toward context. The doc comment even cited a rationale ("see the note on renderContentsText") that was never actually written.

Fixed by extending the same pattern used for cards: `renderContentsText` now emits a `payloadPointer()` naming the payload and its size (`"page text: 41302 chars in structuredContent.results[].data"`), matching `rowsPointer()`. The payload rides exactly once, in `structuredContent`.

Also cleaned up while in there:
- Deleted the dead row-table renderers (`renderContentRows`, `markdownTable`, `cell`) orphaned by commit 5's `rowsPointer()` switch — no references remained, and `tsc` has no `noUnusedLocals` / there is no lint script, so they'd have rotted silently.
- Dropped a vestigial `void fenced;` and a stale comment in `renderWebResult` that still described snippet-fencing after commit 5 stopped rendering snippets at all.
- Corrected the `tako_contents` comment that claimed `url`/`urls` mutual exclusion the schema never enforced; it now documents the actual behavior (`urls` wins if both are sent).

The regression's only guard had been a test whose *title* said "slims structuredContent to metadata only (payload channels dropped)" while its *body* asserted the opposite (the full payload present). Retitled and split so each half is asserted explicitly: payload present in `structuredContent`, absent from the text.

## 8. `fix(tools)` / `docs(llms-full)` — every surface describing the envelope caught up

`tako_search` and `tako_answer`'s `DESCRIPTION` still told the calling model that the markdown "Tako Data"/cited-cards section includes the cards' "recent rows". That was accurate before commit 5's `rowsPointer()` refactor, when `renderCard` rendered a full row table inline — but not after: the markdown now carries only a rows-count pointer, and the actual rows live in `structuredContent.cards[].content`. Left uncorrected, a model reading only the tool description would have no signal that it needs to look at `structuredContent` at all. Reworded both descriptions to name where the rows (and, for search, the web snippets) actually ride; regenerated `registry/*.json` off the edit.

`llms-full.txt` carried the identical stale claim for `tako_answer` ("cited cards inline their recent rows alongside it — the values arrive with the prose"). Corrected the same way, so every surface — tool description, registry, llms-full — now describes the same envelope.

## 9. `fix(available_data)` — raise `MAX_COVERAGE_NAMES` to match the server's own cap (closes §3's known gap)

Follow-up on §3's coverage_filter removal: the 200-name ceiling was set deliberately *below* the server's own counting cap (`graph/related` stops counting at 250, `total_capped: true`) to save one page round-trip. That made it a second, tighter cap layered on top of the server's, and it is what produced §3's known gap — Centene's `250+` metrics landing capped at 200 behind normalized-accounting noise, with `Total membership - Medicare` past the cut and no `coverage_filter` left to reach it.

Raised `MAX_COVERAGE_NAMES` from 200 to 250 so it matches, not undercuts, the server's cap: any node at or under 250 now gets its **complete** coverage list. The pagination loop already scales off this constant, so no loop change was needed — it costs one extra sequential page in the worst case (`ceil(250/100) = 3` vs `2`), well inside the existing `MAX_COVERAGE_PAGES = 4` slack.

**Verified, not assumed (review):** that `ceil(250/100)` arithmetic holds only if the server actually pages at `PAGE_LIMIT = 100`, which had never been measured — and review correctly noted that if the real page size were small (~12–25, as "50 shown" would imply), `MAX_COVERAGE_PAGES = 4` would be the binding cap and this constant could not fix anything. Measured on prod: Centene returns ~200 names, which is unreachable in 4 pages at any page size below 50 and requires **at least two full 100-item pages**. So `PAGE_LIMIT = 100` is the real page size, `MAX_COVERAGE_NAMES` was genuinely the binding constraint at 200, and 3 pages is the correct worst case for 250. `MAX_COVERAGE_PAGES` is not the lever here.

⚠️ **What this does NOT prove:** that 250 closes §3's gap. Prod still runs the 200 cap, so the ~50 names this bump newly exposes can't be inspected from prod, and `Total membership - Medicare` is absent from all ~200 currently returned. Whether it falls inside 200–250 or past the server's own 250 counting cap (in which case the lever is server-side, not this constant) needs one call against staging once this lands. §3's gap should be read as *narrowed and measurable*, not closed.

The cost side of review's note stands as an accepted trade: this does add a third sequential page and ~25% more names to a tool advertised as "Free and fast" and positioned as the recommended first call. Taken deliberately — a coverage list that silently undercuts the server's own ceiling is the worse failure for a tool whose entire job is answering "is this covered?". The truncation note is unchanged and still fires exactly when it should: `truncated` stays driven by the server's own `total_capped` flag (plus `total > names.length`), so it now only shows for nodes genuinely past the server's 250 cap, not for the 200–250 range this commit newly exposes in full.

---

## 10. `fix(contents)` — review: error-gate leak, batch character budget, batching docs

A full line-by-line review of §4-§9 caught four more issues confined to `tako_contents.ts`:

- **`errorText` gated on `DjangoHttpError` specifically**, but `DjangoNotFoundError` / `DjangoUnauthorizedError` / `DjangoBadRequestError` / `DjangoTimeoutError` / `DjangoResponseParseError` are *sibling* subclasses of the base `DjangoError`, not subclasses of `DjangoHttpError`. In a partial-batch failure (the single-URL path is unaffected — it re-throws whole instead of going through `errorText`), this silently dropped a 404's self-correcting `modelGuidance` and leaked the internal request path into model-visible text for everything else — a Safety Rules violation. Gated on `DjangoError` instead, with a `"body" in reason` guard since not every subclass carries one.
- **No batch-wide character budget.** `max_chars` defaulted to 100k *per URL*, and that default exists specifically because the backend's own default (up to 1M chars, ~250k tokens) was unreadable for any client — but a 10-URL batch defaulting to 100k each reconstructs exactly that blowup, spread across urls instead of concentrated in one. Added `BATCH_CHAR_BUDGET` (200,000, split evenly across the batch) for the silent default only; an explicit caller `max_chars` is left untouched at any batch size.
- The tool's own docstring/`DESCRIPTION` still said "One URL per call" — stale since §4, contradicting the `urls` param's own description two lines down. `llms-full.txt`'s entry had the identical defect.
- Closed test gaps: neither `url` nor `urls` given, both given together (`urls` wins, now asserted not just documented), and a >1-result `renderContentsText` case (no test reached the header/numbered-section/failed-entry formatting).

⚠️ ~~**Judgment call, flagged not fully resolved:** `BATCH_CHAR_BUDGET = 200,000` is a round starting number, not a measured one.~~ **Addressed in §14** — bumped to 250,000 and instrumented so the number becomes measurable instead of argued.

## 11. `fix(render)` — review: stale docs, dead code, a rowsPointer bug, and re-fixing the test that didn't get fixed

Five more findings in `_render_markdown.ts`, all left behind by the §7/§9 payloadPointer/rowsPointer refactors:

- `renderContentsText`'s own docstring, plus an orphaned duplicate stacked above `payloadPointer`, both still claimed the function fences the raw payload into text — the exact opposite of what it does now. Corrected the live one, deleted the orphan.
- A stale comment inside `renderContentsItem` explaining *why* page text is fenced survived directly above the comment explaining why it *isn't* fenced at all anymore. Deleted the stale half.
- `WIDGET_KEYS` was dead: `slimSearchStructured` became a full-object spread in §5 and stopped reading it.
- `rowsPointer`'s `shown` computation chained `??`, but an empty array's `.length` is `0` — not `undefined` — so `0 ?? nextBranch` doesn't fall through (nullish coalescing only treats `null`/`undefined` as empty). A card with `dataset.rows: []` but populated `records` silently lost its pointer line entirely. Switched to `||`.
- **§7 claimed the structuredContent-duplication test was "retitled and split so each half is asserted explicitly." It was neither** — the title still said "slims structuredContent to metadata only (payload channels dropped)" while the body asserted the payload *is* present. This is the exact title/body contradiction §7 identifies as the original regression's failed guard, reintroduced by the commit that claimed to fix it. Split for real this time into two tests, one per guarantee.

## 12. `fix(tools)` — review: soften tako_search's "does not deliver values" overstatement

"It locates data; it does not deliver values" and "a card's rows are usually license-gated" both overstated the common case: `include_contents` defaults true, and an `exportable: true` card (the majority) inlines a free 20-row preview by default. Reworded to say search *does* deliver a preview for exportable cards, a license-gated card carries none at all, and a web result is a snippet — the "`tako_answer` is faster for a plain value" routing point survives, now grounded accurately.

⚠️ **Judgment call, deliberately not resolved:** review suggested naming the actually-gated sources ("S&P Global, Visible Alpha, FactSet, CoinMarketCap...") in place of "usually license-gated." Did not do that — can't verify that list is accurate/current from the codebase, and a wrong or stale vendor list in a model-facing description is worse than the vaguer-but-safe wording landed here. **Decision (confirmed): do not name them, and not as a follow-up edit to this string either.** The reason is stronger than "unverified": a stale vendor name in a model-facing description is *worse* than vague wording, because the model pattern-matches on it and mis-routes in both directions — skipping a `tako_contents` call on a card it wrongly assumes is gated, or trusting a card as free because the list aged out. The authoritative signal already ships per-result and cannot drift: every card carries `exportable`, and gated cards carry `values_hint`, which is exactly why `_search_results.ts`'s `slimCard` stamps a **positive** "no" rather than leaving the key missing. The description's job is to set the prior ("don't assume every card has rows"), not to duplicate ground truth that arrives with the data. If vendor-level intuition is ever wanted up front, it should be generated from the backend's licensing config with drift-checking, not hand-typed here.

⚠️ **Separately flagged, not closed:** §6 (the `tako_answer`/`tako_search` routing repositioning) is still in the PR body without live call-count evidence, unlike every other claim here. That data needs a live MCP client against prod, the same way §1's skill fixes are backed by specific queries and results.

**Recommendation: don't block on it.** §6 is framed in its own commit message as framing only — no gating, no tiered exposure, nothing removed, the model still chooses — so it is a reversible prompt-engineering change, not a behavioral gate, and warrants a lower evidentiary bar than the correctness fixes here. The protocol to close it when someone has a connected session: 8–10 representative "get me a value" queries across financial / macro / web-traffic plus a couple of generic ones; run each down both paths (old: `tako_search` → follow-up; new: `tako_answer` direct); record round trips, total tokens/cost, and whether the correct figure actually came back; report the aggregate delta in the same shape as §1 and §3.

## 13. `docs` — review: remaining stale wording + a missing blank line, synced across every surface

`llms-full.txt`'s `tako_search` paragraph, and both tools' trailing "machine essentials ride separately in structuredContent" lines, still described the pre-§5/§7 shape (rows in the markdown; structuredContent framed as metadata-only). `README.md`'s `tako_answer`/`tako_contents` table rows had the same two defects. All corrected. Separately, all three skill files (and their manually-synced `README.md` copies) lost the blank line before `## Output` in §6's em-dash cleanup pass — heading sat flush against the last `## Examples` bullet. Fixed in all six places.

---

## 14. `fix(contents)` / `chore` — review round 3: the batch budget becomes measurable, and the merge blocker

**`BATCH_CHAR_BUDGET` 200,000 → 250,000, and instrumented.** §10 landed 200,000 explicitly flagged as guessed. Rather than argue the number, it is now (a) chosen on a stated principle and (b) observable.

The principle: a batch call should cost about the same *total* context as one large single-URL call, split thinner across more urls — batching buys fewer round trips, not a proportionally larger content budget. 250,000 over 200,000 because 2-URL batches then land at 125k/url, above the 100k single-URL default, so they stay byte-for-byte unaffected; a 10-URL batch gets 25k chars/page (~6k tokens) instead of 20k, a real cushion for slightly-longer-than-average pages, and the worst case is still ~62.5k tokens total, well under a 128k-token context floor. Deliberately **not** sublinear (`budget / sqrt(N)`): more code for uncertain benefit, and flat division is what the review itself suggested. The three escape hatches for "I need one big document" are untouched — fetch that url alone (full 100k), pass `query` (passages are trimmed by `extractPassages` regardless, so the split never applies), or set `max_chars` explicitly (bypasses the budget entirely, by design).

The instrumentation: `fetchOne` now emits a `console.warn` (`[tako] tako_contents batch max_chars cap bit …`, tagged with `batch_size` and `per_url_cap`) exactly when the **derived** default cut a page — batching drove the per-url cap below the single-URL 100k default *and* the page was long enough to hit it. An explicit caller `max_chars` getting cut is normal and does not log. So `wrangler tail --search "batch max_chars cap bit"` answers "is 250,000 right" from production truncation rates within days, instead of waiting for someone to notice a clipped batch result. Tests derive their expectations from the imported constant rather than hardcoding it, so the next tuning pass doesn't strand stale numbers, and both the fires and does-not-fire cases are covered.

**Merge blocker cleared: `workers/node_modules` is no longer committed.** It was a mode-120000 symlink to the author's home directory, added in `1bb4be0` and carried by every commit after it, so a squash-merge would have landed it on `main` and broken `npm ci` on a fresh clone. Root cause is exactly as diagnosed in review: `.gitignore` had `node_modules/` **with a trailing slash**, which matches directories only, and git stores a symlink as a blob rather than a tree — so the pattern never applied. `git rm --cached workers/node_modules` plus dropping the slash, with a comment on the pattern recording why, since the obvious "tidy" edit is to put the slash back. The symlink stays on disk, so worktrees linking to the main checkout's installed deps keep working — the full suite still passes through exactly that symlink. Audited the rest of the index: no other tracked symlinks.

**The other named blocker was already fixed in §10** — `errorText` gates on the `DjangoError` base class, not the `DjangoHttpError` sibling, so 404 `modelGuidance` survives and the internal request path no longer leaks into model-visible text on a partial-batch failure.

---

## Testing

4/4 typecheck passes, **630 worker + 16 widget tests** (up from 619 — §10/§11 add coverage for the error-gate fix, the batch-budget split, the >1-result render path, and the properly-split structuredContent test; §14 adds the two cap-logging cases), `schemas:check` and `registry:check` clean. Verified on the final merged branch state, not just per-commit. The render-fix tests in §7 were checked against a reverted fix — exactly the two bug-catching ones fail, so they are not vacuous.


workers/src/tools/tako_contents.ts
```diff
@@ -155,8 +174,200 @@ const outputSchema = z.object({
...
 
+/** Fetch ONE url. Throws on failure so the caller can decide whether a single
+ *  URL's error degrades to an entry or fails the whole batch. */
+type Input = z.infer<typeof inputSchema>;
+
+/** One URL's failure, as text the model can act on. Prefers the
+ *  self-correcting `modelGuidance` the 403/404 paths attach, so a gated card
+ *  inside a batch still explains itself instead of surfacing a bare status. */
+function errorText(reason: unknown): string {
+  if (reason instanceof DjangoHttpError) {
```
noahjax: **[Issue]** This gate matches 403 and nothing else. `DjangoNotFoundError`, `DjangoBadRequestError`, `DjangoUnauthorizedError`, `DjangoTimeoutError` and `DjangoResponseParseError` are *siblings* of `DjangoHttpError`, not subclasses — see `django.ts:97-189`, where only `DjangoHttpError` is the catch-all. Two consequences, both confined to a partial-batch failure, which is the path this PR adds:

1. **A 404's `modelGuidance` is discarded.** `fetchOne` builds it 60 lines up, then `errorText` falls straight through to `reason.message`, so the model reads `Django returned 404 for POST /api/v1/contents/` instead of "Check the URL came from a search/answer result verbatim...". That is exactly the retry loop the guidance exists to break, and every retry is billed. This function's own docstring promises the opposite: "Prefers the self-correcting `modelGuidance` the 403/404 paths attach".
2. **401/400/timeout/parse-error leak the internal path** into model-visible text as `Django returned <status> for POST /api/v1/contents/`. AGENTS.md Safety Rules: "Error responses must never expose internal URLs or stack traces."

Single-URL calls are unaffected, since all-fail re-throws into `djangoErrorToToolResult`. The new test covers 403 — the one status that already works.

Fix is one word, plus adding `DjangoError` to the import at :24. `modelGuidance` and `status` both live on the base class; `body` does not, so it needs a guard:

```ts
if (reason instanceof DjangoError) {
  if (reason.modelGuidance !== undefined) return reason.modelGuidance;
  const detail = "body" in reason ? extractErrorDetail(reason.body as string) : undefined;
  const status = reason.status !== undefined ? String(reason.status) : "transport error";
  return `Fetch failed (${status}${detail !== undefined ? `: ${detail}` : ""}).`;
}
return reason instanceof Error ? reason.message : String(reason);
```

Keeping the trailing `reason.message` is right — the wire-guard throws plain `Error`s whose text is already model-safe. Worth a test per non-403 status.

workers/src/tools/tako_contents.ts
```diff
@@ -47,10 +47,29 @@ const DESCRIPTION = [
 // silently join the MCP input surface), then re-describe them for the MCP
 // surface — including the one documented divergence (default mode → inline),
 // the `max_rows` row cap, and the `content_format` serialization.
+/** Max URLs per call. The backend takes one URL per request, so a batch
+ *  fans out to N subrequests; this bounds both the Workers subrequest budget
+ *  and the bill a single call can run up. */
+export const MAX_CONTENTS_URLS = 10;
```
noahjax, line 56: **[Issue]** No batch-wide character budget. `max_chars` defaults to 100k **per URL**, and the comment at :98 justifies that default precisely because the backend's 1M default is "up to 1M chars = 250k tokens - observed in the wild and unreadable for any client". A 10-URL batch reconstructs exactly the condition that default was added to prevent: 10 x 100k = 1M chars in a single tool result, roughly 250k tokens.

The round-trip argument in section 4 is right and the cap of 10 is not the problem — the fix belongs on the character budget:

```ts
const perUrlCap = Math.floor(BATCH_CHAR_BUDGET / targets.length); // BATCH_CHAR_BUDGET ~ 200_000
// in fetchOne: maxCharsAsked ?? (inline ? Math.min(100_000, perUrlCap) : undefined)
```

An explicit caller `max_chars` should still cap *down* per URL, so no lever is lost. Card URLs are already bounded by `max_rows <= 2000`, so this is a web-text-only blowup.

workers/src/tools/tako_contents.ts
```diff
@@ -28,9 +28,9 @@ import { extractPassages } from "./_passages.js";
...
   renderContentsText,
   slimContentsStructured,
-  type ContentsOutputLike,
+  type ContentsBatchLike,
 } from "./_render_markdown.js";
-import type { ToolModule } from "./types.js";
+import type { ToolContext, ToolModule } from "./types.js";
 
 const DESCRIPTION = [
   "Fetch the real content behind one result URL from tako_search or tako_answer — the rows behind a Tako card, or a web page's full text.",
```
noahjax: **[Suggestion]** `DESCRIPTION` line 1 still reads "behind **one** result URL", and the file docstring at :12 still says "One URL per call". This is the same defect class section 7 fixed for `tako_search`/`tako_answer`, inside the very tool section 4 batched. The `urls` param description at :57 tells the model to batch, so the two channels contradict each other and the model has to pick. `llms-full.txt:66` opens with the same sentence before saying "Takes 1-10 URLs per call" two clauses later.

workers/src/tools/tako_contents.ts
```diff
@@ -175,164 +386,53 @@ const takoContents = {
...
-      typeof dataText === "string"
-    ) {
-      const extracted = extractPassages(dataText, passageQuery);
-      dataText = extracted.data;
-      note = extracted.note;
-      cut = cut || extracted.truncated;
+    // Fan out: the backend takes ONE url per request, so a batch is N
+    // subrequests issued concurrently. allSettled, not all — one URL's 403
+    // (a license-gated card) must not discard the pages that did resolve.
+    const settled = await Promise.allSettled(
```
noahjax, line 488: **[Suggestion]** Test gaps on the new path:

- neither `url` nor `urls` — the runtime throw at :390 is the only enforcement of a constraint the schema cannot express, and nothing covers it
- both sent — `urls` wins, documented at :62 as a deliberate choice, unasserted
- `renderContentsText` with more than one result — all the new header / `###` section / failed-entry formatting lives there and no test reaches it

Also: the single-item render test in `_render_markdown.test.ts` omits `url`, which the handler always sets on every entry, so its `md.startsWith("> 2 match(es)")` assertion passes against a shape production never emits.

workers/src/tools/_render_markdown.ts
```diff
@@ -663,3 +699,17 @@ export function renderContentsText(o: ContentsOutputLike): string {
 
   return blocks.join("\n\n");
 }
+
+/** tako_contents as text: one section per requested url, each with the payload
+ *  itself (page text raw, csv fenced, json payloads fenced). This is where the
```
noahjax: **[Issue]** This docstring says the function emits "each with the payload itself (page text raw, csv fenced, json payloads fenced). This is where the JSON-escaping tax on 100k-char pages actually dies" — the exact opposite of what it now does. The payload rides in `structuredContent` and this function emits a pointer.

Section 6 opens by faulting a doc comment that "cited a rationale that was never actually written." This is the same defect, reintroduced by the commit that fixed it. It also matters more than a normal stale comment: the escaping-tax rationale is now inverted, because the payload moved *into* the JSON channel.

workers/src/tools/_render_markdown.ts
```diff
@@ -617,23 +630,46 @@ export interface ContentsOutputLike {
...
+
+/** structuredContent for tako_contents: the full batch INCLUDING each item's
+ *  payload. This is the ONLY copy — `renderContentsText` emits a pointer, not
+ *  a duplicate, for the same reason search cards get `rowsPointer()`. */
+export function slimContentsStructured(o: ContentsBatchLike): Record<string, unknown> {
+  return { ...(o as unknown as Record<string, unknown>) };
 }
 
 /** tako_contents as text: the payload itself (page text raw, csv fenced,
  *  json payloads fenced), led by the passage note and trailed by a one-line
```
noahjax: **[Issue]** The old `renderContentsText` docstring was orphaned here when `payloadPointer` was inserted between it and its function. Lines 647-650 and 651-652 are now two stacked doc comments describing different functions, and a reader hovering `payloadPointer` gets the wrong one first. Delete 647-650 (the accurate replacement is at :703, once corrected).

workers/src/tools/_render_markdown.ts
```diff
@@ -647,13 +683,13 @@ export function renderContentsText(o: ContentsOutputLike): string {
...
   // back. Card csv/json get the same treatment (a cell containing ``` would
   // otherwise close the fence early).
-  if (o.data !== undefined) {
-    blocks.push(fenced(o.data, o.format === "csv" ? "csv" : "text"));
-  } else if (o.records !== undefined) {
-    blocks.push(fenced(JSON.stringify(o.records), "json"));
-  } else if (o.dataset !== undefined) {
-    blocks.push(fenced(JSON.stringify(o.dataset), "json"));
-  }
+  // The payload rides in structuredContent, NOT here. Emitting it in both
```
noahjax, line 674: **[Nit]** The stale comment survived directly above its own correction: 681-685 explains why page text is fenced, 686-690 explains why the payload is not in this channel at all. Delete 681-685.

workers/src/tools/_render_markdown.ts
```diff
@@ -82,36 +98,31 @@ const WIDGET_KEYS = [
...
+ * This is the MCP-spec-natural channel — a client that reads structuredContent
+ * (the obvious choice when a tool advertises an outputSchema) must find the
+ * cards there, not an empty envelope. Slimming this side to metadata made
+ * those clients silently perform worse than ones reading the text block.
+ *
+ * The payload is carried EXACTLY ONCE: `renderText` is a compact index
+ * (titles, headline values, pointers) rather than a second copy of the rows,
+ * so hosts that count both channels are not billed twice.
+ */
 export function slimSearchStructured(o: SearchOutput): Record<string, unknown> {
```
noahjax, line 104: **[Nit]** `WIDGET_KEYS` (:92) is dead now that this function is a spread — nothing else reads it. Same rot section 6 cites for deleting `renderContentRows` / `markdownTable` / `cell`: `tsc` has no `noUnusedLocals` and there is no lint script, so it sits there silently.

workers/src/tools/_render_markdown.ts
```diff
@@ -148,49 +152,21 @@ type LooseContent = {
...
-  }
-
-  if (typeof c.data === "string" && c.data.trim() !== "") {
-    const lines = c.data.split("\n").filter((l) => l !== "");
-    const shown = Math.max(0, lines.length - 1); // minus header line
-    return `${totalNote(shown)}\n${fenced(c.data.trim(), "csv")}`;
-  }
-
-  return undefined;
+  const shown =
```
noahjax, line 157: **[Nit]** A card with `dataset.rows: []` but populated `records` now yields no pointer at all. `Array.isArray(c.dataset?.rows) ? c.dataset.rows.length : undefined` returns `0`, and `0 ?? x` is `0`, so the `??` chain never reaches `records` and the `shown === 0` guard below returns `undefined`. `renderContentRows` fell through to `records` here (its guard was `rows.length > 0`). Edge case, but the fix is `|| undefined` on the first term.

workers/src/tools/tako_search.ts
```diff
@@ -44,21 +44,21 @@ import {
 import type { AppUiResource, ToolContentBlock, ToolModule } from "./types.js";
 
 const DESCRIPTION = [
-  "Fastest, direct retrieval of data or metrics from the live web and proprietary data sources, returned as structured cards and web links; the top card auto-renders inline as a chart. It replaces a generic web search for data lookups.",
+  "Reconnaissance and chart retrieval across the live web and proprietary data: many results at once, returned as structured cards and web links, and the top card auto-renders inline as a chart.",
```
noahjax, line 47: **[Suggestion]** Two absolutes overstate the case: "It locates data; it does not deliver values" and "A card's rows are usually license-gated". `include_contents` defaults true and non-gated cards inline a free 20-row preview — `README.md:252` still advertises exactly that, and the skills tell the model to read headline values off `description`. A model that believes search never carries values will spend a `tako_answer` call it did not need.

Suggest naming the gated sources instead of generalizing: "rows from S&P Global, Visible Alpha, FactSet and CoinMarketCap are license-gated; other cards inline a free 20-row preview."

Separately: `1bb4be0` (the answer-first / search-as-specialist repositioning) is not in the PR description at all, though it is the highest-blast-radius change here — it shifts the traffic mix toward the arbiter path, with different price and latency, across `registry/server.json`, `lhm.plugin.json`, `llms-full.txt` and all three skills. Every other claim in this PR carries a live call count; this one carries none. Worth its own section in the body, and ideally a number.

llms-full.txt
```diff
@@ -33,7 +33,7 @@ Tools are discovered at runtime via `tools/list`. There is no per-call `api_toke
...
...ad or compound query. To see what proprietary data exists — or confirm a specific figure exists before you spend a query — check `tako_available_data` first (free). Ground a query by pinning resolved graph ids. For a card's full rows call `tako_contents` — only when the card is marked `exportable: true`. A non-exportable card (`exportable: false`, usually license-gated) carries no rows: read its headline value from the card's `description` when it carries one, or get specific figures via `tako_answer` with the card's `nodes` ids pinned (each such card carries a `values_hint` saying exactly this). Repeated source/methodology paragraphs are hoisted out of the cards into the trailing Source Notes section of the markdown, keyed by source name — look a card's `source_name` up there (truncating clients lose boilerplate before data). For one written answer use `tako_answer`; for open-ended research use `tako_agent` (opt-in). Results arrive as a markdown document — a Tako Data section (per card: headline, exportable flag, node ids, chart link, recent rows), then Web Results, then Source Notes; machine essentials (request_id, usage, chart-widget fields) ride separately in structuredContent.
```
noahjax: **[Nit]** This paragraph still promises "a Tako Data section (per card: headline, exportable flag, node ids, chart link, recent rows)" — the rows are a count pointer now. Same paragraph, and `tako_answer`'s at :53, still end "machine essentials (request_id, usage, ...) ride separately in structuredContent", which now understates it: `structuredContent` carries the whole payload. `README.md:252-253` describes the old shape too.

skills/tako-financial-research/SKILL.md
```diff
@@ -1,45 +1,70 @@
...
+- Comparison → tako_answer {"query": "How do Coca-Cola and PepsiCo compare on annual revenue?", "sources": ["data", "web"]} → check both entities appear in the cited card's `nodes` before treating it as a comparison
+- Chart is the deliverable → tako_search {"query": "Coca-Cola vs PepsiCo revenue", "sources": ["data", "web"]} → embed the card; ask `tako_answer` if you also need the values
+- Coverage question (free; note the arg is `q`) → tako_available_data {"q": "Costco"} → then pin what it returns: tako_answer {"query": "Costco Wholesale Corporation annual revenue", "sources": ["data"]}
+- Growth rate / ratio → pull the levels, then compute: tako_answer {"query": "Apple annual revenue for FY24 and FY25", "sources": ["data", "web"]} → compute the % change yourself
+- Breadth recon → one narrow `tako_search` per company in parallel with `"sources": ["data"]` to see what exists; switch to `tako_answer` per company once you need the figures
+- Not in the graph → tako_answer {"query": "When is Nvidia's next earnings date?", "sources": ["data", "web"]} → no card carries it; the answer comes from the web citations, so say the figure is web-sourced
 ## Output (tight and structured)
```
noahjax, line 68: **[Nit]** Missing blank line before `## Output`, so the heading sits flush against the last `## Examples` bullet. Same in `skills/tako-macroeconomics/SKILL.md:61` and `skills/tako-web-traffic/SKILL.md:55` — the rewrite ate the separator in all three, and the README copies inherit it.

noahjax (changes requested):
Verified locally on the merged branch state: typecheck clean, 619 worker + 16 widget tests pass, `schemas:check` and `registry:check` clean. The freshness/relevance fix in section 2 is a real bug with the wire shape verified behind it, and section 6's `payloadPointer` cleanly extends the `rowsPointer` pattern.

Two things gate merge, both a few lines: a committed `node_modules` symlink, and an `instanceof` gate in the new batch path that discards 404 guidance and leaks the internal API path. Detail inline.

**Blocker: `workers/node_modules` is committed as a symlink to the author's home directory.**

```
workers/node_modules -> /Users/eric/tako-mcp/workers/node_modules   (mode 120000)
```

Added in `1bb4be0` and present in every commit after it, so it lands on `main` on squash-merge. `.gitignore` has `node_modules/` with a trailing slash, which matches directories only; a symlink is a blob, so it slipped through. It breaks `npm ci` on a fresh clone. Fix: `git rm --cached workers/node_modules`, and add `node_modules` without the slash to `.gitignore`.

**`coverage_filter` was released.** It shipped in **v0.15.0** (published 2026-07-29T01:47Z) and in that release's `registry/server.json`, the discovery card `publish-mcp.yml` pushes to the public MCP registry. Staging has served it since #179 merged. The v0.15.0 *production* deploy failed, so prod likely never served it, and the practical blast radius is small. But removing a released input under a `fix:` title gives a patch bump where AGENTS.md says a breaking change bumps the minor pre-1.0. Zod strips unknown keys, so a caller pinned to the old schema gets a silently unfiltered result rather than a 400. Worth correcting the claim in the description either way.

**On the channel inversion (section 5).** Agreed on carrying the payload in `structuredContent`: v0.14.0's `mcp.ts` set `result.structuredContent = output` and `tako_search` had no `renderText`, so every release up to v0.14.0 shipped the full payload in *both* channels. This restores that side, and no structuredContent-reading client can regress. Worth being explicit that the text half is the new bet: no release has ever served pointer-only text, and `mcp.ts:1002` names the long-tail hosts (Cursor, Windsurf, Gemini CLI, LibreChat) whose forwarding behavior is unverified here. One smoke check on a non-Claude host would retire that risk cheaply.

**Where the description claims a cleanup that did not land.** Section 6 opens by faulting a doc comment that "cited a rationale that was never actually written" and says the misleading test was "retitled and split". Neither happened: `renderContentsText`'s docstring now describes the opposite of what the function does, the old docstring was orphaned above `payloadPointer`, and the test title still reads "payload channels dropped" while asserting the payload is present. Flagged individually below.

Unrelated FYI: the release deploy to production failed for both v0.14.0 and v0.15.0, so whatever ships next carries everything since v0.13.0 at once.

workers/src/tools/_render_markdown.test.ts
```diff
@@ -421,49 +392,66 @@ describe("renderAgentRunMarkdown + slim", () => {
...
+        download_url: "https://signed/csv",
+        expires_at: "2026-07-29T00:00:00Z",
+        cost: 0,
+      }],
       cost: 0,
     });
     expect(md).toContain("Download: https://signed/csv (expires 2026-07-29T00:00:00Z)");
   });
 
   it("slims structuredContent to metadata only (payload channels dropped)", () => {
```
noahjax: **[Issue]** Section 6 says this test was "retitled and split so each half is asserted explicitly." It was neither. The title still reads "slims structuredContent to metadata only (payload channels dropped)" while the body asserts the payload **is** present and not dropped — the exact title/body contradiction section 6 identifies as the regression's only guard having failed.

It is worth actually splitting, because the two halves guard different things and one of them is the thing that regressed:

```ts
it("carries the full payload in structuredContent", () => { /* toEqual(...) */ });
it("keeps the payload out of the text channel", () => { /* not.toContain("big page text") */ });
```

As written, a future change that puts the payload back into the text channel fails a test whose name says the opposite of what it checks — which is how the first regression got through.

workers/src/tools/tako_contents.ts
25eliu: Fixed in e91e204: `errorText` now gates on `DjangoError` (the base class), not `DjangoHttpError`, with an `"body" in reason` guard since DjangoTimeoutError/DjangoResponseParseError carry no body. Added tests for both the 404-in-batch (modelGuidance preserved) and 401-in-batch (no path leak) cases you flagged.

workers/src/tools/tako_contents.ts
25eliu: Fixed in e91e204: added `BATCH_CHAR_BUDGET` (200_000), split evenly across the batch for the silent default only — an explicit caller `max_chars` is left untouched at any batch size (matches this tool's over-asks-fail-fast policy for `max_rows`). Flagged the 200_000 number in-code as a starting point, not a measured one — happy to tune it if you have a better number. Added tests for the 4-url split, the 2-url no-op boundary, and explicit-max_chars-unaffected.

workers/src/tools/tako_contents.ts
25eliu: Fixed in e91e204: docstring, DESCRIPTION, and llms-full.txt's tako_contents entry all corrected to reflect the 1-10 url batching.

workers/src/tools/tako_contents.ts
25eliu: Fixed in e91e204: added tests for neither url/urls given, both given together (urls wins, now asserted), and a >1-result renderContentsText case (header/numbered-sections/failed-entry). Also fixed the single-item render tests to carry `url` (which the handler always sets) with an order-aware assertion instead of `.startsWith`, in workers/src/tools/_render_markdown.test.ts (commit dff2260).

workers/src/tools/_render_markdown.ts
25eliu: Fixed in dff2260: corrected renderContentsText's docstring to describe the pointer it actually emits, not the fencing it used to do.

workers/src/tools/_render_markdown.ts
25eliu: Fixed in dff2260: deleted the orphaned stale docstring block.

workers/src/tools/_render_markdown.ts
25eliu: Fixed in dff2260: deleted the stale fencing comment, kept the correct one.

workers/src/tools/_render_markdown.ts
25eliu: Fixed in dff2260: deleted the dead `WIDGET_KEYS` constant.

workers/src/tools/_render_markdown.ts
25eliu: Fixed in dff2260: switched the `??` chain to `||` so an empty-but-present `dataset.rows` correctly falls through to `records`.

workers/src/tools/tako_search.ts
25eliu: Fixed in 6be8ad0, but only partially resolved by design — see the commit message. Reworded to stop claiming search never delivers values (it does, for exportable cards, via the free 20-row preview) without inventing a specific vendor list for the license-gated case, since I can't verify that list from the codebase. Left as a judgment call for you: do you want specific source names (S&P Global, Visible Alpha, FactSet, CoinMarketCap, ...) named in the model-facing description, and if so, is that list accurate/current? On the missing call-count evidence for 1bb4be0 in the PR body: I added a section covering it (§6 in the current body) but have no live-call data to cite — that's a gap I can't close without actually running traffic against prod, flagging it to you rather than fabricating a number.

llms-full.txt
25eliu: Fixed in 2821048: llms-full.txt's tako_search paragraph and both tools' trailing structuredContent lines corrected; README.md's tako_answer/tako_contents rows had the same defects and got the same fix.

skills/tako-financial-research/SKILL.md
25eliu: Fixed in 2821048: added the missing blank line in all three skill files and their README.md copies.

workers/src/tools/_render_markdown.test.ts
25eliu: Fixed in dff2260: actually split into two tests this time — "carries the full payload in structuredContent (the only copy)" and "keeps the payload OUT of the text channel — a pointer only" — with a comment above them explaining why the split matters (this exact test collapsed into one title/body contradiction is how the regression got through review).

workers/src/tools/_available_data.ts
```diff
@@ -37,19 +37,21 @@ export const PAGE_LIMIT = 100;
...
+// every node up to the server's own cap.) Names are reordered headline-first
 // across everything FETCHED before this slice is taken, so low-signal
-// accounting names are what the cap drops, and `total`/`truncated` still
-// report when more exist server-side — a "200 of 250+" list is explained,
-// not an unexplained second cap.
-export const MAX_COVERAGE_NAMES = 200;
+// accounting names are what a genuine >250 cap still drops, and
+// `total`/`truncated` still report when more exist server-side — a
+// "250 of 400+" list is explained, not an unexplained second cap.
+export const MAX_COVERAGE_NAMES = 250;
```
noahjax, line 52: **[Suggestion]** Good change on its own, but the stated rationale is arithmetically inconsistent with the observation it claims to fix, and I do not think §3's gap is actually closed.

`selectCoverage` is `names = ordered.slice(0, MAX_COVERAGE_NAMES)`, and `orderMetricNames` only reorders — its own docstring says "never drops a name." So `names.length === min(items fetched, MAX_COVERAGE_NAMES)`. If Centene showed **~50** names against ~250 available, the 200 cap was never the binding constraint: at 200 you would have seen 200. Raising it to 250 cannot turn 50 into 250.

The actual limiter has to be the pagination loop — `MAX_COVERAGE_PAGES = 4` multiplied by the server's *real* page size. Landing at ~50 names implies pages of roughly 12-25 items, not the `PAGE_LIMIT = 100` the arithmetic assumes.

That matters for the fix, not just the write-up: "`ceil(250/100) = 3`, well inside the existing `MAX_COVERAGE_PAGES = 4` slack" holds only at full 100-item pages. For any actual page size below ~63, four pages cannot reach 250 and `MAX_COVERAGE_PAGES` becomes the binding cap — so the Centene case stays broken while the constant now claims otherwise.

Meanwhile the change does add a third sequential page and ~25% more names to the payload of the tool advertised as "Free and fast" and positioned as the recommended first call, reversing what the old comment called "a deliberate token/latency trade."

Not blocking, and I would not revert it — a ceiling that matches the server's rather than undercutting it is the right shape. But worth one free `tako_available_data` call on Centene to count the returned names before the commit message and §9 stand as the explanation. If the count comes back near 50 again, the lever is `MAX_COVERAGE_PAGES` (or a page-size assumption that needs measuring), not this constant.

noahjax (approved):
Approving — every finding from the last round is addressed, and I checked the code rather than the commit messages.

Verified on `1ee1396`: typecheck clean, **630 worker tests** (up from 619) + 16 widget, `schemas:check` and `registry:check` clean.

Both merge gates are properly closed. `errorText` now gates on the `DjangoError` base with `DjangoError` imported and `body` guarded by an `in` check, and the two new regression tests are the right ones — a 404 inside a batch keeps its `modelGuidance` plus the spliced detail, and the 401 test asserts `not.toContain("/api/v1/contents")` with the AGENTS.md rule cited inline. The symlink is untracked and the `.gitignore` entry carries the reason (trailing-slash patterns match directories only, so a mode-120000 blob slips through) — that comment is worth more than the fix.

`BATCH_CHAR_BUDGET` landed well: applying the split to the **default only** keeps n=1 and n=2 byte-identical to today while n=10 drops to 25k, and deriving the test's batch size from the constant means the assertion keeps biting if the constant moves. The split test at `tako_contents.test.ts:410` is the kind that survives a retune.

Two things to carry forward — neither blocking.

**1. §9's rationale does not match the code** (see the inline comment on `MAX_COVERAGE_NAMES`). The 200-name cap cannot have been what limited Centene to ~50 names, so raising it to 250 probably does not close §3's gap. Cheap to settle: re-run the Centene call and count the names. It is the free tool.

**2. The title still says `fix:`, so this ships as a patch.** §3 correctly notes that removing a released input "wants a **minor** bump pre-1.0, not the patch a `fix:` title implies" — but nothing acted on it, and squash-merge feeds the PR title straight to release-please. If you want the minor, retitle before merging.

workers/src/tools/_available_data.ts
25eliu: Ran the call you asked for. You're right that the write-up was wrong, and the measurement inverts which part.

**Anonymous `tako_available_data q=Centene` against prod** (`mcp.tako.com`, v0.15.0, so `MAX_COVERAGE_NAMES = 200`):

```
Centene Corporation (ent::centene_corporation::f41593cd) — metrics, total "250+"
  names returned: ~200
CNC (STOCK_TICKER) — metrics, total 59
  names returned: 59  (complete, well under the cap)
```

~200 rather than an exact 200 because the text channel is comma-joined and at least two metric names contain commas themselves (`Net PP&E, Excl. Operating Leases`, `EBT, Incl. Unusual Items (Normalized)`) — 201 comma fields, so 200 ± 1. It is the cap, not 50.

**So the arithmetic critique lands, but the conclusion flips.** Your reasoning about `selectCoverage` is exactly right — `orderMetricNames` returns `[...clean, ...noisy]` and never drops, so `names.length === min(fetched, MAX_COVERAGE_NAMES)`, and "50 shown against 250 available" was therefore impossible with a 200 cap. That was the tell, and the resolution is that **the 50 was the bad number**, not the cap theory. Reaching ~200 names is unreachable in 4 pages at any page size below 50 and needs at least two full 100-item pages, so `PAGE_LIMIT = 100` is the real page size, `MAX_COVERAGE_NAMES` genuinely was the binding constraint at 200, and `ceil(250/100) = 3` inside `MAX_COVERAGE_PAGES = 4` holds. `MAX_COVERAGE_PAGES` is not the lever.

**Where you're still right and I've corrected the claim: §3's gap is not closed.** Prod runs the 200 cap, so the ~50 names this bump newly exposes cannot be inspected from prod, and I confirmed `Total membership - Medicare` is absent from all ~200 currently returned. Whether it sits in the 200–250 band or past the server's own 250 counting cap — in which case the lever is server-side and no constant here fixes it — needs one call against staging once this lands. §9's heading claimed "closes §3's known gap"; that was unearned. The body now says **narrowed and measurable**, not closed, with the staging check named as the follow-up.

Both the §3 note ("50 shown" → the measured ~200 out of `250+`) and §9 (the page-size assumption now stated as measured, with the not-proven part called out) are corrected in the PR description. The `68837a2` commit message keeps the stale number — I'd rather not rebase an approved branch for it, and the squash-merge body will carry the corrected text.

Your cost point is recorded as an accepted trade rather than argued away: this does add a third sequential page and ~25% more names to a tool advertised as "Free and fast" and positioned as the first call. Taken deliberately, on the view that a coverage list silently undercutting the server's own ceiling is the worse failure for a tool whose whole job is answering "is this covered?".

PR #191, opened 2026-07-30
feat(available_data): entity+metric lookup, a relevance gate, and a pin form that works

25eliu:
## What this changes

`tako_available_data` was the weak link in the routing chain: it could tell you a name existed but not hand over the handle that fetches it, it reported coverage for the wrong company, and the `next_call` it advertised was null on every real query. This branch fixes the tool and the three instruction surfaces that pointed at a pin form measured to be a no-op.

Everything here was found by calling live staging/prod, not by reading code. Measurements are in the commit messages and code comments.

### 1. A lookup path: `q` + `metric`

`tako_available_data {q: "Carnival", metric: "passenger cruise days"}` now resolves the entity+metric **pair** and returns a runnable handle.

- **~0.63s vs ~3.0s**, and **~600 chars of output vs ~8.5k**, versus the coverage drill it replaces.
- Splitting the halves is what makes it accurate: the combined phrase resolves the metric fine but destroys entity resolution (`"Pfizer R&D expense"` returns no Pfizer at all in the top 10).
- Both halves echo back their resolution plus 2 alternates. The top metric is right ~80% of the time and the top three ~93–95%, and a wrong top-1 is wrong *visibly* (`metric="total revenue"` → `Total Odds`, with `Revenues` at rank 1) — so naming the runners-up converts a silent failure into a free self-correction.
- Swapped arguments are **diagnosed, never auto-corrected**: `q="gross margin", metric="Nvidia"` used to return `found: true` with a runnable handle for garbage. Detection needs *unfiltered* probes — `types=entity` returns an entity for "gross margin" no matter how entity-unlike it is. Measured 0 false positives over 20 correct pairs (including adversarial names like `Block`, which ranks the metric `Blocked Shots` first) and 9/9 on swapped ones.

### 2. A relevance gate

`graph/search` is fuzzy and its nodes carry **no score field**, so there is nothing backend-side to threshold on. Ungated, the drill budget went to whatever ranked highest:

- `q="Carnival Corporation"` returned `found: true` describing **Cuscal Limited** — an Australian payments company that carries "Carnival Corporation" as an alias. That is poisoned data upstream; ranking name matches above alias matches mitigates it here. **A backend alias audit is the real fix.**
- `q="UnitedHealth Group"` spent half its drill budget on **Blackstone Inc.**
- `q="the vibes of tuesday"` reported 250+ metrics on **Tuesday Morning Corporation**.

The gate is bidirectional token containment over name-or-alias. Aliases are mandatory, not a refinement: `SpaceX` → `Space Exploration Technologies Corp.` shares zero name tokens and matches only through its alias list. It **fails open** — a false "Tako has no data on X" is worse than the noise the gate removes — and it decides *confidence*, never *order*, on the metric half, because token containment gets morphological variants backwards (it rejects the correct rank-0 `Revenues` for `metric="revenue"` while keeping `Avnet Revenue Total Revenue`).

### 3. Inspect wide, render narrow (was: expand 2, render 2)

`SELECT_TOP_N = 4` / `RENDER_FULL_N = 1`. These were one number that conflated two different costs — a `limit=1` probe never enters the model's context, while a rendered coverage list is ~8.3k chars. Measured on prod, the second match's list was **8,283 of a 19,046-char answer**: 43% of the payload spent on an entity nobody asked about. Depth 4 is not arbitrary — over 16 resolvable queries the first plausible candidate *with coverage* sat at rank <2 in 14 of them but rank <4 in all 16 (`q="Carnival"` is the case depth-2 misses: ranks 0–2 are zero-coverage name variants).

### 4. `structuredContent` carries node ids

`selectCoverage` discarded every coverage node id on one line (`items.map(i => i.name)`), so the tool could name a metric but not hand over the handle. `coverage.items[]` now pairs each name with its id, and `next_call` populates whenever the measure is known.

### 5. The pin form, everywhere

Measured on prod: a pin at the default `strict: false` **does not steer retrieval at all** — a deliberately *wrong* node changed nothing. `strict` is an OR over pinned nodes, so adding the entity id widens the filter back out (it turned "no such card" into a plausible-looking wrong metric). Only **the metric node alone, with `strict: true`** works.

Every surface that previously advised the broken form now names the working one: `tako_search`/`tako_answer`/`tako_contents` descriptions, `SERVER_INSTRUCTIONS`, zero-result guidance, card `values_hint`, README and all three bundled skills.

`SERVER_INSTRUCTIONS` also never mentioned `tako_answer` at all — the tool the instructions elsewhere call the starting point.

### 6. `structuredContent` can no longer violate the published schema

`registerTool` takes a `ZodRawShape`, so the SDK rebuilds our `z.looseObject` schemas as **strict** and publishes `additionalProperties: false`. When a slim hook failed conformance the old fallback served the *full* output — which carries `sources_glossary`, `cards`, `web_results`, none of them declared — guaranteeing rejection, and a strict client discards the **entire result including the text block**, on a call already billed. It fired on data-grounded responses (they are the ones with a glossary), destroying precisely the good ones. Now: narrow to declared keys, re-check, and omit `structuredContent` entirely if that still fails.

### 7. Card ordering, and a wrong-number failure

The backend's rank is not recency-aware. `tako_search "US inflation rate"` returned `United States Inflation Rate` (latest row **2024-01-01**) above `United States CPI Inflation Rate (Seasonally Adjusted)` (latest row **2026-06-01**). A model reading top-down reports a 2½-year-old figure as current — a wrong number carrying a citation. `orderCardsByUsefulness` orders rows-bearing → freshest → relevance → backend order. **Ordering only**; nothing is dropped, and it yields entirely if the answer prose carries positional citation markers.

### 8. Token trimming

`usage`'s nested cost breakdown cost 197 published tokens each on `tako_search` and `tako_answer` — ~394 of a 7,235-token connect surface — to describe a per-call breakdown no routing decision reads. The breakdown is still **emitted**; only the advertised shape narrows.

### 9. `npm run golden`

20 hand-checked routing cases. Unit tests pin pure logic and smoke pins the protocol; neither could see any bug in this PR.

Assertions are **invariants, never fixtures**: `tako_search` is nondeterministic (identical requests returned 1, then 2, then 1 cards across three consecutive runs), so exact card sets would flake on day one. Resolution *is* deterministic across five runs, so that is what gets pinned. Expectations are hand-checked because a token-overlap scorer was tried and measured to over-credit (it passed `metric="employees"` → `Revenue Per Employee`).

The 20 resolution cases are **free** (graph endpoints only); 3 execute their `next_call`. CI runs it after smoke with `continue-on-error`, since it asserts against live graph data.

## What this does NOT fix

> [!WARNING]
> **CORRECTED 2026-07-31 — this section's original claim was wrong.** It read:
>
> > *Card coverage is the real ceiling, and it is a backend matter. On fresh cases resolution was 3/3 but retrieval 1/3: mainstream financial and macro metrics have cards, operational KPIs (backlog, GMV, passenger cruise days) do not.*
>
> Measured with matched arms — same tool, same query text, 3 repeats, the only variable being whether the resolved node is pinned — **the cards exist and the pin was hiding them**:
>
> | query | pinned `strict` | unpinned |
> |---|---|---|
> | `Carnival Corporation passenger cruise days` | `[0,0,0]` | `[3,3,3]` |
> | `United States CPI` | `[0,0,0]` | `[3,3,3]` |
> | `Apple Inc. P/E ratio` | `[0,0,0]` | `[3,3,3]` |
> | `United States government debt to GDP` | `[0,0,0]` | `[3,3,3]` |
>
> And the recovered cards are the requested metric, not lookalikes:
> `Carnival Corporation Ltd. Passenger Cruise Days 2 (Annual)`,
> `Carnival & plc Passenger cruise days (PCD) (Quarterly)`, answering *"25.7 million
> passenger cruise days in the quarter ended June 30 2026"*.
>
> Across 20 handles: **11 retrieve strictly fewer cards pinned than unpinned**, 7
> unchanged, 2 better pinned.
>
> **The real cause is near-duplicate metric nodes.** The graph holds an empty twin
> beside the card-backed one and `available_data` resolves the empty one:
> `mt::passenger_cruise_day` is dead while the cards hang off a node named
> `Passenger Cruise Days 2`; `mt::consumer_price_index` is dead while
> `mt::cpi_inflation_rate_s` is live. Filed as **KE-812** (Urgent).
>
> So "operational KPIs have no cards" was an artifact of this PR's own pinning, and
> Tako's coverage is better than this description claimed. Reproduce with
> `npm run eval:pin`.

**What IS a real ceiling:** `available_data` cannot tell a card-backed metric node from its empty twin, because nothing in `graph/search` or `graph/related` exposes retrievability. So it can report a name, set `found: true`, and hand over a handle that retrieves nothing — measured at ~26% of emitted handles. That needs KE-812; no amount of MCP structure fixes it.

**Also corrected:** "0 cards from the `next_call` is the definitive answer" is not safe advice. On a zero-card pinned result the same query unpinned often returns the data, so that instruction converts recoverable misses into a confident "Tako has no data". Tracked as F13 in `workers/scripts/evals/available-data/FINDINGS.md`.

Also unfixed, deliberately: the entity half of the lookup path is **not** coverage-probed. Probing it was implemented and measured — it fixes the echoed entity on ambiguous names but costs **0.23s → 1.48s (6.4×)** per call. Under `strict` the metric pin drives retrieval and the alternates already name the covered variant, so it was reverted with the measurement recorded.

## Two pre-existing breaks on main, fixed here

Both were smoke assertions passing over nothing:
- `available_data`'s summary was read off `structuredContent`, where it has never been (it renders to the text channel).
- `tako_contents`'s `.data`/`.download_url` were read off the envelope root, which has returned `undefined` since the tool became batched.

## Test plan

- [x] 711 worker + 16 widget tests
- [x] 4 typecheck projects (`npm run typecheck`)
- [x] `schemas:check`, `registry:check`
- [x] 20/20 golden against live staging
- [x] smoke green against a local worker on the staging backend — except `tako_visualize`, which 402s because the test account has zero credits (`get_credit_balance → 0`), unrelated to this branch
- [x] receipt-line noun fix verified live on both node kinds
- [ ] deploy to staging and re-run golden + smoke against the deployed worker
- [ ] spot-check a real ChatGPT / Claude session for the `q` + `metric` path


workers/src/tools/_match_gate.ts
```diff
@@ -0,0 +1,205 @@
...
+  // `Space Exploration Technologies Corp.` because no name match survives to
+  // outrank it. Ties (both name matches, e.g. Delta Air Lines vs Delta Corp)
+  // keep the backend's order.
+  const byName: T[] = [];
+  const byAlias: T[] = [];
+  for (const c of candidates) {
+    if (plausibleMatch(query, { name: c.name })) byName.push(c);
+    else if (plausibleMatch(query, c)) byAlias.push(c);
+  }
+  const kept = [...byName, ...byAlias];
```
noahjax, line 160: **[Issue]** This promotion demotes the correct node on real queries.

Ran this function over live staging `graph/search` results, then took each candidate's coverage from `graph/related?limit=1`:

| query | backend rank0 | gate rank0 |
|---|---|---|
| `US inflation` | United States - **250** metrics | US Savings Inflation Securities - **1** |
| `S&P 500 earnings` | S&P Global Inc. - **250** | Earnings - **3** |
| `SpaceX` | Space Exploration Technologies Corp. - **250** | SpaceX - **0** |

The defence against poisoned aliases is real, but a bare name match is not evidence that the node is the one the caller meant - `US Savings Inflation Securities` matches `US inflation` by name and carries one metric. On `main` the ungated backend order got all three right, so this is a regression on the discovery path.

The cheapest fix is not here: leave the partition alone and make the coverage promotion at `tako_available_data.ts:583` pick the best-covered probed candidate rather than the first non-zero one. That keeps the Cuscal fix and removes this failure at the same time. A golden case with `{q: "US inflation"}` asserting the rendered match is `United States` would pin it.

workers/src/tools/_match_gate.ts
```diff
@@ -0,0 +1,205 @@
...
+  candidates: readonly T[],
+): { kept: T[]; gated: boolean } {
+  // Name matches rank above alias-only matches, order preserved within each
+  // group. This is NOT general re-ranking (see the note above) — it is the one
+  // defence against poisoned alias data, which is real and observed: on
+  // staging, `Cuscal Limited` (an Australian payments company) carries the
+  // aliases "Carnival Corp", "Carnival Corporation", "Carnival Cruise", so a
+  // gate that treats names and aliases alike ranks it above the actual
+  // `Carnival Corporation`. Preferring the node whose OWN NAME matches fixes
+  // that without breaking the case aliases exist for: `SpaceX` still resolves
```
noahjax: **[Issue]** This claim does not hold on live staging.

`q="SpaceX"` returns a node literally named `SpaceX` alongside `Space Exploration Technologies Corp.`, so a name match *does* survive to outrank the alias match, and this function returns `SpaceX` at rank 0. The right answer still gets rendered, but only because that impostor happens to carry zero coverage and the promotion at `tako_available_data.ts:583` fires. The mechanism named here is not the one doing the work.

Worth correcting, because the comment is what a future reader will trust when deciding whether the partition is safe to keep - and the same reasoning is what `_match_gate.test.ts:147` encodes, by constructing a candidate set where no name match exists.

workers/src/tools/tako_available_data.ts
```diff
@@ -198,108 +557,63 @@ const tako_available_data = {
...
+      drillMatches(candidates.slice(0, RENDER_FULL_N)),
+      Promise.all(candidates.slice(RENDER_FULL_N).map((n) => coverageProbe(n))),
+    ]);
+
+    let matches = firstDrill;
+    let rendered = candidates.slice(0, RENDER_FULL_N);
+    if (!firstDrill.some(hasLiveCoverage)) {
+      // Rank 0 was a zero-coverage name variant. Promote the first PROBED
+      // candidate that actually has coverage and drill that one instead.
+      const promoted = probes.find((pr) => pr.total > 0);
```
noahjax: **[Issue]** `total > 0` is too weak a bar to rescue a mis-ranked rank 0.

Measured on staging, `q="US inflation"` gives candidates `[US Savings Inflation Securities (1), United States (250), Inflation Rate (63), CPI Inflation Rate SA (1)]`. Rank 0 has coverage 1, so `hasLiveCoverage` is true, this branch never runs, and a one-metric node is rendered as the answer with `found: true`. Same shape on `q="S&P 500 earnings"` (`Earnings` at 3 beats `S&P Global Inc.` at 250).

Two changes make this robust:

1. Pick the best-covered probe, not the first non-zero one: `probes.reduce((best, p) => (p.total > (best?.total ?? 0) ? p : best), undefined)`.
2. Enter this branch when rank 0 is materially out-covered by a probe, not only when it has zero coverage.

That second condition is what converts the gate's name-match preference from a hard ordering into a tiebreak, which is what the comment at `_match_gate.ts:26` says it is meant to be.

workers/src/tools/_search_results.ts
```diff
@@ -488,6 +506,22 @@ export const usageSchema = z
...
+ * `usageSchema` above stays the internal contract (it parses and types the
+ * wire, breakdown included, and the breakdown is still EMITTED). This declares
+ * just `total_cost_usd` because publishing the nested compute/data objects cost
+ * 197 tokens on each of tako_search and tako_answer — ~394 of a 7,235-token
+ * connect surface — to describe a per-call cost breakdown no routing decision
+ * reads. Loose, so the emitted breakdown still validates: nested loose objects
+ * publish permissive `additionalProperties`, unlike the top level which the SDK
+ * rebuilds strict.
+ */
+export const usageAdvertisedSchema = z
```
noahjax, line 583: **[Issue]** `usageAdvertisedSchema` is exported and referenced by nothing.

`searchSlimOutputShape` and `answerSlimOutputShape` (`_render_markdown.ts:53` and `:70`) both still declare `usage: usageSchema`, so the nested `compute` / `data` objects are still published and the ~394 tokens the PR description claims back are still being spent. Either wire this into both slim shapes or drop it - as it stands the doc comment describes a change that did not land.

workers/src/tools/_search_results.ts
```diff
@@ -599,20 +648,146 @@ function buildZeroResultGuidance(
...
+      index,
+      rows: latestDataPoint(card),
+      asOf: freshnessEpoch(card),
+      relevance: relevanceRank(card),
+    }))
+    .sort((a, b) => {
+      const aHasRows = a.rows !== null;
+      const bHasRows = b.rows !== null;
+      if (aHasRows !== bHasRows) return aHasRows ? -1 : 1;
+      if (a.rows !== null && b.rows !== null && a.rows !== b.rows) return b.rows - a.rows;
```
noahjax, line 810: **[Issue]** This compares magnitudes across cards that `temporalMagnitude` produces on different scales.

`temporalMagnitude` returns `Date.parse(v)` for strings (epoch-ms, ~1.7e12) but passes numbers through unchanged. That is safe in `capRecentRows`, which only ever compares two values from the *same* card. Here it compares across cards.

So a card whose `date`-typed column holds numeric years (`2024`) yields `rows = 2024` and sorts below every string-dated card forever - including genuinely staler ones. Annual series are the likely shape to hit this.

Normalising in `latestDataPoint` would contain it: treat a bare number in a plausible year range as `Date.UTC(year, 0, 1)`, and otherwise only compare values that came from the same parse path.

workers/src/tools/tako_available_data.ts
```diff
@@ -156,36 +201,350 @@ const tako_available_data = {
...
+        };
+      }
+
+      return {
+        // `found` here reports RESOLUTION, not card availability — no free
+        // signal for the latter exists (see the description). The next_call
+        // is what confirms it.
+        // Both halves required: a resolved metric with no entity is not a
+        // usable pair, and the summary is routing the caller elsewhere.
+        found: pair.entity !== null && pair.metric !== null && metricConfident,
```
noahjax, line 570: **[Suggestion]** `found` now means two different things, and the advertised description states only one.

`fullOutputSchema.found` says *"True when at least one match has live data coverage - not mere node resolution. A resolved node with no coverage (or whose coverage lookup failed) yields false."* On this path it is set from resolution plus metric confidence, with no coverage lookup performed at all. The code comment above acknowledges it; the description the model reads does not, and the model has no way to tell which path produced the response.

Suggest the description name both: coverage-backed on the discovery path, resolution-only on the lookup path, with the `next_call` as the coverage check. Otherwise an agent that trusts `found: true` here will report data exists when only a name does - which is the exact failure the "What this does NOT fix" section warns about.

workers/src/tools/_render_markdown.ts
```diff
@@ -415,33 +484,123 @@ interface CoverageMatchLike {
...
+
 export function renderAvailableDataMarkdown(o: AvailableDataFullOutput): string {
+  // The lookup path resolves a pair instead of drilling a coverage list.
+  if (o.metric_query !== undefined && o.matches.length === 0) {
+    return renderAvailableDataPairMarkdown(o);
+  }
+  // Nothing plausibly matched: the summary already says these resolutions are
+  // almost certainly wrong, so printing 200 of their metric names would spend
+  // ~8.5k chars elaborating on an answer we just disclaimed.
+  if (o.confident === false) return o.summary;
```
noahjax, line 601: **[Suggestion]** Fail-open ends up stricter than no gate at all, and the two channels disagree.

When `gateCandidates` fails open, three things fire together: `found: false`, `next_call: null`, and this line drops every coverage name from the text. But `slimAvailableDataStructured` still emits `matches[].coverage.items` with `total > 0`. So a structuredContent reader sees coverage on a response whose `found` is false and whose prose says the resolutions are "almost certainly NOT what you asked for".

The contract at `_match_gate.ts:120` is that a false "Tako has no data on X" is worse than the noise the gate removes. `gateCandidates` honours that; this line does not. Withholding `next_call` already buys the safety - no priced call is spent on an unvetted target - so rendering the names with the low-confidence warning in front would keep the caller's answer without re-introducing the risk.

Separately: the handler has already paid for a full paginated drill plus three `limit=1` probes by the time it gets here. If the intent really is to suppress, it is cheaper to skip the drill when `gate.gated === false`.

workers/scripts/panel.ts
```diff
@@ -0,0 +1,632 @@
...
+});
+
+/** Where a key came from, so "I added the key" and "agent mode is off" can
+ *  never both be true without the reason being on screen. */
+const keySource = (name: string): string => {
+  if (process.env[name] === undefined) return "not set";
+  return originalEnv.has(name) ? "from environment" : "from .env";
+};
+
+server.listen(PORT, () => {
```
noahjax: **[Suggestion]** Bind to loopback.

`server.listen(PORT)` with no host binds every interface, and this panel proxies to prod with `PROD_TAKO_API_KEY` attached server-side. Anyone on the same network can reach `http://<your-lan-ip>:8801` and spend priced calls against that token, with no auth in front.

`server.listen(PORT, "127.0.0.1", ...)` - the header comment already says "tokens stay in the process", and this makes that true.

workers/src/mcp.ts
```diff
@@ -410,42 +412,92 @@ export function createMcpServer(
...
+ * `sources_glossary`, `cards`, `web_results` and more, none of them declared,
+ * so against the published `additionalProperties: false` it is guaranteed to
+ * be rejected — and a strict client discards the ENTIRE result, including the
+ * text block that holds the answer. The failure fires on data-grounded
+ * responses (they are the ones with a glossary), so it destroys precisely the
+ * good ones.
+ *
+ * Instead: narrow to the declared keys and re-check. That drops the undeclared
+ * extras rather than the answer. Only if the narrowed object still fails do we
+ * omit `structuredContent` altogether — spec-legal, and the payload rides in
```
noahjax: **[Nit]** "spec-legal" is not accurate.

MCP requires that a tool defining an `outputSchema` provide structured results conforming to it, so omitting `structuredContent` here is a spec violation too - just a quieter one than shipping a payload every strict client rejects.

The decision is right and the reasoning above it is the useful part. Suggest saying what it actually is: both branches violate the spec, and this one costs the caller only the structured channel instead of the whole result, text block included. A reader who trusts "spec-legal" will not think to look for the client-side fallout.

noahjax (changes requested):
Reviewed the two paths against live staging by importing this branch's own `gateCandidates` and running it over real `/api/beta/graph/search` results, then checking each candidate's coverage with `graph/related?limit=1`. The measurement discipline in this PR is excellent and the pin-form correction is clearly right. One finding blocks.

## The name-over-alias promotion picks the wrong node on ordinary queries

`gateCandidates` partitions name matches ahead of alias matches. On live staging that changes which node the discovery path drills, in the wrong direction, on 3 of 8 queries I sampled:

| query | backend rank0 | gate rank0 (drilled instead) |
|---|---|---|
| `US inflation` | United States - **250** metrics | US Savings Inflation Securities - **1** |
| `S&P 500 earnings` | S&P Global Inc. - **250** | Earnings - **3** |
| `SpaceX` | Space Exploration Technologies Corp. - **250** | SpaceX - **0** |

The zero-coverage promotion at `tako_available_data.ts:583` rescues only the third. For the first two, coverage is 1 and 3, both `> 0`, so `hasLiveCoverage` is true, no promotion fires, and the tool renders a 1-metric shell company as the answer with `found: true` while `United States` is demoted to a one-line "Also resolved" receipt.

This is the Cuscal/Carnival failure reintroduced from the opposite direction, and it is a regression against `main`, where ungated backend order resolved all three correctly. Making the promotion pick `argmax(coverage)` among the probed candidates instead of the first with `total > 0` fixes it without touching the gate.

Neither the unit suite nor the golden set can see it: `_match_gate.test.ts` constructs a SpaceX case where no name match exists (on staging one does), and no golden case uses a compound discovery `q`. A case like `{q: "US inflation"}` asserting the rendered match is `United States` would pin it.

## Design, as discussed

The gate belongs in the backend long-term. `graph/search` nodes carry no score, so there is genuinely nothing to threshold on today, and shipping token containment here is a reasonable way to test the combined-retrieval idea before committing it to the API. Please **file a follow-up** for the backend side: a relevance score on `graph/search`, plus the alias audit the description calls the real fix for the Cuscal poisoning. Same for `orderCardsByUsefulness` - the stale-top-card bug is a wrong number carrying a citation, and it affects the API, the app and the agent, not only MCP. Worth a ticket so the MCP-side ordering can eventually retire.

## Also

- `usageAdvertisedSchema` is defined and referenced by nothing, so PR item 8 is not delivered.
- `orderCardsByUsefulness` compares temporal magnitudes across cards on mixed scales.
- `found` means two different things across the two paths, and the schema description states only one.

Details inline.

workers/src/tools/_match_gate.ts
25eliu: Confirmed and fixed — I reproduced all three of your rows against staging before touching anything, and `q="US inflation"` was worse than the table shows: `found: true` on the 1-metric product, with `United States` demoted to an "Also resolved" receipt.

Took your cheapest-fix route: the partition is untouched and the promotion at `tako_available_data.ts` now does the work. Two changes, both yours — `argmax(coverage)` instead of first-non-zero (which mattered here: the probes are `United States` 250 **and** `Inflation Rate` 63), and the branch is entered when rank 0 is out-covered, not only when it is zero.

The "materially out-covered" threshold is an absolute floor rather than a ratio: `SHELL_COVERAGE_MAX = 5`. The failure is degeneracy, not proportion — the losers held 1 and 3 — and a ratio fires on `q="Delta"`, where both Delta nodes cap at 250 and only backend rank separates them. Live, after:

| q | rendered | was |
|---|---|---|
| `US inflation` | `United States` — 250+ | US Savings Inflation Securities — 1 |
| `S&P 500 earnings` | `S&P Global Inc.` — 250+ | Earnings — 3 |
| `Delta` | `Delta Air Lines, Inc.` — 250+ | unchanged |
| `Carnival Corporation` | `Carnival Corporation Ltd.` — 250+, no Cuscal | unchanged |

Added your golden case (`gate-shell-loses-to-coverage`, asserting the rendered match is `United States`) plus `gate-no-overpromotion` on `{q: "Delta"}` to pin the other direction. 22/22 green.

**One thing your run surfaced that I would not have found otherwise.** Two of my verification runs hit transient `graph/related` failures, and on `q="Delta"` that promoted `Delta Corp Limited` into the answer — because `hasLiveCoverage` is false for an `unavailable` match, so a *failed lookup* was being read as *no coverage*. That predates this branch and the old zero-only promotion had it too. Promotion now requires actually knowing rank 0's coverage; otherwise rank 0 keeps its own "couldn't load coverage, retry" line. Test pins it.

Filed KE-805 for the backend score, since this is the second heuristic stacked on the absence of one.

workers/src/tools/_match_gate.ts
25eliu: You are right, and I confirmed it: `q="SpaceX"` on staging returns a node literally named `SpaceX` at rank 0, `gateCandidates` returns it first, and the answer only survives because that impostor carries zero coverage and the promotion fires.

Comment rewritten to say what actually happens rather than what I assumed. It now names both cases where the partition gets live data wrong — the SpaceX impostor, and the `US inflation` shell from your other finding — and says explicitly that coverage handles them downstream, so nobody reads the partition as deciding relevance.

`_match_gate.test.ts` fixed too. Kept the old case (it does pin the alias mechanism when no name match exists) and added one that constructs the candidate set staging actually returns, asserting the impostor comes back at rank 0. Better to have the gate's real behaviour pinned than to have a test quietly vouching for a mechanism that is not doing the work — the golden suite covers the end-to-end outcome.

workers/src/tools/tako_available_data.ts
25eliu: Both changes taken, and your candidate list for `q="US inflation"` matched mine exactly.

1. `argmax` — done. Not cosmetic here: `Inflation Rate` (63) is also non-zero, so first-non-zero depended on backend ordering.
2. Enters the branch on material out-coverage, not only zero — done, as `rank0Coverage <= SHELL_COVERAGE_MAX && best.total > rank0Coverage`.

I used an absolute floor (5) rather than a proportional margin, because a ratio fires on `q="Delta"` (both nodes at the 250 cap) and on `q="Inflation Rate"` (a legitimately 63-entity metric whose rivals are thinner). The observed losers held 1 and 3, so degeneracy is the thing worth detecting. Reasoning and both counter-cases are recorded on `SHELL_COVERAGE_MAX`, with a note to retire it alongside the gate once KE-805 lands.

And you are right that this converts the name preference from a hard ordering into a tiebreak, which is what `_match_gate.ts:26` always claimed — I updated that comment to stop overclaiming in the meantime.

workers/src/tools/_search_results.ts
25eliu: Correct — item 8 had not landed. Wired `usageAdvertisedSchema` into both slim shapes and verified against a live `tools/list` off a running worker rather than trusting the code this time:

```
before: usage -> {total_cost_usd, compute: {cost_usd}, data: {cost_usd, datasets}}
after:  usage -> {total_cost_usd}
```

on both `tako_search` and `tako_answer`. The breakdown is still emitted — the advertised schema stays `.loose()`, so nested extras validate.

workers/src/tools/_search_results.ts
25eliu: Real bug, fixed as you suggested — normalised in `latestDataPoint` rather than in `temporalMagnitude`, so `capRecentRows` keeps its current semantics (same card, one parse path, where a raw serial is fine).

New `comparableEpoch` converts each recognised shape to epoch-ms: a bare year in 1000–9999 → `Date.UTC(y, 0, 1)`, epoch-seconds (1e8–1e11) → ×1000, epoch-ms (1e11–1e14) → through. Anything else returns null instead of a number on an unknown scale, which the caller already handles as "no freshness signal" — that is the second half of your suggestion, and it means an unrecognised value can no longer silently sort a card last forever.

Two tests: a bare-year 2026 card must outrank a string-dated 2019 one (the cross-scale case), and two bare-year cards must still order against each other (so the normalisation is not a no-op that merely stops comparing).

workers/src/tools/tako_available_data.ts
25eliu: Agreed — the code comment admitting it while the advertised description did not was the worst of both. Description now names both paths and the check that settles it:

> Means different things on the two paths, because only one of them can check coverage for free. Without `metric` (discovery): at least one match has live data COVERAGE — not mere node resolution […]. With `metric` (lookup): both halves RESOLVED to confident graph nodes, and no coverage check was performed — a chart may still not exist behind them. Either way, running `next_call` is what confirms retrievable data exists; 0 cards from it is the definitive answer.

That also makes it consistent with the "What this does NOT fix" section instead of quietly contradicting it, and a model can now tell which path answered from the presence of `metric_query`.

workers/src/tools/_render_markdown.ts
25eliu: The channel disagreement was real and is fixed — but I took your *second* suggestion rather than the first, so flagging the deviation explicitly.

I skipped the drill when `gate.gated === false` instead of rendering the names behind the warning. Reasoning: rendering them keeps the disagreement pointed the other way. On `q="the vibes of tuesday"` it would print ~8.5k chars of Tuesday Morning Corporation's metrics under prose saying the resolution is almost certainly wrong, and that is the case the suppression was added for. Skipping the drill makes the two channels agree *by construction* — there is nothing to disagree about — and takes the cost saving you identified in the same move. Live, one `graph/search` call and nothing else:

```
prose:  No graph node confidently matches "the vibes of tuesday". The closest
        resolutions are **Tuesday Morning Corporation (ORG)**, … — almost
        certainly NOT what you asked for.
struct: found=false  next_call=null
        Tuesday Morning Corporation | coverage.total=0 | items=0
        Tuesday Health, LLC         | coverage.total=0 | items=0
        Vibe Advertising Inc.       | coverage.total=0 | items=0
        Vibe HCM, Inc.              | coverage.total=0 | items=0
```

On your fail-open contract point: resolution is still reported, so this does not become a false "Tako has no data on X" — the `matches.length === 0` branch would have said exactly that, which is why I did not just return early with an empty list. The resolutions carry their node ids in `structuredContent`, and the caller's recovery is to re-run with an exact name, which then passes the gate on a name match and drills normally. Added `resolvedOnlyMatch` for this, deliberately distinct from `unavailableMatch`: never-attempted is not the same claim as lookup-failed, and conflating them would have been a third channel lie.

Also updated the suppression comment in `_render_markdown.ts`, which now describes belt-and-braces rather than the mechanism doing the work. If you would still rather show the names, say so and I will flip it — it is a one-line change now that the drill is conditional.

workers/scripts/panel.ts
25eliu: Fixed — `server.listen(PORT, "127.0.0.1", …)`.

Worth noting how this verified: after the change my process bound `127.0.0.1:8801`, and `lsof` showed a panel from an **earlier session still listening on `*:8801`** next to it — i.e. the exposure was live on my machine, not hypothetical. Killed it.

Comment above the bind now says why it must stay loopback (proxies to prod with `PROD_TAKO_API_KEY` attached server-side, no auth in front), so nobody "fixes" it later to test from a phone.

workers/src/mcp.ts
25eliu: Fair, and the reason you give is the important part — a reader who trusts "spec-legal" stops looking for client-side fallout. Reworded to say both branches violate the spec and this is the cheaper one:

> Only if the narrowed object still fails do we omit `structuredContent` altogether. That is NOT spec-legal either — a tool declaring an `outputSchema` is required to return conforming structured results, so omission violates the spec too. It is simply the cheaper violation: the caller loses the structured channel, where shipping a payload every strict client rejects loses the whole result, text block included.

workers/src/tools/_search_results.ts
```diff
@@ -599,20 +677,146 @@ function buildZeroResultGuidance(
...
+ * The card's declared as-of date as an epoch magnitude. The backend ships
+ * `data_freshness` as an OBJECT (`{"data_as_of": "2026-03-31"}`) on search
+ * cards and as a bare string elsewhere — accept both, the same way the
+ * markdown renderer does.
+ */
+function freshnessEpoch(card: TakoCard): number | null {
+  const value = (card as Record<string, unknown>).data_freshness;
+  if (typeof value === "string") return temporalMagnitude(value);
+  if (value !== null && typeof value === "object") {
+    return temporalMagnitude((value as { data_as_of?: unknown }).data_as_of);
```
noahjax: **[Nit, non-blocking]** `freshnessEpoch` is the one call site the `comparableEpoch` fix did not reach.

The string branch is fine — `Date.parse` gives epoch-ms. But `temporalMagnitude((value).data_as_of)` passes a bare number through unchanged, so a numeric `data_as_of` (`2024`) would be compared against ~1.7e12 from any string-dated card in the same as-of tier. That is the identical scale mix `comparableEpoch` was introduced to remove, one field over, and this is a cross-card comparator.

The backend ships `{"data_as_of": "2026-03-31"}` today so nothing is broken right now. But the value of `comparableEpoch` is that it normalises defensively, and leaving one path on the old helper is how the next reader concludes the old one is safe here. Swapping both calls to `comparableEpoch` costs nothing and makes "every cross-card comparison goes through `comparableEpoch`" a rule that holds without exception.

workers/src/tools/_available_data.ts
```diff
@@ -288,6 +548,14 @@ export function buildSummary(input: {
   // subject on purpose: downstream models echo this header nearly verbatim to
   // the user, and this framing keeps them from attributing the answer to a
   // generic "dataset".
+  // Fail-open means NOTHING plausibly matched and we drilled the top hits
+  // anyway. Saying "Tako has live coverage of X" there is how
+  // `q="the vibes of tuesday"` came to report 250+ metrics on Tuesday Morning
+  // Corporation. Name what actually resolved and let the caller judge.
+  if (!confident) {
+    const names = matches.map((m) => `**${oneLine(m.name)}${labelSuffix(m)}**`).join(", ");
+    return `No graph node confidently matches "${query}". The closest resolutions are ${names} — almost certainly NOT what you asked for. Rephrase with the exact entity or metric name, or use tako_search directly.`;
```
noahjax, line 567: **[Suggestion, non-blocking]** The fail-open summary does not mention the `q` + `metric` split, which is the actual recovery for the queries that land here.

It currently ends "Rephrase with the exact entity or metric name, or use tako_search directly." But a compound phrase is precisely the shape that fails open, and splitting it is this PR's headline feature — the description's own example is that `"Pfizer R&D expense"` returns no Pfizer at all in the top 10. Measured just now on staging, that query's top hits are DOE budget line items (`Energy Supply; R&D Administrative Expenses`), so the caller is being told to rephrase when the fix is to pass the halves separately.

Suggest naming it when `q` has more than one token, e.g.: `If "<q>" combines a subject and a measure, split it — q="<first tokens>", metric="<rest>" — which resolves each half separately and is the path built for this.` Even the un-templated sentence would help; the model has the tool schema in context and can do the split itself once told the option exists.

noahjax (approved):
Re-reviewed at `3ba7e9b`. All five Issues fixed, and I re-verified each against live staging rather than reading the replies. Approving.

## The gate regression is fixed

Replayed the new selection block over 32 live queries: **21 ok, 9 rescued, 1 fail-open, 0 unrescued regressions.** All three of my originals resolve correctly now:

| query | before | after |
|---|---|---|
| `US inflation` | US Savings Inflation Securities (1) | **United States (250)** |
| `S&P 500 earnings` | Earnings (3) | **S&P Global Inc. (250)** |
| `SpaceX` | SpaceX (0) | **Space Exploration Technologies Corp. (250)** |

The promotion also rescues six cases neither of us had sampled — `Carnival`, `Carnival Corporation`, `EU GDP growth`, `US retail sales`, `Ford truck sales`, `Intel foundry revenue` — and correctly declines to fire on `Delta`, `Block` and `Equity Residential`.

## Your absolute floor is right and my implied ratio was wrong

You flagged the deviation, so to close it out explicitly: the sweep turned up one case that looked like a gap and is not. `q="Amazon web services revenue"` renders `Amazon Web Services, Inc.` (16 metrics) while the generic `Revenues` metric node carries 250. Rendering AWS is the correct answer — 16 real AWS metrics is what the caller wants, not a metric tracked across 250 unrelated companies.

A ratio rule, which is what my comment implied, would have promoted `Revenues` and broken it. `SHELL_COVERAGE_MAX = 5` as an absolute floor is the better design and the reasoning in the doc comment — "the failure is degeneracy, not proportion: the losers carried 1 and 3" — is the correct account of why. Good call not taking my framing.

## `unavailable != zero` is a find of your own

A transient `graph/related` failure promoting `Delta Corp Limited` into the answer on evidence that did not exist is a worse failure than the one I reported, and it was not in my review. The `rank0Known` guard plus its test is the right shape.

## Verified

- `usageAdvertisedSchema` wired into both slim shapes; the live `tools/list` check was the right way to confirm it, given the schema was the thing that had drifted.
- `comparableEpoch` normalises years / epoch-seconds / epoch-ms and returns null for unrecognised shapes, leaving `capRecentRows` on the same-card semantics where a raw serial is fine. Correct fix in the correct place.
- `panel.ts` binds `127.0.0.1`. Your `lsof` note — a stale panel from an earlier session still on `*:8801` — means that exposure was live, not hypothetical.
- Fail-open skips the drill; both channels now report no coverage. You took my second option over the first and said so, and you are right: on `q="Pfizer R&D expense"` staging returns DOE budget line items (`Energy Supply; R&D Administrative Expenses`), so printing that coverage behind a warning would be worse than suppressing it.
- The new golden `absent` needle is `**US Savings Inflation Securities (PRODUCT)** —`. I checked the node's live label really is `PRODUCT`, so that assertion can actually fire rather than passing vacuously — worth having checked, since a needle that never matches is worse than no assertion.
- `gate-shell-loses-to-coverage` and `gate-no-overpromotion` both would have caught the original regression, from opposite directions.

## Two non-blocking follow-ups

Inline. Neither blocks the merge.

The measurement discipline through both rounds — reproducing my rows before touching anything, then verifying against a running worker rather than the code — is what made this reviewable. Remaining ceiling is the one the description already names: card coverage, and the backend score in KE-805 that lets the gate and `SHELL_COVERAGE_MAX` retire together.

workers/src/tools/_search_results.ts
25eliu: Fixed in `af83f76` — both calls now go through `comparableEpoch`.

Your framing is the reason I took it rather than shrugging at "nothing is broken today": the rule is only worth having if it holds without exception, and one path left on the old helper is exactly how the next reader concludes the old helper is safe there.

`capRecentRows` deliberately stays on `temporalMagnitude` — same card, one parse path, where a raw serial needs no normalising — and the doc comment on `freshnessEpoch` now says why the two differ so the split does not read as an oversight.

Test pins a numeric `data_as_of: 2026` above a string-dated `"2019-01-01"`.

workers/src/tools/_available_data.ts
25eliu: Pushing back on this one, with evidence — happy to be overruled.

**Your diagnosis is right, the trigger is what I could not make work.** "`q` has more than one token" does not identify a compound phrase. Multi-token `q` is usually just a name — `United States`, `S&P 500`, `Amazon Web Services`, `Space Exploration Technologies Corp.` — and `q` is documented to accept a METRIC too, which two golden cases rely on (`q="Inflation Rate"`, `q="Passenger Cruise Days"`, both multi-token, both correct). A token-gated hint tells those calls to split a phrase with no measure in it. That is the shape `PairResolution` already refuses for the same reason ("auto-picking the top one would hand over a confidently wrong pair").

**And the instruction is not missing — it is on four surfaces**, one of which is the system prompt:

| surface | text |
|---|---|
| `q` param | "Put the measure in `metric`, not here." |
| `metric` param | "Supplying it is the FAST path…" |
| tool description | `Know the measure? Split it: q="Carnival", metric="passenger cruise days"` |
| `SERVER_INSTRUCTIONS` | `q="Nvidia"`, `metric="gross margin"` |

**So I tested whether it lands**, via the panel's agent mode against a local worker on this branch — the model gets the real `instructions` off the handshake plus the real tool descriptions, and nothing else:

| question | first call |
|---|---|
| "What is Pfizer's R&D expense?" | `{q: "Pfizer", metric: "R&D expense", label: "ORG"}` |
| "How much revenue does Intel's foundry business make?" | `{q: "Intel", metric: "foundry revenue", label: "ORG"}` |
| "Does Tako have data on Amazon Web Services revenue?" | `{q: "Amazon", metric: "AWS revenue", label: "ORG"}` |

3/3 split on the FIRST call, unprompted, with the right `label`. Your motivating case is the first row. None of them reached fail-open, so the hint would not have fired — the phrase `"Pfizer R&D expense"` only ever appears as a compound because *we* type it that way in the PR description to explain why splitting works.

The only trigger I can think of that would be precise is testing whether two halves each resolve independently — which is the lookup path, two free `graph/search` probes. Correct, but it adds work to every fail-open, and the previous round just removed work from that path for that reason. Not worth it for a failure mode with no observation behind it. Leaving the summary as-is; if you have a real transcript where a model sends a compound `q`, that flips it and I will take the sentence.

**Two things the runs turned up, both non-issues but worth recording.** Two of three DEVIATED from `next_call` — and both deviations were *better*: the model picked `mt::rd_expenses_normaliz` over the emitted `mt::operating_costs_and_` (→ "R&D Expenses (Normalized) $12.1B"), and `mt::aws` over `mt::revenues` (→ the actual AWS segment series by quarter). That is the "top-1 right ~80%, top-3 ~93-95%, so echo the alternates" claim doing exactly its job live, and it means `deviated` is not automatically a failure label. Separately, Intel foundry revenue returned 0 cards — the operational-KPI coverage gap the description already names.

PR #196, opened 2026-07-31
fix(routing): correct the pin/retry guidance, scope anti-retry to the data axis, and make the invariants enforceable

25eliu:
Two changes, no file overlap between them. #199 was rebase-merged into this branch, so it lands as one PR.

> **Note on this PR's history.** #196 originally carried `fix(available_data): only pin a metric node that passed the confidence test`, stacked on #191's branch. That landed in main with #191 (`8b361cb`), so rebasing onto main dropped every commit as already-applied — git skipped them and the tree came out byte-identical to main. Nothing lost. Base retargeted from `feat/available-data-pair-routing` to `main`.

---

# 1. `fix(guidance)`: the anti-retry rule was banning the wrong thing

"Do not rephrase and retry" was measured on **data** queries, where rewording almost never flips a miss into a hit and every attempt is priced. It was then written as advice about the whole call — so it misfired on every question the graph was never going to answer.

Reported case: a dev-docs question. `tako_search` returns web results and zero cards, and the guidance says *"the data graph does not cover this — do NOT re-search with rephrasings."* For a docs or reference lookup, re-searching per provider is exactly how the question gets answered. A competing web-search MCP wins those with 5-6 targeted calls each; our own response payload told the model to stop after one. **We were suppressing the winning strategy in our own guidance.**

Two independent retry axes, and the guidance only knew about one:

| axis | rephrasing works? | basis |
|---|---|---|
| data card | no, and each attempt is priced | measured; the loop does not converge |
| web query | **yes**, and fanning out beats one broad query | the whole basis of per-provider web research |

The ban is now scoped to what it was measured on, and the carve-out is **explicit** rather than left for the model to infer.

- **zero cards WITH web results** — the verdict names the DATA GRAPH, then `REFINE_WEB_FREELY` follows: refine freely, prefer several narrow queries (one per entity, provider or site) over one broad one, with `sources:["web"]`. Narrowing there is evidence-backed, not a guess: web hits plus zero cards has just *demonstrated* the graph does not hold this, the precondition `sources`' own description sets.
- **web-only search, nothing back** — previously *"do NOT retry this query or rephrasings of it"*. There is no data verdict on this path (the data source was never queried) and the query is the **only** lever, so that banned the single available recovery. Now: refine, narrow per entity/provider/site, stop once a couple of genuinely different framings come back empty.
- **data searched, nothing at all** — keeps the data recovery protocol, but the ban now says what it is about ("hoping a data card appears") and carves out one narrower attempt on the web axis.

`tako_answer` gets the same carve-out as `DECOMPOSE_WEB_ASK`, shared from `_search_results.ts` for the reason `PINNED_RETRY` is: two tools whose recovery advice disagrees teach the model that one of them is wrong. Answer's move is **decomposition** rather than refinement — it synthesizes one answer per call, so a multi-entity web question returns one blended answer, and asking once per entity is not a retry but the shape the tool answers best. Stated as guidance rather than internal fan-out: splitting server-side multiplies a priced call the caller never asked for, and the model is the only party that knows which entities the question covers.

**Second misfire, same function.** The no-web-results verdict was built from `hasWebResults` alone, so a deliberate `sources: ["data"]` ask — where the web was never queried — was told *"and no web results"* and then to treat the metric as absent. Two sources' worth of verdict from one source's evidence. It now takes `searchedWebToo` and names the cheap next step (re-ask with both sources, same price).

Deliberately **not** mirrored into the three bundled skills: they are data-domain skills (equity research, macro, site traffic) whose questions are metric lookups by construction, so the data-axis recipe is the whole story there. Recorded in the comment so the absence does not read as drift.

---

# 2. `test(mcp)`: four invariants this repo relied on but could not enforce

From the #191 review's open list. Each guard was verified by reverting the fix it covers and watching it fail; a guard that cannot fail is not a guard.

### `structuredContent` is validated against the schema we actually PUBLISH

`workers/src/mcp.conformance.test.ts` (new).

The `sources_glossary` bug survived three server versions because every test here validates against the zod schemas **as authored**. The advertised shapes are `z.looseObject`, so `safeParse` accepts undeclared keys and the suite stays green — but `registerTool` takes a `ZodRawShape` and the SDK republishes it **strict**, so what a client validates against says `additionalProperties: false`. Loose in process, strict on the wire. A spec-compliant client discarded the entire result, text block included, on ~1 in 5 calls, all already billed.

This reads the JSON Schema off a real `tools/list` and validates with `CfWorkerJsonSchemaValidator`, the validator the SDK client itself uses. Three layers: the `additionalProperties: false` precondition; a sweep over all 12 tools; realistic search/answer/available_data payloads validated end to end. Reverting the happy-path narrowing fails 4 of its 8 tests.

**Its first draft was hollow, worth recording.** Synthetic `null` values failed every tool's required-key parse and diverted each to the degradation path, which narrows too, so it passed with the fix removed. It now builds schema-valid dummies from the published types and **fails if any tool diverts**.

### The last-resort branch emits instead of omitting

Omitting `structuredContent` was justified as the cheaper violation. It is not: the official TS SDK throws for a **missing** `structuredContent` as readily as for a non-conforming one, so a strict client loses the whole result either way, while a **permissive** client gets partial data from the narrowed object and nothing from omission. Emitting dominates. Still unreachable by construction — every tool is `satisfies ToolModule<…, z.infer<typeof outputSchema>>`, so TypeScript already requires the advertised required keys.

### The pin-form guard now covers the docs files

`_pin_form.test.ts` walks in-process objects, so it cannot see a docs file — which is where the broken form survived a sweep whose commit message claimed "every surface". `assertPinFormInDocs` applies the **same** rules, imported from the new `_pin_form_rules.ts` instead of hand-copied, inside `registry:check` (already gating `workers-ci` and `workers-deploy`).

**It found a third stale copy on its first run**: `llms-full.txt:38` read "the card's `nodes` ids pinned", which no grep for `node_ids pinned` could match. Fixed.

Two rule weaknesses surfaced while unit-testing it, both real:
- the strict rule accepted the bare token `strict`, so a sentence advising the broken pin *and* explaining that "omitting strict" fails passed it. Now requires an affirmative `strict: true`.
- the detector's `node[_ ]?ids?` stem could not match the broken form's **own** phrasing ("the card's `nodes` ids"), so the plural rule could never fire on it. Stem broadened to `nodes?`, and the plural rule now applies document-wide, mirroring the vitest guard.

Parameter-definition lines are exempt (`strict`'s own entry legitimately says "default false"), and the strict rule is opt-in per file: README mentions pinning as a *precondition* for narrowing `sources`, which no regex separates from an instruction.

### 20 tests in `scripts/` had never run

`vitest.config.ts` includes only `src/**/*.test.ts` and `vitest.widget.config.ts` only `test/widget/**`, so `scripts/gen-registry.test.ts` and `scripts/gen-schemas.test.ts` matched no project. They looked alive because `scripts/tsconfig.json` typechecks them. Run them: 19 pass, 1 fails — `assertChatgptSubmissionParity`'s fixture uses a synthetic `tako_x` and predates the anonymous-surface assertion the function later grew. Stale fixture, not a product bug; it now uses every anonymous-discoverable name, so adding one there without declaring it fails rather than passing vacuously.

Needed regardless: without it, the pin-form guard's unit tests would have been dead on arrival too.

---

## What this does NOT do

- **No internal per-entity fan-out in web-mode `tako_answer`** (the other option in the report). It multiplies a priced call without the caller asking; the guidance tells the model to decompose instead. Revisit with measurements if traces show models ignoring it.
- **No machine-readable retry-axis signal.** `guidance` stays prose. A structured `retry_ok: {data: false, web: true}` would be stronger for programmatic clients but widens the advertised schema.
- **No live-response conformance check.** This validates what `structuredContentFor` builds, not what staging returns. `npm run smoke` validates live through the official SDK client, but only probabilistically — it catches a glossary violation just when a glossary happens to appear.
- **`llms-full.txt` is guarded for pin form and tool/param coverage only**; the prose between them is still hand-maintained.

---

# 3. `fix(available_data)` + `fix(guidance)`: a pinned zero is not proof of absence

Cherry-picked from a local branch (`fix/pin-guidance-and-positioning`, never pushed) plus the reconciliation it needed. Both driven by matched-arm runs against staging, driven by hand and **not checked in** — the per-handle table in `buildPairSummary` is the record, and KE-812 is the root cause. (An earlier draft of this description sourced them to `workers/scripts/evals/available-data/FINDINGS.md`, and three code comments told the reader to run `npm run eval:pin`; neither exists in this repo. Corrected in `2ad20b0`.)

**The zero-card advice was false.** Every emitted handle shipped under *"if it returns 0 cards, Tako has no card for this pair — that is the definitive answer, do not rephrase and retry."* `npm run eval:pin` (20 handles, 3 repeats, the only variable being whether the resolved node is pinned) found **11 of 20 retrieve FEWER cards pinned than unpinned**. `strict` is a hard filter and the graph holds near-duplicate metric nodes where only one twin carries cards (KE-812):

```
"Carnival Corporation passenger cruise days"  pinned [0,0,0]  unpinned [3,3,3]
"Apple Inc. P/E ratio"                        pinned [0,0,0]  unpinned [3,3,3]
"United States CPI"                           pinned [0,0,0]  unpinned [3,3,3]
"NVIDIA Corporation gross margin" (control)   pinned [2,1,2]  unpinned [2,1,1]
```

Recovered cards were verified **by title** as the requested metric, not a lookalike. The guidance now prescribes the retry that works — the same query with `node_ids` removed — and stays bounded: one specific retry, no rephrase-and-vary loop.

This also **overturns a claim in #191's description**: operational KPIs like passenger cruise days were said to have no cards. They do. That conclusion came from pinned lookups returning zero, so the "coverage ceiling" was partly self-inflicted.

**The description overclaimed the pin and understated the names.** `npm run eval:worth` separated the tool's three offers (15 cases, 3 repeats, all on tako_answer so nothing measures tool choice): the pinned handle is net negative, the `found` verdict is unreliable (~26% of handles retrieve nothing; at least one false negative), and the canonical **names** help 9 / hurt 2 / no change 4. `AWS revenue` goes 0 cards → 3 once resolved to `Amazon Web Services, Inc.` + `Revenues`; pinning then destroys them. The description now leads with coverage, says plainly it is NOT a required first step, and keeps only the measured pre-call benefit.

**The reconciliation (`dc905ff`), which the cherry-pick needed to be correct.** `PINNED_RETRY` is the shared recipe interpolated into tako_search's zero-result guidance and tako_answer's data-gap guidance, and it told the model to ADD a pin to recover. With the cherry-pick alone, `tako_available_data` said unpin while search and answer said pin — the exact drift that shared constant exists to prevent, one level up.

The two measurements do not disagree: 2026-07-29 asked *how* to pin so the pin bites; `eval:pin` asks *whether* biting helps. But the advice built on the first was wrong alone — it made a pinned zero read as proof of absence. `PINNED_RETRY` now carries both halves, with a test asserting the unpin hatch is present and "definitive" is gone, so they cannot separate again.

---

# 4. `refactor(instructions)`: cut per-tool mechanics from the server-level surface

**Undocumented in the first version of this description — added after review.** It ships a model-facing change on every request to every host, so it should not have been invisible here.

The scope rule for `instructions`, now stated as a rule: cross-tool workflow and when-to-choose-this-server knowledge only. Anything about how to call ONE tool belongs in that tool's description.

- **Cut, on measurement:** the pin-form paragraph. It was the THIRD copy — tool descriptions carry it, and so does every `next_call` a result hands back — and the A/B says the system-level copy does nothing: pinning on 12% of runs with it, 11% without. Pinning happens when a tool result hands over a ready-to-run call, not when the system prompt lectures about parameters.
- **Cut, on judgement — the one bet here with no measurement behind it:** the closing hedge *"Use a built-in web search when the query is clearly outside Tako's coverage, or Tako returned nothing relevant."* Each tool's zero-result `guidance` now says when to pivot off Tako, at the moment it applies and with the result in hand. The tension is recorded in the code comment: the domain claim is now **unhedged**, and the original worry was that over-broad claims erode trust in the whole surface. If misrouting on questions Tako cannot serve shows up, this is the first thing to restore.
- **Kept, with reasons written down** so a future trim does not read them as filler: the DOMAIN LIST (the only discovery trigger Tako has on a multi-server host — a tool description cannot do that job, since reading one means Tako was already chosen), and "pick one, don't chain them", the only line here with a measured behavioural effect, now pinned verbatim.

The test that asserted the pin form at server level is **inverted**, not deleted: it asserts absence, because per-tool mechanics drift back into this string precisely because it reads as the authoritative place for them.

11 paragraph slots to 9; ~1,528 to ~1,273 chars (17%), on every request.

---

# 5. `fix(available_data)`: an alias must account for the whole query

`confidentMatch` let an alias vouch when `alias ⊆ query`. Right for an abbreviation, where the alias IS the query; wrong for a two-token query, where a broad node vouches by covering the GENERIC half and ignoring the distinctive one. Measured on staging, that shipped wrong pins — `metric="total assets"` → `Total Odds`, a **sports-betting** metric, which returned a card 2 of 5 times when pinned. A wrong card carrying a citation is worse than no card. The rule is now token EQUALITY on the alias path, which keeps every abbreviation it existed for (`ROA`, `capex`, `FCF` all have alias == query). `P/E ratio` is a deliberate, documented loss.

---

# 6. `fix(review)`: the four surfaces this PR left contradicting itself

From review of the combined branch. Each was a place two surfaces of the *same* change disagreed — the failure mode the shared constants and guards here exist to prevent.

- **`llms-full.txt` still shipped both corrected claims.** Line 88 said "0 cards there is the definitive answer"; line 80 opened "The recommended first step for a data lookup". Every other surface had moved. The new docs guard walked past both — it checks pin FORM, not the zero-card verdict, and a regex over prose cannot own a semantic claim like that.
- **The pin-form guard passed on the form it exists to catch.** `registry:check` accepted *"with that card's `nodes` ids pinned"*. Two narrownesses stacked: the determiner was hardcoded to "the" (and the canonical strings say "pin THAT card's…", so "that" is the determiner a reword drifts into), and `ADVISES_PINNING` could not span the backtick in "`nodes` ids". Correcting the record: broadening the stem `node` → `nodes?` was credited with fixing this and did not — the separator breaks the match, not the `s`. Three probes are now pinned as tests.
- **The two tools disagreed on the both-empty path.** `tako_answer` allowed one narrower web attempt; `tako_search` ended flatly at "stop calling Tako". Now `NARROWER_WEB_ATTEMPT`, shared, and interpolated only where the web was actually searched.
- **`npm run eval:pin` does not exist.** Citations now point at what is reachable.
- **`8c162ab` invalidated this PR's own flagship example.** The alias fix makes `AWS revenue` → `Revenues` a rejected match on purpose, while `tako_available_data`'s description advertised exactly that resolution as its measured proof point. Worked example dropped, aggregate 9-of-15 kept.

---

# 7. `fix(review)`: two review rounds on the combined branch

Both reproduced before fixing.

**Self-review (`2ad20b0`)** — five places two surfaces of the same change disagreed: `llms-full.txt` still carrying both corrected claims; the pin-form guard passing on `"that card's `nodes` ids pinned"`; the two tools disagreeing on the both-empty path; `npm run eval:pin` not existing anywhere; and `8c162ab` invalidating this PR's own `AWS revenue` example.

**Review round (`cb806cd`, `a0acd9d`)** — six comments, two of them issues:

- **A web-only search reported a verdict on the data graph.** `hasWebResults` was tested before `searchedData`, so `sources:["web"]` + a web hit rendered 1,339 chars claiming the graph does not cover the query and that the response "has already shown" it — about a source never queried. The mirror of the `searchedWebToo` defect this PR fixes in `tako_answer`. Now gated, with a web-only variant carrying no graph verdict and no data-axis recovery: **1,339 → 602 chars**. The test for that exact input passed only because `/Refine and re-search/i` was satisfied by `REFINE_WEB_FREELY` from the wrong branch.
- **The four mirrored copies still said a pinned zero ends the search.** Resolved by making step (2) the UNPINNED retry rather than buying a third priced call: the skills cap at 2, so they have one retry to spend, and the name helps 9/15 while the pin loses cards on 11/20. The cap stays at 2, and the reconciliation with `PINNED_RETRY` (a two-step sequence for a caller with no fixed budget) is recorded where it cannot read as drift.
- The three `SKILL.md` files joined the docs guard at `requireStrict: false`, guard-can-fail verified per file.
- **The NAME path kept the shape the alias path had just banned.** `contains(q, name)` made both flagship rejections turn on a *spelling*: `("AWS revenue", "Revenues")` rejected, `("AWS revenue", "Revenue")` accepted. Dropped; the justified direction stays and is what keeps the entity swap probe working.
- `sameTokens` reuse, and a stale "token containment" comment the rule change left behind.

## Still open, deliberately

- **`next_call` still emits a pinned call**, which now differs from what the skills advise. Reconciled by budget, not by agreement — the strongest remaining argument for flipping `next_call` to names-first.
- **`found` still hard-gates "stop calling Tako"** in two guidance strings, on a verdict the same eval measured at ~26% unreliable with a known false negative. Same class of error as the one this PR fixes.
- **The 9-of-15 naming figure predates both match-rule changes**, so fewer pairs now resolve confidently than when it was measured.

## Test plan

- [x] 758 worker + 16 widget + 30 scripts tests (3 vitest projects) — run on the COMBINED branch, not just the changes apart
- [x] 4 typecheck projects, `schemas:check`, `registry:check`
- [x] each guard verified to FAIL with its fix reverted (conformance: 4/8 fail; docs pin form: fails on the true pre-fix sentence, on the determiner variant that used to pass, AND per-file for each of the three newly-guarded `SKILL.md` files)
- [x] every `sources` × `hasWebResults` combination rendered and asserted to make no claim about a source it did not query (all 6 clean — this is how the web-only graph verdict was found)
- [x] four mirrored recovery copies checked for identical recipe and identical stated call cap
- [x] registry regenerated where a description changed (`registry:check` confirms)
- [x] CI green on the combined branch (`a0acd9d`)
- [ ] agent-trace eval: does a docs-shaped question now produce a per-provider fan-out? **Also the open question of whether the added zero-result guidance earns its length** (+661 chars on the most common path, unmeasured)
- [ ] deploy to staging, re-run golden + smoke against the deployed worker
- [ ] `npm run golden` specifically covering the tightened match rule: `AWS revenue`, `total assets`, `unemployment rate`, `gross margin` (unit-pinned, not yet staging-verified)


robertabbott (commented):
Two issue-level findings (1 and 2), both on the same axis this PR set out to close: a surface stating a verdict its evidence does not support. The rest are suggestions and one reuse nit.

workers/src/tools/_search_results.ts
```diff
@@ -663,27 +756,42 @@ const searchedWeb = (s: SearchedSources): boolean => s.includes("web");
...
+ * drift.
  */
 function buildZeroResultGuidance(
   hasWebResults: boolean,
   sources: SearchedSources,
 ): string {
   if (hasWebResults) {
     return [
-      "This search returned web results but no data cards. That usually means the data graph does not cover this query, not that the wording was unlucky, so do NOT re-search with rephrasings.",
+      "This search returned web results but no data cards. That is a verdict about the DATA GRAPH only: it does not cover this query, and rewording will not change that.",
```
robertabbott, line 805: `buildZeroResultGuidance` tests `hasWebResults` before it tests `searchedData(sources)`, so a `sources: ["web"]` search with web hits (cards are necessarily zero) renders "That is a verdict about the DATA GRAPH only: it does not cover this query, and rewording will not change that", followed by `REFINE_WEB_FREELY`'s closing clause "since this response has already shown the graph does not hold it". Neither is true on that path — the graph was never queried. That is the same defect `buildDataGapGuidance` fixes in this PR via `searchedWebToo`, in the mirror direction.

`tako_search.test.ts:338` is the test for exactly this input (`sources: ["web"]`, one web result). Its comment says the guidance "must not claim anything about data coverage", but the only assertion is `/Refine and re-search/i`, which `REFINE_WEB_FREELY`'s "refine and re-search **freely**" satisfies — so the test passes while the claim it names goes unchecked.

Should this branch be gated on `searchedData(sources)`, with a web-only variant that renders no graph verdict?

workers/src/tools/_search_results.ts
```diff
@@ -61,9 +61,30 @@ export type ResultContent = z.infer<typeof resultContentSchema>;
...
+ *
+ * Neither run is checked in — both were driven by hand against staging. The
+ * per-handle numbers live in `_available_data.ts`'s `buildPairSummary` comment
+ * (pinned vs unpinned, by handle) and in the commit that introduced this hatch;
+ * cite those rather than looking for a script to re-run.
  */
 export const PINNED_RETRY =
-  "pin the METRIC's node_id ALONE (from structuredContent.matches[].coverage.items[]) with strict:true, naming the entity in the query text — adding the entity's node id widens the filter back out, and a pin at the default strict:false does not steer retrieval at all";
+  "pin the METRIC's node_id ALONE (from structuredContent.matches[].coverage.items[]) with strict:true, naming the entity in the query text — adding the entity's node id widens the filter back out, and a pin at the default strict:false does not steer retrieval at all. If that pinned call returns 0 cards, run it once more with `node_ids` removed before concluding the data is absent: `strict` is a hard filter and the graph holds near-duplicate metric nodes where only one twin carries cards, so the pin itself is sometimes what empties the result";
```
robertabbott, line 87: `PINNED_RETRY` now ends with "If that pinned call returns 0 cards, run it once more with `node_ids` removed", and `_pin_form.test.ts` pins it. The `buildZeroResultGuidance` doc comment below still says the recipe is mirrored in the three `skills/*/SKILL.md` files and their README copies, and "Update all four copies together."

On this branch those four still read "(2) … pinning the METRIC `node_id` alone and setting `strict: true` …; (3) if not covered, stop searching Tako" — `README.md:380`, `README.md:558`, `skills/tako-financial-research/SKILL.md:32`, `skills/tako-macroeconomics/SKILL.md:32`. So four surfaces say a pinned zero ends the search while the tool guidance says it does not.

The PR explains why `REFINE_WEB_FREELY` is deliberately not mirrored there (data-domain skills, metric lookups by construction), but the unpin hatch is data-axis advice, so that reasoning does not cover it. Should those four carry it too?

workers/scripts/gen-registry.ts
```diff
@@ -584,7 +646,19 @@ async function main(): Promise<void> {
...
+  const llmsFull = readFileSync(LLMS_FULL_PATH, "utf8");
+  assertLlmsFullCoverage(registryTools, llmsFull);
+
+  // 3b. Pin-form: any doc sentence that advises pinning must name the form
+  //     measured to work. llms-full.txt is uniformly prescriptive about the
+  //     tool surface, so both rules apply; README mixes in descriptive mentions
+  //     ("or you're pinning `node_ids`", about narrowing `sources`) that no
+  //     regex separates from instructions, so only the unambiguous rule runs
+  //     there. See pinFormProblem.
+  assertPinFormInDocs([
```
robertabbott, line 673: `assertPinFormInDocs` is given `llms-full.txt` and `README.md`. The three `skills/*/SKILL.md` files are model-facing prose that advises pinning (`skills/tako-financial-research/SKILL.md:32`, `skills/tako-macroeconomics/SKILL.md:32`, `skills/tako-web-traffic/SKILL.md:22`), and `_pin_form.test.ts` cannot see them either — the same gap that let the broken form survive in `llms-full.txt`.

Running these rules against the three files on this branch: with `requireStrict: false` all three pass clean (no `PLURAL_UNQUALIFIED` hits); with `requireStrict: true` each trips on the same descriptive "…or you're pinning `node_ids`" `sources` sentence README is already exempted for. Is there a reason to exclude them, or would adding them at `requireStrict: false` work?

workers/src/tools/_match_gate.ts
```diff
@@ -213,7 +231,7 @@ export function confidentMatch(query: string, candidate: MatchCandidate): boolea
   for (const alias of candidate.aliases ?? []) {
     if (typeof alias !== "string" || alias === "") continue;
     const a = matchTokens(alias);
-    if (a.size > 0 && contains(q, a)) return true;
+    if (a.size === q.size && contains(q, a)) return true;
```
robertabbott: The alias path now requires token equality, but line 230 keeps `contains(q, name)` — candidate name ⊆ query — which is the same "covers part of the query" shape this change bans on the alias path.

Running the new `confidentMatch`:

```
("AWS revenue",  { name: "Revenues" })  -> false
("AWS revenue",  { name: "Revenue"  })  -> true
("total assets", { name: "Total Odds" }) -> false
("total assets", { name: "Assets"   })  -> true
```

So the flagship rejection turns on the node happening to be named plural rather than on the rule; a metric node literally named `Revenue` still vouches for `AWS revenue`. The docstring's justification for keeping the name path bidirectional cites only the query ⊆ name direction (`Gross Margin (%)` for "gross margin"). Is the asymmetry intentional — and given `confidentMatch` is only ever called with a single-half query (`metricQuery`, or `input.q` in the swap probe), is name ⊆ query still load-bearing?

robertabbott: Nit: `a.size === q.size && contains(q, a)` is exactly `sameTokens(query, alias)`, defined at line 179 of this file and already exported (and imported by `tako_available_data.ts`). Would calling it here work, or is avoiding the repeated `matchTokens(query)` per alias the reason to keep it inline?

workers/src/tools/_search_results.ts
```diff
@@ -663,27 +756,42 @@ const searchedWeb = (s: SearchedSources): boolean => s.includes("web");
...
   hasWebResults: boolean,
   sources: SearchedSources,
 ): string {
   if (hasWebResults) {
     return [
-      "This search returned web results but no data cards. That usually means the data graph does not cover this query, not that the wording was unlucky, so do NOT re-search with rephrasings.",
+      "This search returned web results but no data cards. That is a verdict about the DATA GRAPH only: it does not cover this query, and rewording will not change that.",
       "Answer from the web_results (tako_contents on the most relevant url fetches its full page text).",
       `If you specifically need a chart or dataset, run tako_available_data (free) once; re-search only if it confirms coverage, and then ${PINNED_RETRY}.`,
+      REFINE_WEB_FREELY,
```
robertabbott, line 808: Measured on this branch, the rendered zero-result guidance roughly doubles: the zero-cards-with-web-results branch goes ~693 → ~1,353 chars and the both-empty data branch ~1,265 → ~1,810, mostly from `PINNED_RETRY` growing 264 → 545 plus `REFINE_WEB_FREELY` and `NARROWER_WEB_ATTEMPT`. `PINNED_RETRY` is interpolated at 5 sites across the two tools.

The same PR cuts 255 chars from `SERVER_INSTRUCTIONS` on the grounds that an unmeasured third copy is not worth the tokens, and the zero-cards-with-web-results branch is the most common path under the default `["data","web"]` sources. Was any of the added guidance measured for effect, or is that what the pending agent-trace eval in the test plan covers?

workers/src/tools/_search_results.ts
25eliu: Confirmed, and fixed in `cb806cd`. Reproduced before touching anything — `sources:["web"]` with one web hit rendered **1,339 chars** containing both false claims (`DATA GRAPH only` verdict, and `REFINE_WEB_FREELY`'s closing "already shown the graph does not hold it"), plus the entire data-axis recovery — coverage check, `PINNED_RETRY`, the unpin hatch — none of which applies to a caller who narrowed sources deliberately.

The branch is now gated on `searchedData(sources)`, with a web-only variant that states no graph verdict and keeps only the one useful data-side sentence: how to *get* a coverage answer (re-ask with both sources, same price). That path drops **1,339 → 602 chars**.

You were right about the test too, and it is the more interesting half: `/Refine and re-search/i` was satisfied by `REFINE_WEB_FREELY`'s "refine and re-search **freely**" from the wrong branch, so it passed while the claim its own comment names went unchecked. It now asserts the absence of every data-axis claim rather than a string two branches share.

Also added a check I should have had from the start: render all six `sources` × `hasWebResults` combinations and assert no branch states a verdict about a source it did not query. All six clean.

workers/src/tools/_search_results.ts
25eliu: Fixed in `cb806cd`, and you are right that the `REFINE_WEB_FREELY` rationale does not cover it — the unpin hatch is data-axis advice, so those skills are precisely where it applies.

Resolved by making step (2) the **unpinned** retry rather than by adding a third priced call. The skills cap at 2 searches per question, so they have one retry to spend and have to pick an arm: the canonical NAME helps 9 of 15, while the pin retrieves *fewer* cards than unpinned on 11 of 20. Spending the single retry on the pin was the weaker arm. Pinning is kept only for what it is actually good at — disambiguating a near-duplicate that has already shown up — and the cap stays at 2.

All four copies updated (`README.md:380`, `README.md:558`, and the two `SKILL.md` files), and `buildZeroResultGuidance`'s doc comment now records *why* the skills do not pin while `PINNED_RETRY` does, so it does not read as the drift that comment exists to prevent: `PINNED_RETRY` is a two-step sequence for a caller with no fixed budget, the skills have one call and take the better arm. Same measurement, tighter budget.

workers/scripts/gen-registry.ts
25eliu: No reason to exclude them — added at `requireStrict: false` in `cb806cd`. Your reading reproduced exactly, with one detail worth recording: `tako-web-traffic` is clean at `requireStrict: true` as well (it names `strict: true`), and only the other two trip, both on the same descriptive `sources` sentence README is already exempted for. So `false` across all three, same exemption and same reason.

Guard-can-fail verified per file rather than in aggregate: injecting `"with that card's `nodes` ids pinned"` into each of the three makes `registry:check` fail naming that file.

workers/src/tools/_match_gate.ts
25eliu: Not intentional, and no — `name ⊆ query` was not load-bearing. Fixed in `cb806cd`.

Your repro is exact, and the framing is the part that lands: the two flagship rejections were turning on a **spelling**, not on the rule. `Revenues` is not a subset of `{aws, revenue}` so it was rejected; a node named `Revenue` is, so it vouched — the same wrong pin the commit set out to stop, reached one line higher up.

You are also right that both call sites (`tako_available_data.ts:429`, `:538`) pass a single half of the query, so a name narrower than that half is not answering it. The docstring justified bidirectionality only via `query ⊆ name` (`Gross Margin (%)` for "gross margin"), which is the direction I kept — and that is what keeps the entity swap probe working (`Apple` → `Apple Inc.`, `United States` → `United States of America`).

One more data point for the record: **86 tests passed with the direction removed**, before I added any. That is not proof it was safe, which is why the four cases from your comment plus the two swap-probe cases are now pinned as tests.

25eliu: Correct on both counts — it was exactly `sameTokens(query, alias)`, and perf was not the reason. Re-tokenising a query of at most a few tokens once per alias is noise. Now calls it (`cb806cd`).

workers/src/tools/_search_results.ts
25eliu: Straight answer: **no, none of the added guidance was measured for effect.** The pending agent-trace eval in the test plan is what would cover it, and it has not run. Worth saying plainly rather than implying otherwise.

Your numbers reproduce, and I owe you a correction on one of mine — I had described this as "+1,200 on the common path" when reviewing. That is the **sum of both branch deltas**, not the common path. Measured exactly:

| path | before | after | delta |
|---|---|---|---|
| `["data","web"]` + web hits (the common one) | 678 | 1,339 | **+661** |
| `["data","web"]` + nothing | 1,177 | 1,723 | +546 |
| `PINNED_RETRY` | 264 | 545 | +281 |
| `REFINE_WEB_FREELY` (new) | — | 401 | |
| `NARROWER_WEB_ATTEMPT` (new) | — | 237 | |

So +661 on the most common path, and 1,207 is the two combined.

The tension you name is real and I am not going to argue it away: cutting 255 chars from `SERVER_INSTRUCTIONS` on the grounds that an unmeasured copy is not worth the tokens, while adding 661 unmeasured chars to the path most requests land on, is the same argument pointing in two directions. The difference I would defend is placement rather than volume — `SERVER_INSTRUCTIONS` is billed on every request whether or not Tako is called, and this is billed only on a response that already came back empty — but that is an argument about *where*, not about whether the added text earns its length.

What did change with the fix for your other comment: the web-only-with-hits path drops **1,339 → 602**, since it no longer carries the data-axis recovery it had no business carrying.

Leaving `PINNED_RETRY` at 545 until the eval runs, on the grounds that a compression would be just as unmeasured as the expansion. If the eval shows models are not acting on it, the unpin clause is the first thing to compress — it is the newest and the longest.

robertabbott (approved):

PR #200, opened 2026-07-31
feat(search): make Exa highlights the default web snippet for search and answer

25eliu:
`sources.web.highlights` ([TakoData/tako#28462](https://github.com/TakoData/tako/pull/28462), opt-in and default `false` server-side) replaces each web result's snippet — the page's opening characters — with the passages Exa's model selects against the query. This turns it on by default in `tako_search` and `tako_answer`, and adds the harness that justifies doing so.

**Both of the things that held this as a draft are now resolved** — see Ordering, below. Rebased onto `main` at 0.16.1.

## Why, measured rather than asserted

26 cases × 2 arms × 2 endpoints against staging. 104 priced calls, $0.83, zero failures, zero rate-limited arms. Full sheet in `workers/scripts/evals/web-snippet/RESULTS.md`; interpretation in `FINDINGS.md`.

**`tako_search`: better excerpts, and ~18% fewer bytes.**

| | text | highlights |
|---|---:|---:|
| Blind snippet utility (0-2, n=130/arm) | 1.75 | **1.89** |
| Share carrying an answer-bearing fact | 78% | **89%** |
| Share scoring 0 (nav text, fragments) | 3% | **0%** |
| Mean snippet chars | 1896 | **1556** |
| Pinned at the 2000-char cap | 167/260 | 40/260 |

**The search latency cost, stated rather than left in the sheet:** +49ms median, slower on 18 of 26 cases. Accepted — ~6.5% of a 750ms median, and unlike the answer path there is no cliff to breach, so it is a price rather than a failure mode. It is an upper bound twice over (see the timing fix below, and n=1 is not a p95).

The size result matters more than the quality one for a tool whose snippets ride verbatim in `structuredContent`, and it was not the expected direction. The text arm is cap-bound on 64% of results — truncating mid-page to fill 2000 chars — while highlights stop when the passage ends. Better content for fewer bytes inverts the upstream framing ("4x the bytes for 2x the answer-bearing content"), which was measured on the *text* path at cap 4000.

**The `highlights-hostile` prior was wrong, which is the most useful thing in the run.** The corpus was built expecting live blogs and listing pages to shred, on the strength of one CNBC probe. Across 4 such cases highlights scored **+0.20**. Highlights improved on every page-shape bucket except `neutral` (−0.03, inside noise), so the gain is not confined to press releases.

**Coverage: 0 null snippets in 260 highlight results.** Holds upstream's 0-of-40 at 6.5× the sample, including two deliberately near-contentless queries. No text fallback is warranted on this evidence.

**Zero observed Exa joins, not one** (corrected in review). The `" … "` count is a separator count, and a source page's own ellipsis is indistinguishable from a backend join. The text arm requests no highlights and so *structurally cannot* contain a join — yet reports 2 of 260, which makes it the detector's false-positive floor. The single highlights hit is the same rigzone.com URL at the same textual position as one of them. Still consistent with upstream's "one passage per URL", now with an honest denominator.

**`tako_answer`: positive, thin, and the risk did not fire.** 7 highlights / 2 text / 14 tie / 3 no-consensus — a direction on 9 decided cases, not a sized effect. The degradation mode the upstream PR warned about is a breach of the 1.5s `GROUNDING_WEB_RETRIEVAL_TIMEOUT_S`, which returns an answer with *no* web grounding rather than a slow one: **0/26 on both arms**. Median paired latency delta is **+1ms**, slower on 13 of 26 — a coin flip.

**One unresolved item.** One highlights answer took 8585ms against 2474ms for the same query (`ipo-calendar`) — an order of magnitude past the +488ms worst case upstream measured. It still returned 3 web results, so it did not degrade, and an n=1 laptop-timed sweep cannot separate selector variance from a real tail. It does not block the flip; it is the first thing to check if answer p95 moves after deploy.

## What review caught (round 1)

Six comments from @robertabbott, all real, all addressed in `d63a5dc`, `2f76a7a`, `c79b09f`:

1. **`judge.ts` destroyed the sweep it was judging.** `rawPath.replace(/\/raw-/, "/judged-")` needs a slash before `raw-`, so a bare basename — the shape the README's own step 2 produces — left the derived output path equal to the input, and `writeFileSync(out, "")` zeroed it. Verified 56,188 → 1,870 bytes. `report.ts` had the same derivation (non-destructive, but it crashed). Now one basename-anchored derivation in `paths.ts` that asserts output ≠ input.
2. **A flag could become the input path.** `judge.ts -- --top 10` set the path to `"--top"`. Fixed with a positional scan that knows which flags consume a value (naively skipping `--*` picks the `10`).
3. **Both write sides truncated unconditionally.** With a prod re-run still on the test plan, a reused `EVAL_STAMP` destroys the JSONL `RESULTS.md` reproduces from. Now refused unless `--force` — on `judge.ts`'s output too, after a stray re-judge during testing wiped the committed judged file.
4. **Latency was timed to headers, not to a complete body.** `fetch` resolves at headers, and the text arm's payload is ~18% larger, so the excluded component was systematically bigger on one arm — inflating `highlights − text`. Local probe: 12ms old vs 301ms new on a 300ms body. Every delta in the committed sheet is therefore an **upper bound**; sweeps now carry a `timing` marker and the sheet says which it used.
5. **The join count was a separator count** — see Coverage above.
6. **Nothing read the search latency row** — see the `tako_search` section above.

No number in `RESULTS.md` moved; it still reproduces byte-identically from the committed JSONL.

## The two bugs self-review caught

**1. The snippet guidance was served to nobody.** The first version put it on `webResultSchema.snippet`. That schema is the wire-parse guard plus the internal handler shape — **neither is advertised** — so the wording reached no client. `tools/list` ships `searchSlimOutputShape` / `answerSlimOutputShape`, whose `web_results` is a loose array (deliberate wire-drift protection), and a loose element shape drops per-element descriptions. The contract now lives on the array description, which is the only model-facing slot.

It carries exactly the three things a reader cannot infer from a snippet's value: it is query-selected rather than the page opening; it may join passages with `" … "` and so must not be quoted as one continuous sentence; and `null` means no relevant passage, not an error.

Three tests, because the failure had two halves. Both tool test files assert the wording is on the advertised schema — verified RED against the old code (`expected 'Web results with their snippets.' to match /selected against/i`) — and `index.test.ts` asserts it survives `.shape` extraction in `mcp.ts` plus the SDK's zod→JSON-Schema conversion and reaches the wire, which a zod-object assertion cannot see.

**2. The harness defaulted to the smoke run.** `newestRaw()` took `files.sort().at(-1)` as "the newest", but `-` (0x2D) sorts before `.` (0x2E), so `raw-2026-07-31.jsonl` (4-case smoke) beat `raw-2026-07-31-full.jsonl` (26-case sweep) sitting beside it. Both `judge.ts` and `report.ts` defaulted to it, so the documented no-argument workflow judged the smoke run — and `report.ts` would have regenerated the committed `RESULTS.md` from 4 cases, which reads exactly like a full sweep at a sixth of the denominator. That is the path the "re-run against production" item below walks.

Fixed by refusing to guess rather than guessing better (mtime is no improvement — a fresh clone stamps every file with checkout order). With more than one sweep in `results/`, both scripts exit listing the candidates and their case counts.

Same commit: the judge's header claimed snippets "are labelled A/B by a per-item coin flip" and the loop called `void coin(...)`, computing a flip and discarding it. Snippet utility is scored one excerpt at a time on an absolute rubric — no second text in the prompt, so nothing to blind. The scoring was never wrong and the arm was never in the prompt, but the comment described a mechanism that wasn't there. Also: both judge calls swallowed every error with a bare `catch`, so a bad key or a 429 became an `unscored` row indistinguishable from a real coverage gap — the exact conflation the harness's own design notes exist to prevent.

**`RESULTS.md` reproduces byte-identically from the committed JSONL** — verified by regenerating it. The reported findings stand; it was the tooling around them that was wrong.

## Ordering — cleared

Both blockers are resolved:

- **Backend #28462 is on production.** `ce1754de` is an ancestor of `origin/production`, and `Terraform Production CI/CD` for the containing sha (`3afee899`) concluded **success** at 2026-08-01T00:56Z. This mattered because `sources.web` is `extra="forbid"` server-side, so the key 400s every web-inclusive search and answer on a backend that predates it.
- **#197 (the OpenAPI sync) merged**, so `highlights` is in the generated Zod on `main`. The two sync commits dropped out of this branch during the rebase as already-upstream. The `satisfies z.input<typeof SearchRequest>` guard in both builders means this fails to compile without it, which is the intended signal rather than a silent 400.

## Scope notes

- `snippet_max_chars` stays at **2000** on both tools — deliberately one variable at a time. The cap-binding asymmetry above (64% of text results at cap vs 15% of highlight results) is an argument for revisiting it as a *separate* change, not a companion to this one: raising it now would mostly buy more text-path padding.
- Adds **`@anthropic-ai/sdk` as a devDependency** for the blind judge. Not bundled into the Worker, but it is a new dep — flagging it since nobody asked for one.
- Deliberately unchanged, each re-checked against post-rebase `main`: `tako_search`'s "a web result is only a snippet, not a value" (it routes "what is X" to `tako_answer`; better snippets are a reason to keep it, not soften it); `llms-full.txt` (mirrors a DESCRIPTION line that is still accurate); `renderWebResult` still omits the snippet from markdown (it rides in `structuredContent`; both channels would bill the prose twice); `slimWebResult` slims only `content`, so the snippet survives to `structuredContent` on the answer path; `registry/server.json` carries no output schemas, verified clean.
- Nothing in `src/` reads `.snippet` programmatically, so a `null` snippet cannot break a code path — it rides through into `structuredContent`.
- Measurement only — `scripts/golden.ts` remains the pass/fail gate. Nothing in the harness is wired into CI.

## Test plan

- [x] Rebased onto `main` (0.16.1); `MERGEABLE`, no conflicts
- [x] 764 unit + 16 widget + 30 script tests pass
- [x] Typecheck passes on all 4 tsconfig projects (`tsc --noEmit` alone misses 3 of them)
- [x] `registry:check` and `schemas:check` clean
- [x] New guards verified RED before GREEN
- [x] Live staging A/B, 104 calls, 0 failures — `RESULTS.md` committed
- [x] `RESULTS.md` verified reproducible byte-for-byte from the committed JSONL
- [x] Backend #28462 confirmed deployed to production (ancestry + successful deploy run)
- [ ] Re-run the sweep against production — needs a **prod** API key; only a staging key was available locally (a staging key 401s on `tako.com`)
- [ ] Watch answer p95 after deploy, given the 8.6s outlier


workers/scripts/evals/web-snippet/FINDINGS.md
```diff
@@ -0,0 +1,103 @@
...
+## Coverage: the null-snippet risk did not materialise
+
+**0 null snippets in 260 highlight results.** Upstream measured 0 across ~40
+and declined to add a text fallback on that basis; this holds at 6.5× the
+sample, including two deliberately near-contentless queries (`market outlook`,
+`AI spending`). The `webResultSchema.snippet` description still documents
+`null`, because absence is in the contract even if it is rare — but no fallback
+is warranted on this evidence.
+
+Only 1 of 260 highlight snippets contained the `" … "` multi-passage join,
```
robertabbott: `report.ts:213` counts multi-passage snippets by testing for `" … "`, and the **text** arm — which requests no highlights and therefore cannot contain a backend join — returns 2 of 260. The backend emits that separator in exactly one place, `exa_provider.py:418` (`highlights = " … ".join(passages)`), so those 2 are the source pages' own ellipses surviving `_clean_excerpt`'s whitespace collapse.

The single highlights hit is the same `rigzone.com` URL at the same textual position as one of them: the text arm reads `…adjustments in April and November 2023 … met virtually on 5 July 2026`, the highlights arm reads `…April and November 2 ... 23 … met virtually on 5 July 2026`. The ` ... ` there is the intra-passage newline collapse that `exa_provider.py:408-417` documents, not a join.

Does that leave any confirmed Exa join in the run? Should this read as 0 observed joins against a 2/260 false-positive floor on the detector, rather than 1 — and does the advertised wording ("It may join passages from different parts of the page with ' … '") need to account for a model that will read a source's own ellipsis as a passage boundary?

workers/scripts/evals/web-snippet/judge.ts
```diff
@@ -0,0 +1,378 @@
...
+}
+
+async function main(): Promise<void> {
+  const rawPath = process.argv[2] ?? resolveRaw();
+  const rows = readFileSync(rawPath, "utf8")
+    .split("\n")
+    .filter((l) => l.trim() !== "")
+    .map((l) => JSON.parse(l) as CaseRow);
+
+  const out = rawPath.replace(/\/raw-/, "/judged-");
```
robertabbott: `rawPath.replace(/\/raw-/, "/judged-")` needs a slash before `raw-` to match. Invoked as `cd results && npx tsx ../judge.ts raw-2026-07-31-full.jsonl`, `out` is the input path unchanged, and line 284's `writeFileSync(out, "")` then truncates the sweep before appending judged rows over it.

That is the one file `RESULTS.md` is reproducible from, and the loss is silent. Should this anchor on the basename (or assert the derived path differs from the input) the same way `assertRawShape` guards `report.ts`'s boundary?

workers/scripts/evals/web-snippet/FINDINGS.md
```diff
@@ -0,0 +1,103 @@
...
+Numbers in `RESULTS.md`; this file is the interpretation, which the generated
+sheet deliberately does not contain.
+
+## Verdict
+
+**Flipping `highlights` on for `tako_search` is well supported.** Flipping it
+for `tako_answer` is supported but on a thin denominator, and one latency
+outlier is unexplained — details and the reason it still ships below.
+
+## `tako_search`: better excerpts, and 18% fewer bytes
```
robertabbott, line 14: `RESULTS.md:53` reports the paired search delta as median +49ms, max +522ms, highlights slower on 18 of 26 cases. That is directional in a way the answer row (+1ms, 13 of 26) explicitly is not, and `report.ts:139` titles the whole latency block "the gate on `tako_answer`", so nothing in this file or the PR body reads the search row.

`tako_search` defaults to `count: 10` and is the higher-volume tool. Is +49ms median on 69% of cases the intended read — a cost accepted for 18% fewer bytes and +0.14 utility — or does the search row want its own sentence here, so the next reader does not have to derive it from the generated sheet?

workers/scripts/evals/web-snippet/run.ts
```diff
@@ -0,0 +1,223 @@
...
+async function call(path: string, query: string, arm: Arm, withCount: boolean): Promise<CallRow> {
+  const body = { query, sources: { web: webSource(arm, withCount) } };
+  const t0 = Date.now();
+  try {
+    const res = await fetch(`${apiBase}${path}`, {
+      method: "POST",
+      headers: { "X-API-Key": apiKey as string, "Content-Type": "application/json" },
+      body: JSON.stringify(body),
+    });
+    const ms = Date.now() - t0;
```
robertabbott: `ms` is captured before `await res.text()` on line 128, so it measures time-to-headers, not time to a complete response. The two arms differ by ~18% in payload (`RESULTS.md:65-66`: 1896 vs 1556 mean chars, ×10 results), so the excluded component is systematically larger on the text arm — which inflates the `highlights − text` delta the report calls the metric that gates the change.

Should `t0` be closed after the body is read, or is header latency the intended measurement given the arbiter reads the body server-side anyway?

workers/scripts/evals/web-snippet/judge.ts
```diff
@@ -0,0 +1,378 @@
...
+  for (const f of files) {
+    const cases = readFileSync(join(RESULTS, f), "utf8").split("\n").filter((l) => l.trim() !== "").length;
+    console.error(`    ${f}  (${cases} case${cases === 1 ? "" : "s"})`);
+  }
+  console.error(`\n  npx tsx scripts/evals/web-snippet/judge.ts <path-to-raw.jsonl>`);
+  process.exit(1);
+}
+
+async function main(): Promise<void> {
+  const rawPath = process.argv[2] ?? resolveRaw();
```
robertabbott: `process.argv[2] ?? resolveRaw()` takes the first argument positionally, but lines 70-78 also read `--top` and `--conc` from `process.argv`. `npm run eval:web-snippet:judge -- --top 10` sets `rawPath` to `"--top"` and dies in `readFileSync` with an ENOENT that names a flag. The README documents the path and both flags on the same step, so flag-first is a shape a reader will produce. Should the positional pick the first argument that does not start with `--`?

workers/scripts/evals/web-snippet/run.ts
```diff
@@ -0,0 +1,223 @@
...
+  console.log(`target: ${apiBase}  cases: ${cases.length}  arms: text | highlights`);
+  console.log(
+    `  ${cases.length * perCase} priced calls, ~$${(cases.length * (searchOnly ? 0.014 : 0.032)).toFixed(2)}`,
+  );
+  console.log(`  snippet_max_chars=${SNIPPET_MAX_CHARS} (mirrors the MCP tools), throttle=${THROTTLE_MS}ms\n`);
+
+  const outDir = join(dirname(new URL(import.meta.url).pathname), "results");
+  mkdirSync(outDir, { recursive: true });
+  const out = join(outDir, `raw-${process.env.EVAL_STAMP ?? "latest"}.jsonl`);
+  writeFileSync(out, "");
```
robertabbott, line 188: `writeFileSync(out, "")` truncates unconditionally, and `EVAL_STAMP` defaults to `latest` (line 171). The remaining test-plan item is a re-run against production; a stamp collision with `2026-07-31-full` destroys the committed JSONL that `RESULTS.md`'s byte-identical reproduction claim rests on. Given `judge.ts` and `report.ts` now refuse to guess *which* sweep to read, should the write side refuse to clobber one that already exists?

workers/scripts/evals/web-snippet/judge.ts
25eliu: Confirmed and fixed in d63a5dc — this was the worst of the six.

Reproduced exactly as you describe. `cd results && npx tsx ../judge.ts raw-2026-07-31-full.jsonl`: **56,188 bytes → 1,870**, and it printed `wrote raw-2026-07-31-full.jsonl` on the way out.

You also found it in a second place. `report.ts:90` had the identical derivation — not destructive there, since that path is only ever read, but a bare basename made it parse the raw rows as `JudgedCase` and die in `.flatMap` on an undefined `snippets`. Both were the same bug duplicated, so the derivation now lives once in `paths.ts`, anchored on the basename via `dirname`/`basename`, and asserts the result differs from its input before anything writes — the `assertRawShape` posture you pointed at, applied at the other boundary. That file also absorbs the duplicated `resolveRaw`, which is how the two copies drifted in the first place.

Verified: input sha256 unchanged, `judged-fixture.jsonl` written alongside, and `RESULTS.md` still regenerates byte-identically from the committed JSONL — now also when invoked with a bare basename.

workers/scripts/evals/web-snippet/judge.ts
25eliu: Fixed in d63a5dc. Reproduced: `npm run eval:web-snippet:judge -- --top 10` → `Error: ENOENT: no such file or directory, open '--top'`.

One correction to the suggested shape, which I hit while implementing it: picking the first argument that does not start with `--` selects the **`10`** out of `--top 10`, which then fails as a path — same class of bug, one step further along. The value-consuming flags have to be named, so `paths.ts` carries `VALUE_FLAGS = ["--top", "--conc"]` and the positional scan skips a flag plus its value.

Now: `-- --top 10` falls through to the refuse-to-guess listing (correct — no path given), and `--top 10 <path>` resolves the sweep with `top=10` honoured.

workers/scripts/evals/web-snippet/run.ts
25eliu: Yes — fixed in 2f76a7a, and extended one place further than you asked.

`refuseToClobber` now guards the write side: if the target exists it exits 1 naming the file and the three ways out (different `EVAL_STAMP`, move the file, `--force`). Verified it fires **before the first priced call** — `EVAL_STAMP=2026-07-31-full` exits 1 with the committed sweep's sha256 unchanged and zero fetches issued.

The extension: `judge.ts`'s output write has the identical exposure, and I have direct evidence it bites. While testing the previous commit I ran a stray re-judge against the full sweep, and it silently wiped the committed `judged-2026-07-31-full.jsonl` — recovered only because git had it. Same hazard, same fix, so the guard covers both write sides.

Your framing is the one I used in the commit message: given `judge.ts` and `report.ts` now refuse to guess which sweep to *read*, refusing to clobber one is that same rule at the other end.

workers/scripts/evals/web-snippet/run.ts
25eliu: Confirmed — `t0` now closes after the body is read (2f76a7a). Header latency was not the intended measurement; it was an oversight, and your read of the direction is right.

Quantified it with a local probe that flushes headers then stalls 300ms before the body: **12ms measured the old way, 301ms the new**. So in the worst case the old clock was excluding nearly the entire exchange.

On the bias direction, to state it explicitly since it matters for the ship decision: the excluded component is larger on the **text** arm (~18% more payload), so the old numbers *inflated* `highlights − text`. The committed +49ms search and +1ms answer deltas are therefore **upper bounds** — the true gaps are smaller. The bias ran in the conservative direction, but the sheet did not say so.

Rather than hand-annotate one run, `CallRow` now carries a `timing` marker and `report.ts` states what the clock covered, defaulting to `headers` when the field is absent. The 2026-07-31 sweep is pre-fix and now declares that in its own generated header, so a pre-fix and a post-fix sweep can never be read as the same measurement. No number in `RESULTS.md` moved — regenerating from the committed JSONL adds only the caveat.

workers/scripts/evals/web-snippet/FINDINGS.md
25eliu: Your forensics are exactly right, and the answer to "does that leave any confirmed Exa join in the run" is **no — zero**. Corrected in c79b09f.

I pulled all three hits out of the raw JSONL to check independently:

| arm | case | text |
|---|---|---|
| text | `opec-decision` (rigzone) | `…April and November 2023 … met virtually on 5 July 2026` |
| **highlights** | `opec-decision` (**same URL, same position**) | `…April and November 2 ... 23 … met virtually on 5 July 2026` |
| text | `fed-meeting-outcome` (mufgresearch) | `…(without a Fed compass) … - By George Goncalves` |

So it reads as **0 observed joins against a 2/260 false-positive floor**, as you suggested. The framing I took from your comment and put in both the report and FINDINGS: *the text arm IS the floor* — it requests no highlights and therefore structurally cannot carry a backend join, so whatever it reports is the detector reading source punctuation. `report.ts` now names the column `contains " … "` rather than `multi-passage`, and says to subtract the text row before reading the highlights row as evidence of joining.

**On the advertised wording — yes, it needed to account for that, and it now does.** A consumer cannot distinguish the two causes any better than the detector can, so asserting the cause was the wrong shape. Both tool surfaces now read: a `' … '` marks a discontinuity — *joined passages or the page's own ellipsis* — so never quote across it. The actionable half is unchanged, the forward guard against a vendor-default change survives, and it now also covers the case that actually occurs in the data.

One thing I deliberately did **not** touch: the backend's own `highlights` description in `src/generated/schemas.ts` still says "joined by ' … '". That is synced from the OpenAPI spec, so it is a backend edit. Happy to open something against `tako` for it if you think it is worth aligning.

workers/scripts/evals/web-snippet/FINDINGS.md
25eliu: You are right that nothing read it, and it should have. Addressed in c79b09f — the search row now gets its own sentence in FINDINGS.md rather than being left for the next reader to derive.

To answer directly: **the intended read is "a cost accepted", and it is now written down as that** rather than being implied by silence. ~6.5% of a 750ms median, bought for 18% fewer snippet bytes and +0.14 blind utility. The distinction I made explicit alongside it is the one your comment implies: the answer path has a cliff (the hard 1.5s grounding budget, where a breach drops grounding entirely), and the search path has none — so +49ms there is a straight price, not a failure mode. `report.ts`'s latency section is retitled and now frames both endpoints instead of just the gate.

Two things make it an **upper bound** rather than a point estimate, both now stated:
- The sweep timed to response headers (your other comment) — the excluded body-download component is systematically larger on the *text* arm, so the true delta is smaller than +49ms.
- n=1 per case from a laptop is not a p95.

So if search latency ever becomes the binding constraint, that is a number to re-measure properly, not to re-derive from this run.

robertabbott (approved):

PR #203, opened 2026-08-01
feat(widget): interactive Tako cards on Claude, plus a dead skill and three widget bugs

25eliu:
Started as a review of the skills / platform surfaces and turned up a dead skill, three user-visible widget bugs, and a route to rendering Tako's **real** interactive card on Claude rather than a screenshot of it.

Everything new is behind `PUBLIC_CDN_URL`, which is set on staging and **absent from production**. With it unset both new routes decline, the CSP is unchanged, `_meta` carries nothing new, and the widget never fetches. `wrangler deploy --dry-run --env production` reports zero experiment surface.

## The dead skill

`skills/tako-web-traffic/SKILL.md` described itself as `(source: SimilarWeb)`. A bare `: ` in a YAML **plain** scalar is a mapping separator, so the frontmatter did not parse — and Claude Code answers a parse failure with **empty metadata** rather than an error, dropping `name` and `description`. A skill with no description is one the model never learns exists, so it has been inert on every plugin install since that wording landed.

Diagnosed and fixed once before (b12e4c6) — but only in the README's copy-paste blocks, never in the files the plugin ships. Two reasons it survived, both closed here: nothing validated skill frontmatter (`scripts/skills.test.ts` now does, and was verified to fail on the pre-fix file), and `workers-ci` did not trigger on `skills/**` or `.claude-plugin/**`, so the PR that broke it ran no CI at all.

## Three widget bugs, each user-visible

**We fired a CSP violation on every chart.** After painting the PNG the bundle probed the interactive `embed_url` in a hidden iframe. claude.ai hardcodes `frame-src 'self' blob: data:` and ignores declared `frameDomains`, so that raised `securitypolicyviolation` every single render — and the host reports a blocked subresource to the user as *"There was a problem displaying content from tako."*, next to a chart that rendered perfectly. The probe's only audience was the one host where it always failed.

**A failed tool call left a mounted widget that never painted.** An `isError` result carries no `structuredContent`, but the host has already mounted the widget, so nothing drew and the frame sat at a 1px sliver indefinitely. A successful zero-card result already collapsed; a failed call now does too.

**The iframe was a fixed 720px, so cards sat in a near-square box with a band of empty space.** Measured cause: the PNG endpoint renders at a fixed 2400px width with a per-card height — 2.18:1 for a plain chart, 1.91:1 for a Stock Overview, 1.30:1 for a ranked list. No constant fits three aspects. The widget now reads the card's true aspect from the PNG's IHDR chunk via a `Range: bytes=0-63` request (two integers for 64 bytes, not a 170KB render) and re-fits on resize.

## Interactive cards on Claude

Claude now runs **Tako's own** `Card.js` — its view chunks, its Geist fonts, its declarative theme. Not an approximation, so no divergence from ChatGPT, and it covers license-gated cards that a from-scratch renderer never could.

Two proxy routes make it work, both gated:

- `/embed-html/{pub_id}` serves Tako's embed page with CORS (that route sends none of its own). The widget writes the markup into its own document so Tako's config/theme islands and its `Card.js` tag execute natively — no mount contract to reverse-engineer.
- `/cdn-asset/{path}` serves the assets. The CDN reflects `access-control-allow-origin` for `tako.com` alone, and a `type="module"` script is always a CORS fetch, so assets loaded from the CDN inside a sandbox all failed. Serving them from our origin removes the upstream dependency entirely.

Both fetch from a **config-fixed** origin, never one from the request: a caller controls the path and can never redirect the host. `pub_id` is validated `[A-Za-z0-9_-]{1,64}`; asset paths reject traversal; `dark_mode` is allow-listed to two literals.

**It is an upgrade, never a first paint.** The PNG commits first and `document.open()` only fires once the markup is in hand. 502, timeout, non-HTML, network error — each costs an upgrade and leaves today's chart untouched, asserted one test per failure mode. That held live twice under real failures.

Verified end to end in a real browser: `canvas=1, svg=5`, title, 1Y/2Y/Max range selectors, chart/table toggle, Geist, and a live source line.

## Also

- **`tako_visualize` is default-on for ChatGPT and Claude** — the hosts that render its chart inline. Gated by one exported `isWidgetClient` used by both the default-on set and `widgetSuppressed`, replacing two inlined comparisons in separate files. Costs ~3.5k tokens of standing schema on those clients. The free tier is untouched: the free-tier gate runs first, so anonymous connections still cannot reach it, spoofed UA included.
- **Transparent widget surface and host-matched theme.** Two variants painted an opaque `#0f1115`, which on a light host is a dark box with square corners inside the host's rounded container. And `dark_mode=auto` resolves from the OS, so a dark ChatGPT on a light machine got a light card; `withHostTheme` now believes the host when it declares one.
- **The raw `User-Agent` is logged** alongside the `McpClientKind` it resolved to. That kind selects the tool set, widget registration, image blocks, annotation labels and the ChatGPT `securitySchemes` injection — an unrecognized UA degrades all five silently. `detectMcpClient`'s docblock asked for this and nothing logged it.

## Corrections recorded in the commits

Two conclusions this branch was built on turned out wrong, and the commits say so rather than quietly moving on. The `script-src` probe returned ALLOWED, but it loaded a **classic** script and classic scripts need no CORS — necessary, not sufficient. And an early "CORS works" check passed only because it ran from a `tako.com` page, the one origin on the CDN's allow-list.

## Test plan

- [x] Typecheck clean on all four projects
- [x] **925 tests pass** (840 + 42 + 43)
- [x] `registry:check` and `schemas:check` clean
- [x] `wrangler deploy --dry-run` clean for staging and production
- [x] Production dry-run shows zero experiment vars
- [x] Native card verified rendering in a real browser through a real tunnel
- [x] Fallback verified live: widget kept the chart through two real fetch failures
- [ ] Deploy staging and confirm the native card in claude.ai's own sandbox
- [ ] Confirm the *"problem displaying content"* toast is gone with the probe retired
- [ ] Re-run `claude plugin validate` after merge to confirm the skill loads

## Not fixed here

- `tako_visualize` 402s on staging (that account is out of credits), so the native path was verified via `tako_search`.
- Both chart tools use a 130s upstream timeout while claude.ai's is shorter, so a slow call surfaces as *"Unable to reach tako"* rather than an honest error.
- ~~`tako_search` still sends `snippet_max_chars: 2000` while the synced spec now defaults search to 4000.~~ **Resolved on main** — `64c0a4b` / `95b8bb0` kept 2000 as a deliberate per-call token budget and rewrote the comment to say so. This branch's flag-only comment was superseded wholesale in the rebase.
- The widget's own click-through is dead on claude.ai (`allow-popups` is not set on the sandbox). The model's `[Open in Tako]` markdown link still works.


robertabbott (commented):
Reviewed with a multi-pass sweep (claim verification, blast radius, security, cross-layer contracts, failure modes, test adequacy, git archaeology). Findings inline, ordered roughly by weight. The engineering here is strong and the commit messages are unusually honest about what got disproven along the way — most of what follows is non-blocking.

Three things I would want resolved before this ships, all on the new proxy surface rather than the widget fixes:

- `index.ts:54` — `resolvePublicCdnBase` throws rather than declining, and it now runs ahead of the entire OAuth surface. A trailing slash in `PUBLIC_CDN_URL` 500s `/authorize`, `/token` and `/login` while `POST /mcp` keeps working.
- `embed_proxy.ts:315` / `:419` — `mcp.tako.com` is the OAuth origin, and both routes serve documents there under a content type they choose from upstream. The `/embed-html/` consumer reads the body as text and never needs `text/html`.
- `tako_search.ts:301` — the deleted `if (ctx.client === "chatgpt") return undefined;` puts a new 3 s-bounded fetch on the production ChatGPT critical path. That is real and fine, but it is not behind `PUBLIC_CDN_URL`, so "with it unset ... `_meta` carries nothing new" does not hold as written.

Two notes on what I could and could not check. I verified the dead-skill diagnosis independently rather than trusting it — running the repo's own js-yaml 4.1.0 against the base-commit files, `tako-web-traffic` throws `bad indentation of a mapping entry (2:54)` while the other two parse cleanly, so the root cause is exactly as described. But I had no Node toolchain, so `925 tests pass`, `registry:check`, `schemas:check` and `typecheck clean` are **unverified by me** — CI is green and two of the three counts (42, 43) reconstruct exactly from static analysis, but I did not run the suite and am not reporting numbers I did not see.

Happy to re-review, and glad to be wrong on the CloudFront bucket question in particular — I could not see the distribution's cache behaviors from here.

workers/src/index.ts
```diff
@@ -45,6 +46,22 @@ export default {
       return handleIconRequest(url.pathname);
     }
 
+    // CORS-readable proxy for Tako's chart embed page, so the MCP Apps widget
+    // can render the REAL interactive card rather than a PNG of it. Returns
+    // `undefined` — falling through to the 404 below — when the
+    // `PUBLIC_CDN_URL` experiment is off or the pub_id is malformed, so this
+    // route does not exist at all in production. See `embed_proxy.ts`.
+    const embedProxied = await handleEmbedProxy(request, env);
```
robertabbott, line 54: `handleEmbedProxy` opens with `resolvePublicCdnBase(env)`, and that delegates to `validatePublicOrigin` (`env.ts:198-224`), which **throws** on a trailing slash, an unparseable URL, or a non-http(s) scheme rather than returning `undefined`. There is no try/catch here, and this line sits above the whole OAuth subsystem: `/authorize`, `/token`, `/register`, `/login`, `/oauth/stytch_callback`, and both `.well-known` discovery docs are all below it.

So `PUBLIC_CDN_URL="https://d1iyjvzoctsna.cloudfront.net/"` — one trailing character — turns every one of those into a 500. `POST /mcp` returns at line 28 and survives, which means tool traffic would look healthy while the connector's auth flow is completely down. Before this PR that binding was only read while building a widget payload, so a bad value degraded to "no native card".

Would catching-and-declining inside `handleEmbedProxy` / `handleCdnAssetProxy` — a misconfigured experiment binding disables the experiment, exactly as an unset one does — keep the blast radius where it was?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+  // reaching a third-party sandbox unnoticed after an upstream redesign.
+  if (!sanitized.removedCsrf) {
+    console.warn(
+      `[mcp] embed proxy found no csrfToken island to strip for ${pubId} — upstream shape may have changed`,
+    );
+  }
+  return new Response(html, {
+    status: 200,
+    headers: {
+      "content-type": "text/html; charset=utf-8",
```
robertabbott: This makes `mcp.tako.com` serve an executable `text/html` document, and that origin is the OAuth origin. `oauth/cookies.ts:43-61` sets `tako_oauth_state` and `tako_oauth_session` with `Path=/` and no `Domain`, so both ride every same-origin request here, and `oauth/handlers.ts:587-593` states the design's assumption verbatim: "the cookie alone is adequate for the threat model where an attacker has neither XSS on mcp.tako.com nor an active MITM."

Script execution on this origin would not need to read the HttpOnly cookie — it could `fetch('/register')` (open DCR, no auth at `index.ts:146`), then drive `/authorize` + `/token` same-origin with the victim's session cookie attached, and read the auth code out of the same-origin response.

What I could not find is a reason the response has to be `text/html`. The only consumer is `_chart_widget.ts:921-947`, which does `fetch(nativeUrl, {credentials:'omit'}).then(r => r.text())` and then `document.write(patched)` into its own sandboxed document — it reads the body as text and never dispatches on the content type (the sole shape check is `html.indexOf('<html') === -1`). Would `text/plain; charset=utf-8` work identically for the widget? If a browsable type is wanted anyway, would `Content-Security-Policy: sandbox` plus `X-Frame-Options: DENY` close the same-origin execution path without touching the widget?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+  if (!response.ok) {
+    return textResponse(
+      "asset not available",
+      response.status === 404 ? 404 : 502,
+    );
+  }
+
+  const headers = new Headers();
+  const contentType = response.headers.get("content-type");
+  if (contentType !== null) headers.set("content-type", contentType);
```
robertabbott: Reflecting the upstream content-type verbatim means whatever MIME the CDN declares becomes the MIME we serve on `mcp.tako.com`, and I could not convince myself every object on that distribution is inert.

In the Tako repo, `app/config/settings/production.py:135` sets `AWS_S3_CUSTOM_DOMAIN = CLOUDFRONT_URL` over bucket `tako-static-assets-<branch>`, with static files under the `archive/<sha>/` prefix (`AWS_LOCATION`) — i.e. the distribution fronts the bucket root, not an `archive/`-only origin path. Two authenticated endpoints hand out presigned POSTs into that same bucket with a caller-supplied filename: `MyPreSignedUrlView` (`app/backend/api/views.py:91-103`, key `user-images/{user_id}/{uuid}/{file_name}`) and `ContentStyleFileView` (`app/backend/users/api/views.py:747-754`, key `content-style/{user_id}/{uuid}/{file_name}`). Both call `generate_presigned_post` with no `Conditions`, so the stored content-type is whatever the upload negotiates.

I cannot see the CloudFront cache-behavior path patterns from here, so this is a question rather than a claim: is `/cdn-asset/user-images/<uid>/<uuid>/x.html` reachable through this route, and if the stored object carries `text/html`, is anything stopping it from being served as an executable same-origin document? If that is not a firm no, would allow-listing the reflected type (javascript/css/font/image/json, everything else forced to `application/octet-stream`), or confining `assetPath` to the `archive/` prefix the rewrite actually targets, be cheaper than depending on the distribution's routing staying as it is?

workers/src/tools/tako_search.ts
```diff
@@ -281,23 +291,26 @@ const tako_search = {
...
-      image_natural_height: fetched.naturalHeight,
-    };
+    // ChatGPT gets DIMENSIONS ONLY (a 64-byte ranged read): its widget renders
+    // `embed_url` in an iframe and never reads the baked PNG, but it cannot
+    // measure that cross-origin iframe's content, so without the card's real
+    // aspect ratio the iframe falls back to a fixed height and leaves empty
+    // bands under a wide chart. Claude gets the full baked image — there the
+    // PNG *is* the chart.
+    return buildChartExtraMeta(output.image_url, {
+      bakeImage: ctx.client !== "chatgpt",
```
robertabbott, line 310: The base commit opened this hook with `if (ctx.client === "chatgpt") return undefined;` (`git show 50c1105:workers/src/tools/tako_search.ts`). That line is gone, so with `PUBLIC_CDN_URL` unset — i.e. production — a ChatGPT chart call now takes a path it never took before: `buildChartExtraMeta` falls into the `!bakeImage` branch and awaits `fetchPngDimensions(image_url)`, a new outbound request on the critical path of every call, bounded by `PNG_HEAD_FETCH_TIMEOUT_MS = 3_000`.

That makes the PR body's "with it unset ... `_meta` carries nothing new" not quite hold for this client — it returns `{image_natural_width, image_natural_height}` where it previously returned `undefined`, and it is not behind the flag. (`interactive_probe` at `_chart_widget.ts:260` is likewise a new `_meta` key on the Claude side; `git show 50c1105:... | grep interactive_probe` finds nothing on the base.)

Could the ChatGPT p50/p99 delta go in the PR body? It is the one production behavior change the flag does not gate, and `image_url` points at `/api/v1/image/{pub}/`, which renders on demand — so a `Range` header saves the ~170 KB transfer but not the render, which makes that 3 s bound the number that matters. Relatedly, the docblock at `_chart_widget.ts:1630` says the ranged read avoids "the full chart-render latency"; if the render still has to happen before any byte range can be served, would that wording mislead the next reader sizing this timeout?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+        return whole;
+      }
+      if (typeof parsed !== "object" || parsed === null || Array.isArray(parsed)) {
+        return whole;
+      }
+      const rest = { ...(parsed as Record<string, unknown>) };
+      if (!("csrfToken" in rest)) return whole;
+      delete rest.csrfToken;
+      removedCsrf = true;
+      return `<script${attrs} data-tako-stripped="csrf">${JSON.stringify(rest)}</script>`;
```
robertabbott: This round-trip drops upstream's script-breakout escaping. Django's `json_script` escapes `<`, `>`, `&` as `\u003C` / `\u003E` / `\u0026` inside the JSON string specifically so a value can never close the `<script>` element; `JSON.parse` decodes those back to literals and `JSON.stringify` re-emits them raw. Reproduced mechanically — an island body of `{"csrfToken":"T","title":"a \u003c/script\u003e\u003cimg src=x onerror=alert(1)\u003e b"}` comes out of this replace with a literal `</script>`, terminating the island early and turning the remainder into live markup.

I want to be precise about reachability, because two passes disagreed and the narrower reading looks right: the island this regex matches is `server_config` from `make_server_config` (`app/backend/monolith/views.py:1813-1822`), and its four surviving keys are all server-derived — `staticPrefix`, `userLoggedIn`, `hostContext` (only `"consumer"`/`"developer"`), `developerConsoleEnabled`. Caller-authored card content lives in the separate `config-json` island, whose top-level keys are fixed, so `'csrfToken' in rest` is false there and this replace correctly returns `whole` with escaping intact. So this is **latent, not live**.

It is still a silently-removed control that nothing downstream replaces, and it re-arms the moment one user-influenced value joins that dict. Rather than a `.replace(/</g, "\\u003c")` pass on the stringified output, would deleting the key textually — a targeted replace of the `"csrfToken":"..."` member inside the raw body, leaving every other byte exactly as Django emitted it — be the version that cannot regress?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+  const upstream = `${cdnBase}/${assetPath}${url.search}`;
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), UPSTREAM_TIMEOUT_MS);
+  let response: Response;
+  try {
+    response = await fetch(upstream, {
+      signal: controller.signal,
+      redirect: "follow",
```
robertabbott: `redirect: "follow"` is in tension with the guarantee the docstring above states — "The upstream ORIGIN is fixed by configuration (`PUBLIC_CDN_URL`), never taken from the request. That is the whole reason this is not an open proxy."

That holds for the request we *issue*; it does not hold for the response we *serve*. A 3xx anywhere upstream is followed transparently, and every downstream decision is then made against the final hop: `response.ok`, the `content-type` reflected verbatim at line 419, and the body — all re-served under our origin with `access-control-allow-origin: *`. The same applies to `handleEmbedProxy` at line 241, where the `text/html` gate at line 262 would be judging a document from whatever host the chain ended at. The Tako origin does redirect on this class of path (`app/backend/monolith/middleware.py:146` canonical-host 301/308, `:250` `redirect("/login")`), so following is not hypothetical, only currently same-host.

Would `redirect: "manual"` on both fetches — treating a 3xx as an upstream error rather than an instruction — make the docstring's claim actually true, given neither route has any reason to leave its configured origin?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+    contentType !== null && /javascript|ecmascript/i.test(contentType);
+  if (isJavaScript) {
+    // Only JavaScript is buffered, and only up to a bound. Everything else
+    // streams. The upstream origin is fixed by config so this is not an
+    // arbitrary-URL risk, but `Card.js` is already ~1.5 MB and a rewrite means
+    // holding the whole body plus its copy — worth a ceiling rather than
+    // trusting the CDN to stay reasonable forever. Over the cap, stream it
+    // through unrewritten: the assets that actually need rewriting are the
+    // entry bundle and its chunks, all far below this.
+    const declaredLength = Number(response.headers.get("content-length") ?? "0");
```
robertabbott: Does this cap engage on the asset it was written for? It reads `Number(response.headers.get("content-length") ?? "0")`, and Workers strips `Content-Length` (along with `Content-Encoding`) when it auto-decompresses a gzipped response — which is how CloudFront serves JavaScript. If that is right, then for `Card.js` — the ~1.5 MB asset the comment names — `declaredLength` is `0`, the branch never trips, and `await response.text()` on line 449 buffers the whole body regardless of size, which is the exact unbounded case `e3b8366` added this to close.

Two smaller versions of the same gap: a non-numeric header yields `NaN`, and `NaN > MAX_JS_REWRITE_BYTES` is `false`, so that also falls through; and where the header *is* present on a compressed response it describes compressed bytes while the cap governs a decompressed buffer.

The test at `embed_proxy.test.ts:446` supplies `content-length` explicitly, so it passes while the real path bypasses the cap — and the neighbouring `rewrites CDN urls baked inside a JS bundle` test is already silently on the header-absent path, since `new Response(string)` sets no length in workerd. Would reading the bound off the accumulated body instead of a declared header, plus a test for a JS response with no `content-length`, pin this?

workers/src/tools/_chart_widget.ts
```diff
@@ -710,7 +1104,8 @@ const WIDGET_HTML = `<!doctype html>
         // embed iframe load (it doesn't on claude.ai until
         // anthropics/claude-ai-mcp#40 is fixed), the probe upgrades to the
         // interactive chart once it does.
-        if (validEmbed) probeInteractiveIframe(url, h);
+        if (validEmbed && allowProbe) probeInteractiveIframe(url, h);
+        upgradeToNativeCard(withHostTheme(opts.nativeCardUrl), h);
```
robertabbott, line 1143: Two things meet at this call site, and I think the second is the more consequential.

**The empty-string guard does not survive the transform.** `upgradeToNativeCard` guards on `nativeUrl === ""` (line 913), but the value reaching it has been through `withHostTheme` first. When `_meta.native_card_url` is absent, `nativeCardUrl` is `""` (lines 1209-1212) — and `""` *is* a string, so `withHostTheme` skips its `typeof` bail, takes the append branch at line 656, and returns `"?dark_mode=true"`. That clears the guard and reaches `fetch("?dark_mode=true")`, resolved relative to the widget document — a request in nobody's `connect-src`, which is the same blocked-subresource class this PR retires the iframe probe to avoid. Would guarding before the transform, or having `withHostTheme` return `url` unchanged when falsy, close it? Same shape at line 1142.

**The host-theme fix may not reach the host it was written for.** `hostTheme()` reads only `window.openai.theme`. The native path is armed only from `paintImage`, which runs only when `hasOpenAiRuntime()` is false (line 1241 sends every `window.openai` host to the iframe branch). So on claude.ai `hostTheme()` returns `null`, `withHostTheme` leaves the URL alone, and the proxy always sees `dark_mode=auto` — the OS-derived value, i.e. exactly the mismatch `8248f91` set out to fix. ChatGPT has the theme but never takes this path. That would also make the proxy's `dark_mode` allow-list (three test blocks at `embed_proxy.test.ts:680-723`) a guard on a parameter nothing currently sends. Is there a host-theme source for the MCP Apps path — the `ui/initialize` response is received at line 1524 but its contents are deliberately discarded — or is the native card's theme meant to stay `auto` for now?

workers/src/tools/_chart_widget.ts
```diff
@@ -759,6 +1155,148 @@ const WIDGET_HTML = `<!doctype html>
...
+      // otherwise the requested height as before.
+      var fitted = aspectHeight(imgNaturalW, imgNaturalH);
+      committedHeight = fitted !== null ? fitted : h;
+      setFrameHeight(committedHeight);
+      frame.classList.remove("hidden");
+      // Reflow on column changes (host sidebar toggle, window resize, mobile
+      // rotation). Without this the frame keeps its mount-time height and the
+      // empty band comes back at the new width. Skipped once a downgrade has
+      // hidden the frame, and after an embed-reported height has taken over.
+      if (fitted !== null) {
```
robertabbott: The re-fit listener is registered only `if (fitted !== null)`, and `fitted` is computed once at delivery time. `aspectHeight` returns `null` when `clientWidth` is 0 (line 614), which is the normal pre-layout state of a freshly mounted iframe — so a widget that receives its tool-result before layout settles gets **no resize listener at all**, and stays on the `structuredContent.height` fallback for the rest of its life even once the column width becomes known.

That is the failure this block's own comment describes ("the frame keeps its mount-time height and the empty band comes back at the new width"), just reached by a different route. Is the one-shot registration intended, or should the listener attach unconditionally and let `aspectHeight` decline per-event as it already does at line 1261?

I could not find a test either way — grepping `widget-dom.test.ts` for a dispatched `resize` event returns nothing, and the only resize path exercised is `deliverEmbedHeight`, a different handler.

workers/src/tools/_chart_widget.ts
```diff
@@ -276,6 +448,44 @@ function buildFallbackWidgetHtml(embedUrl: string, message: string): string {
...
+ * rewrites `</script` to `<\/script`. The HTML parser then no longer sees a
+ * closing tag, while JS still reads the string as `</script>`. Skipping that
+ * step truncates the entire widget bundle at this line, which is how it was
+ * found: every executing widget test failed at once.
+ *
+ * `__FALLBACK_HEIGHT__` is substituted at upgrade time. It covers the window
+ * where Card.js has mounted but not laid out, so `offsetHeight` reads 0 and the
+ * host would otherwise collapse the card to nothing.
+ */
+const NATIVE_HEIGHT_REPORTER = [
```
robertabbott, line 471: `MAX_INLINE_WIDGET_HEIGHT_PX = 500` is documented at line 285 as a hard constraint of the host rather than a preference — "Claude renders MCP Apps cards inline at up to ~500 px and does NOT give the card its own scrollbar, so anything past that is CROPPED, not scrollable — a taller chart silently loses its bottom edge (axis labels, source line)." The PNG path honours it in both directions: `max-height` + `object-fit: contain` in CSS, and a `Math.min(..., 500)` clamp on what it notifies (line 1086).

This reporter notifies raw `document.documentElement.offsetHeight` with no clamp, and after `document.open()` the CSS that enforced the ceiling is gone with the old document. So the native card — which carries strictly more than the PNG (range selectors, chart/table toggle, the source line) and is therefore likely taller — looks exempt from the ceiling. Is it, or would it lose the same bottom edge the 500 px cap exists to protect, on the very elements the PR verified were rendering?

workers/test/widget/widget-dom.test.ts
```diff
@@ -438,6 +528,360 @@ describe("interactive iframe capability probe (executed)", () => {
...
+
+  it("keeps the PNG when the fetch rejects", async () => {
+    const m = mountWidget(staticWidgetHtml());
+    stubFetch(m, () => Promise.reject(new Error("network down")));
+    deliverNative(m);
+    fireImageLoad(m);
+    await new Promise((r) => setTimeout(r, 10));
+    expect(widgetImg(m).getAttribute("src")).toBe(DATA_URL);
+    expect(m.widgetWin.document.body.innerHTML).not.toContain("REAL CARD");
+  });
```
robertabbott, line 886: The PR body says "502, timeout, non-HTML, network error — each costs an upgrade and leaves today's chart untouched, asserted one test per failure mode." I read every test in this describe and found three of the four: `keeps the PNG when the proxy errors` (502), `keeps the PNG when the fetch rejects` (network), `keeps the PNG when the body is not an HTML document` (non-HTML). I could not find one for **timeout** — nothing here installs a fetch stub that never settles and advances past the 7 s `setTimeout` in `upgradeToNativeCard` (`_chart_widget.ts:916-920`). Is it asserted somewhere I missed?

The untested mode is also the one with real ordering risk: after the timer sets `settled = true` it does not abort the fetch, so a **late-arriving** response still runs `.then`, and the only thing keeping it from calling `document.open()`/`document.write()` over a working PNG is the `if (settled) return` on line 927. Would a test that stubs `fetch` with a deferred promise, advances fake timers past 7 s, *then* resolves with `NATIVE_HTML`, and asserts the image src is still `DATA_URL` and the body still lacks the card markup, be worth adding? That pins the guard rather than the happy path.

workers/src/tools/_chart_widget_native.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      origin: "https://mcp.tako.com",
+      pubId: "abc123",
+    });
+    expect(meta?.native_card_url).toBe(
+      "https://mcp.tako.com/embed-html/abc123",
+    );
+    expect(meta).not.toHaveProperty("image_data_url");
+    expect(meta).not.toHaveProperty("image_natural_width");
+  });
+});
```
robertabbott, line 105: The aspect-ratio fix is the user-visible half of this PR, and every test that exercises it (`widget-dom.test.ts:582/608/618`) hands `image_natural_width/height` in by hand. The code that **produces** those numbers — `parsePngDimensions` (`_chart_widget.ts:1600-1611`) and `fetchPngDimensions` (`:1640`) — has no test at all: `parsePngDimensions` is exported specifically for tests via `__chart_widget_test_only__` (`:1991`) and nothing imports it, and `fetchPngDimensions` is untested too.

So the hand-rolled header parser (`getUint32(0) !== 0x89504e47`, width at byte 16, height at byte 20, the `< 24` length bail, the zero-dimension bail) and the ranged wrapper's `status !== 206` / abort / non-PNG paths are all unpinned. A transposed offset there yields a plausible-but-wrong aspect and reproduces exactly the empty-band bug this PR fixes — silently, and server-side. Given the test-only export already exists for it, would a few byte-level cases (valid 24-byte header → 2400x1101, non-PNG magic → undefined, 12-byte buffer → undefined, zero width → undefined) plus a `fetchPngDimensions` case for 200-with-full-body and for a non-206 error be cheap enough to add here?

workers/scripts/skills.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        expect(parsed.name).toBe(name);
+        expect(typeof parsed.description).toBe("string");
+        expect((parsed.description as string).length).toBeGreaterThan(40);
+      });
+
+      it("writes the description as a folded block scalar", () => {
+        // Enforces the convention, not just the outcome: a plain scalar that
+        // happens to parse today breaks the moment someone adds a colon,
+        // which is precisely how this shipped broken once already.
+        expect(markdown).toMatch(/^description: >-$/m);
```
robertabbott, line 102: The docblock names the recurrence mechanism precisely — b12e4c6 fixed this "only in the README's copy-paste blocks for manual claude.ai upload, never in the SKILL.md files the plugin actually installs" — so the drift between the two copies *was* the shipped regression. It then claims the fix restores parity: "The frontmatter here is byte-identical to the README's blocks as a result, which makes the two copies diffable."

They are byte-identical today (`README.md:443-447` vs `skills/tako-web-traffic/SKILL.md:1-5`, same at README 350 and 528), but nothing asserts it — neither `gen-registry.ts` nor `gen-schemas.ts` reads `skills/`. So the one failure mode the docblock calls proven-to-recur is the one this suite would stay green through: reword a description in `SKILL.md`, leave the README stale, and nothing fails. Would a fourth assertion — extract each skill's ```markdown block from `README.md` and `expect(readmeBlock).toBe(skillFrontmatter)` — be the check the docblock is actually arguing for?

Separately, on the docblock's central claim: I verified it rather than taking it on trust. Node was unavailable, so I loaded the repo's own `js-yaml@4.1.0` dist under JavaScriptCore and ran this file's `frontmatterOf` + `yaml.load` against the base-commit `SKILL.md` for all three skills. `tako-web-traffic` throws `YAMLException: bad indentation of a mapping entry (2:54)`; the other two parse cleanly with full `name` and `description`. So the claim holds for the file it names — but this assertion is a no-op for the other two, and only the `/^description: >-$/m` convention test covers them. Would "verified to fail on the pre-fix `tako-web-traffic/SKILL.md`" be the more accurate wording?

llms-full.txt
```diff
@@ -17,7 +17,7 @@ Hosted endpoint: `https://mcp.tako.com/mcp` (Streamable HTTP).
 - Config-file hosts (Cursor, Windsurf, VS Code, or Claude Code via `claude mcp add`): pass `Authorization: Bearer <TAKO_API_TOKEN>`. Get a token at tako.com -> account settings -> API tokens.
```
robertabbott, line 17: Both of this file's default-surface sentences were updated, but the sibling copies were not, and none of them is guarded.

`llms.txt:16` still says `visualize` is "already on by default for ChatGPT", and `:24` repeats "on by default for ChatGPT" in the capability list. `agent.json:100` still ends "Opt-in (enable with ?tools=visualize; on by default for ChatGPT)." I checked whether anything would catch either: `gen-registry.ts` resolves `METADATA_PATH`, `REGISTRY_PATH`, `LOBEHUB_PATH`, `BARREL_PATH`, `LLMS_FULL_PATH`, `README_PATH`, `SKILL_PATHS`, `SUBMISSION_PATH` — `llms.txt` and `agent.json` appear in none, and neither is in the `workers-ci.yml` `paths:` filter. Since `llms.txt` and `llms-full.txt` are both files agents fetch to learn the surface, is it worth updating them here and adding `llms.txt` to `assertLlmsFullCoverage`'s inputs so the two cannot diverge again?

The rename left a matching trail in code comments: `WIDGET_CLIENT_DEFAULT_ON_TOOL_NAMES` was `CHATGPT_DEFAULT_ON_TOOL_NAMES`, and the old name survives at `workers/src/tools/_optional.ts:29` (a file whose own header calls itself "the single source of truth for that mapping"), `AGENTS.md:68` (which also still points at `mcp.ts` rather than `_surface.ts`), and `workers/src/index.test.ts:542` / `:617`. None is reachable by `tsc`.

workers/wrangler.jsonc
```diff
@@ -73,6 +73,25 @@
       "vars": {
         "DJANGO_BASE_URL": "https://staging.tako.com",
         "PUBLIC_BASE_URL": "https://staging.tako.com",
+        // STAGING ONLY — arms the `script-src` capability experiment. Origin
+        // of the CDN serving Tako's front-end assets: the real chart renderer
+        // (`Card.js` + its lazily imported chunks) and the Geist/Inter fonts.
+        //
+        // Setting it does two things and nothing else: adds this origin to the
```
robertabbott: This comment says setting the var "does two things and nothing else: adds this origin to the chart widget's declared `_meta.ui.csp.resourceDomains`, and ships a probe ... See `nativeCardProbeMeta`." Neither half still holds after this PR.

`_chart_widget.ts:1899-1914` deliberately declares **our** origin rather than the CDN — its own comment says "Declaring the CDN here would be dead weight." And `nativeCardProbeMeta` has no callers: `grep -rn nativeCardProbeMeta src test scripts` returns only its definition at `_chart_widget.ts:163` and the two docblocks pointing at it. Worth noting `e3b8366`'s message states "Removed: `nativeCardProbeMeta`, `runCspProbe`, the `cdn_probe_*` `_meta` fields, the status-line DOM, and their tests" — the other four went, this one did not.

What the var now turns on is materially larger: two public HTTP routes, a `connectDomains` CSP entry, and `_meta.native_card_url`. Since this is the operator-facing description of a production-affecting switch — and the deciding input for the "do not copy it to production yet" instruction three lines down — would you update it, along with the matching docblock at `workers/src/env.ts:49-71` which says the same thing?

workers/wrangler.jsonc
```diff
@@ -73,6 +73,25 @@
...
+        //
+        // Deliberately ABSENT from the production block. Do not copy it there
+        // until the experiment has answered — a `resourceDomains` entry is a
+        // script-src grant, so it should not widen prod for a question staging
+        // can settle.
+        // Staging's asset distribution — NOT production's. The embed page
+        // references the CDN of its own environment, and a mismatch is silent:
+        // zero URLs get rewritten, the card mounts, and no chart ever draws.
+        // The proxy logs loudly when it rewrites nothing, for exactly this.
+        "PUBLIC_CDN_URL": "https://d1iyjvzoctsna.cloudfront.net",
```
robertabbott, line 103: Could this distribution id get a second look against the staging CloudFront console before merge? `d1iyjvzoctsna` is 13 characters; production's `d12w4pyrrczi5e` is 14 (`workers/src/icons.ts:32`, which is a live-serving value, plus both new test fixtures at `embed_proxy.test.ts:29` and `_chart_widget_native.test.ts:27`). Both lengths are legal for CloudFront, so this is not evidence of a typo — but it is the only cloudfront host in the repo that is not the known-good production one, and it has no test.

The reason I would rather it be checked than discovered: the comment two lines up says the failure is silent — "zero URLs get rewritten, the card mounts, and no chart ever draws" — and the only detector is the `console.warn` at `embed_proxy.ts:300`, which fires per request *after* the response has gone out. Staging is the only place this experiment runs, so a wrong value here costs the whole verification cycle rather than failing fast. Is there a cheap way to make it loud instead — a first-request assertion, or a smoke check asserting `rewrites > 0`?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+  const cacheControl = response.headers.get("cache-control");
+  headers.set(
+    "cache-control",
+    cacheControl ?? "public, max-age=3600",
+  );
+  headers.set("access-control-allow-origin", "*");
+  // Fonts are subject to CORS even in `<link>`/`@font-face` form, and the
+  // stylesheet that requests them is now same-origin with the page, so this
+  // header is what keeps Geist/Inter from falling back to a system font.
+  headers.set("timing-allow-origin", "*");
```
robertabbott, line 793: Two smaller things on this handler, both about matching the repo's own prior art.

**Caching.** `icons.ts` proxies this same CloudFront distribution and uses `cf: { cacheEverything: true, cacheTtl }` on its fetch, with the comment "so subsequent worker invocations skip the CloudFront round-trip." This route does not, so every widget render pays a full origin round-trip per asset — plus, for JavaScript, a full buffer-and-string-replace of a ~1.5 MB body each time. Is there a reason the same `cf` hint would not apply here?

**Path shape.** `icons.ts` also chose a whitelist over a pattern, explicitly "so a request for `/icons/../../etc/passwd` or any path Tako hasn't shipped under `images/` returns 404 immediately, without a round-trip to CloudFront." This route takes an arbitrary path against the same origin. Given the rewrite only ever targets `archive/` (and the fonts under it), would confining `assetPath` to that prefix cost anything real? It would also make the reflected-content-type question on line 419 mostly moot.

On the comment itself: `timing-allow-origin` governs Resource Timing API visibility, not whether a cross-origin font loads — `access-control-allow-origin` on the line above is what does that. Is the header needed for something else, or is the attribution just off?

workers/src/index.ts
```diff
@@ -45,6 +46,22 @@ export default {
...
+    if (embedProxied !== undefined) return embedProxied;
+
+    // Companion passthrough for the assets that page references (Card.js, its
+    // lazily imported chunks, the Geist/Inter fonts). Tako's CDN answers CORS
+    // for `tako.com` only, and a `type="module"` script is always fetched in
+    // CORS mode, so serving them from our own origin is what makes the native
+    // card work inside a sandbox. Same gate, same fall-through to 404.
+    const assetProxied = await handleCdnAssetProxy(request, env);
```
robertabbott, line 62: Both new routes land above the rate-limited surface and carry no limiter of their own — `FREE_TIER_RATE_LIMITER` / `FREE_TIER_GLOBAL_RATE_LIMITER` are consulted only inside `handleMcpRequest`. So `/cdn-asset/{path}` is an unauthenticated, unmetered, `ACAO: *` mirror of a CloudFront distribution, and for a JavaScript body it buffers and does a full-body string replace before responding — an upstream fetch plus a 2x in-memory copy per request. `/embed-html/{pub_id}` is the same shape against Tako's own origin: an unauthenticated caller can drive `/embed/{id}/` renders at whatever rate they like, and the 300 s cache only helps for ids they repeat.

Since both bindings are already in `Env` and already keyed per client IP, is there a reason not to run these two through the same limiter — or is the `PUBLIC_CDN_URL` staging gate considered sufficient containment until this reaches production?

workers/src/embed_proxy.ts
```diff
@@ -0,0 +1,489 @@
...
+      // silent: the card mounts and never draws. Loud, because the symptom is
+      // not.
+      console.warn(
+        `[mcp] embed proxy rewrote 0 CDN urls for ${pubId} — PUBLIC_CDN_URL (${cdnBase}) likely does not match the origin this page references`,
+      );
+    }
+  }
+  // Logged because a silent stop in stripping is how a csrfToken would begin
+  // reaching a third-party sandbox unnoticed after an upstream redesign.
+  if (!sanitized.removedCsrf) {
```
robertabbott: This fails open on the property the test file names as invariant #3 ("NOTHING SENSITIVE CROSSES INTO THE SANDBOX"). Every non-match path in `sanitizeEmbedHtml` returns `whole` — `JSON.parse` throws (lines 149-154), the parse yields a non-object (line 155), or `csrfToken` is absent (line 159) — and the only consequence here is a `console.warn` before the document, token still in it, is served with `access-control-allow-origin: *` and handed to a third-party sandbox.

The conservative-non-mangling argument in the module header is persuasive for the *document*, but it extends silently to the *token*: an upstream key rename or a whitespace change that breaks `JSON.parse` would start forwarding the token with nothing but a log line to say so. Given the route's stated fallback is that a failure costs an upgrade and nothing else, would returning a 502 when `removedCsrf` is false *and* the body still contains `csrfToken` be the safer default — so the failure mode is a missing upgrade rather than a forwarded token?

workers/src/embed_proxy.test.ts
```diff
@@ -0,0 +1,461 @@
...
+    vi.stubGlobal("fetch", fetchMock);
+    const res = await handleCdnAssetProxy(req(ASSET), ON);
+    expect(res?.status).toBe(200);
+    expect(String(fetchMock.mock.calls[0]?.[0])).toBe(
+      `${ON.PUBLIC_CDN_URL}/archive/abc/vite_dist/assets/Card.js`,
+    );
+    expect(res?.headers.get("access-control-allow-origin")).toBe("*");
+  });
+
+  it("rejects a path trying to climb out of the origin", async () => {
```
robertabbott: This asserts a weaker property than the name promises, in two ways.

`expect([400, undefined]).toContain(res?.status)` accepts a decline as a pass, but with the experiment ON every input listed here can only return 400 — `undefined` would mean the prefix check failed, which is a different bug. Would `toBe(400)` pin the intended behavior?

More to the point of "climb out of the origin": none of these four exercises the case the guard misses. `assetPath` comes from `url.pathname`, which the URL parser leaves percent-encoded, so `includes("..")` is a literal check against still-encoded text — but the WHATWG parser treats `%2e%2e` (and `.%2e`, `%2e.`, case-insensitively) as a double-dot segment when it parses the upstream string at `embed_proxy.ts:395`. So `/cdn-asset/%2e%2e/%2e%2e/x` passes the check and is then resolved as traversal by `fetch`. It is harmless today — the origin is config-fixed and every path on it is already reachable through this route — but the test claims a confinement that is not there, and it would keep passing if a prefix restriction (per the line 419 thread) were added later and relied on this check. Would adding `%2e%2e` cases, and normalizing via `new URL(...).pathname` before the guard, make the assertion match the guarantee?

.github/workflows/workers-ci.yml
```diff
@@ -1,23 +1,35 @@
...
+      - 'version.txt'
   pull_request:
     paths:
       - 'workers/**'
       - 'registry/**'
+      - 'skills/**'
+      - '.claude-plugin/**'
       - 'chatgpt-app-submission.json'
       - 'llms-full.txt'
+      - 'version.txt'
```
robertabbott, line 43: The new comment states the rule as "Every path below is covered by a guard this workflow runs, so each one has to trigger it" — but the property that actually prevents a skipped guard is the converse: every file a guard *reads* must be a trigger path.

By that reading `README.md` is missing: it is read by `assertPinFormInDocs` inside `registry:check` (`gen-registry.ts` passes `README_PATH`), and it is not in `paths:`. So a PR touching only `README.md` — which is most of what this PR changes outside `workers/**` — skips that guard entirely. Should `README.md` join the list (along with `llms.txt`, if it becomes guarded per the `llms-full.txt` thread), or is the omission deliberate because README edits are expected to ride along with worker changes?

workers/src/index.test.ts
```diff
@@ -424,6 +424,21 @@ describe("worker routing", () => {
       "ui/resourceUri": "ui://tako/embed/chart",
```
robertabbott, line 475: `index.ts` now awaits `handleEmbedProxy` and `handleCdnAssetProxy` on every request that gets past the icon route, but I could not find a `SELF.fetch` test here for `/embed-html/{pub_id}` or `/cdn-asset/...`. All coverage of those routes is handler-level in `embed_proxy.test.ts`, which calls the functions directly and so cannot catch a wiring mistake — ordering against the `/mcp` branch, or the fall-through to the catch-all 404.

The wrangler test config has no `PUBLIC_CDN_URL` (it was added only under the `staging` env block), so a `SELF.fetch("https://example.com/embed-html/VKd7qE8K9Ba16kMFENNQ")` expecting 404 would pin the "invisible in production" property end to end in one line. Given that is listed as property #1 of the feature in `embed_proxy.test.ts`'s own header, is it worth adding here?

workers/src/index.ts
25eliu: Yes — done, and thank you for tracing the ordering. `resolveProxyOrigins` now catches and declines, so a misconfigured binding disables the experiment exactly as an unset one does, which is where the blast radius was before these routes existed.

It resolves `PUBLIC_BASE_URL` through the same catch, since the embed route interpolates that origin too and it throws on the same invariants. Both routes are treated as one feature — the embed page is useless without its assets — so any bad origin turns both off together.

Tested with a trailing slash, an unparseable value, a `javascript:` scheme, and `ftp://`; plus a case for the malformed `PUBLIC_BASE_URL` path. Also added `SELF.fetch` coverage asserting `/health` and `/mcp` still answer, so a regression that made these throw rather than decline fails there too.

workers/src/embed_proxy.ts
25eliu: Changed to `text/plain; charset=utf-8`, plus `Content-Security-Policy: sandbox` and `X-Frame-Options: DENY` — both of the things you offered, since neither costs anything.

You were right that nothing needs it to be HTML: the only consumer reads `r.text()` and its sole shape check is `indexOf("<html")`, so the content type is invisible to it. The upstream gate still requires `text/html` from Tako; it is only what we re-serve that changed.

The cookie/DCR chain you laid out is the reason I did not want to leave this on a "no browser will navigate here" argument.

workers/src/embed_proxy.ts
25eliu: Not a firm no — so both fixes are in.

I checked rather than reasoned about it: `archive/`, `user-images/`, `content-style/` and `static/` all return the same S3 `AccessDenied` through `d1iyjvzoctsna`, i.e. the distribution fronts the bucket root, not an `archive/`-only origin path. That does not prove an uploaded `.html` is served as one, but it does not refute your reading either.

`assetPath` is now confined to `archive/`, and the reflected content type is allow-listed to script/style/font/image/json/text with everything else forced to `application/octet-stream`. `text/html` and `image/svg+xml` are both deliberately off the list.

Confinement turned out to be free, which I could only tell by pulling a real embed page: 15 of its 16 CDN references are under `archive/`, and the sixteenth is `previews/<hash>.png` appearing only in a `<meta twitter:image>` and inside `<noscript>` — never fetched with JS enabled. So `rewriteCdnUrls` is now scoped to `archive/` too, leaving those pointing at the CDN where they belong rather than at a route that would decline.

workers/src/tools/tako_search.ts
25eliu: Correct on both, and the PR body is fixed.

Keeping the fetch: ChatGPT renders `embed_url` in a cross-origin iframe it cannot measure, so without the card true aspect its frame falls back to a fixed height and keeps the empty band — which is the bug this branch is for. But you are right that it is a production behavior change the flag does not gate, and the body claimed otherwise. Both exceptions are now named there: `image_natural_*` on ChatGPT and `interactive_probe` on Claude.

The docblock wording is fixed too. You are right that the render still has to happen before any byte range can be served, so the `Range` saves the ~170 KB transfer and not the latency. It now says that explicitly and points the next reader at `PNG_HEAD_FETCH_TIMEOUT_MS` as the number that governs the worst case, to be sized against render latency rather than 64 bytes of transfer.

I do not have p50/p99 for the ChatGPT path — no production access from here — so I have not put numbers in the body I cannot source. Flagged for staging measurement instead.

workers/src/embed_proxy.ts
25eliu: Fixed, and your reachability analysis was right — I confirmed it against a real production embed page. The `server-config` island is exactly `{staticPrefix, csrfToken, userLoggedIn, hostContext, developerConsoleEnabled}`, all server-derived, so this is latent rather than live as you said.

I went with re-escaping the stringified output rather than the textual delete. The reasoning: `<` / `>` / `&` are valid JSON escapes that decode to exactly the same string, so the island means the same thing byte-for-byte to any consumer that parses it, while the HTML tokenizer can no longer see a closing tag — which is precisely the invariant Django is asserting. A targeted textual replace of the member would have to handle arbitrary JSON string escapes inside the value to find its end correctly, which felt like the more regressable of the two.

The test uses your exact payload and asserts four things: the token is gone, no `</script>` survives so the island did not terminate early, and the parsed `title` still equals the original string.

workers/src/embed_proxy.ts
25eliu: Agreed, and changed to `redirect: "manual"` on both. In workerd that surfaces the 3xx as a non-`ok` response, so it falls into each route existing error branch and becomes a 502.

Your framing is the part I want to keep: the guarantee held for the request we issue and not the response we serve. Tested across 301/302/307/308 on both routes, plus an assertion that each fetch is actually issued with `manual` — otherwise the test would pass on a mock that never redirects.

workers/src/embed_proxy.ts
25eliu: Confirmed and fixed — the cap never engaged.

The bound is now read off the accumulated body rather than a declared header, which covers all three of the gaps you named at once: the stripped `Content-Length` on the auto-decompressed response, the `NaN` fall-through, and the compressed-vs-decompressed mismatch.

You were also right about the test being the reason this survived. The over-cap test now supplies no `content-length` at all and uses a genuinely oversized body, and there is a companion asserting an under-cap JS body with no `content-length` still gets rewritten — so the header-absent path, which is the real one, is the tested one in both directions.

workers/src/tools/_chart_widget.ts
25eliu: Both real, both fixed.

The empty-string guard: exactly as you traced it. `withHostTheme` now returns the url unchanged when it is falsy, before the `typeof` check, so `""` can no longer become `"?dark_mode=true"` and clear the downstream guard. That covers line 1142 as well, since both call sites go through the same helper. The test needs a host theme to reproduce, so it mounts with `window.openai = {theme: "dark"}` and omits `embed_url` to force the PNG path, then asserts zero fetches.

On the second half — you are right, and there is no host-theme source on the MCP Apps path to plumb. `ui/initialize` carries none and the spec defines none. So rather than imply the fix reaches Claude, the comment now says plainly that `window.openai.theme` is the only source, that the native card on claude.ai stays `dark_mode=auto`, and that the proxy allow-list is forward-looking — exercised by ChatGPT committed-iframe url today, ready for the MCP Apps path if a host ever declares a theme.

workers/src/tools/_chart_widget.ts
25eliu: Not intended — fixed. The listener is registered unconditionally now and `aspectHeight` declines per-event, which it already did.

Your route to it is the one that matters: `fitted` is computed once at delivery time, and a tool-result landing before layout settles means `clientWidth` is 0, no listener, and the fallback height for the rest of the widget life. Three tests now: a plain resize re-fit, the pre-layout case (deliver with no width, assert the fallback, then set the width and fire `resize`), and a resize that still has no width declining rather than throwing. The pre-layout one fails against the previous code.

workers/src/tools/_chart_widget.ts
25eliu: It was exempt, and it should not have been — it would lose the same bottom edge, on exactly the elements the PR verified were rendering. Fixed.

The reporter now measures unscaled (clearing any fit from a previous pass first, so the reading is always the natural height and cannot oscillate), and past the cap it scales the document to fit and reports the cap. Widening to `100/s`% before the transform keeps the content filling the frame. That is the document equivalent of the `object-fit: contain` the PNG gets — scaled-and-complete over full-size-and-truncated, which is the same trade the constant docblock already argues for.

Two tests, both through the real `document.write` path: a 600 px fallback reports 500 and carries `scale(0.8333…)` with `transform-origin: top left`, and a 300 px one reports 300 with no transform at all.

workers/test/widget/widget-dom.test.ts
25eliu: You did not miss it — there were three, not four. Added, and the PR body corrected.

I took your suggested shape for the second one, since you are right that it is the guard worth pinning rather than the happy path: stub `fetch` with a deferred promise, fire the bail, THEN resolve with a good `NATIVE_HTML`, and assert the image src is still the data URL and the body still lacks the card markup.

One note on what these prove. Both pass against the pre-fix bundle — the timeout handling and the `settled` guard were already correct, just unasserted. I checked that explicitly rather than assume, by running the new tests against the original `_chart_widget.ts`; the three behavior tests in this round fail there and these two do not. Worth stating so they are not read as fixes.

Rather than fake timers, the widget own `window.setTimeout` is patched to capture the 7000 ms timer, so the bail fires deterministically and no other timer is affected.

workers/src/tools/_chart_widget_native.test.ts
25eliu: Added, all of them — and you are right that this is the wrong thing to leave unpinned, given a transposed offset reproduces the exact bug the branch fixes, silently and server-side.

`parsePngDimensions`: the three real card aspects (2400x1101, 2400x1257, 2400x1845) off a synthesized IHDR, a 24-byte header and a 23-byte one to pin the `< 24` bail in both directions, JPEG/GIF/HTML-error-page magic, empty and 12-byte buffers, and zero width / zero height / both.

`fetchPngDimensions`: 206, 200-with-full-body, an assertion that the request actually carries `Range: bytes=0-63`, 404/500/502, a JSON body, and both a network throw and an `AbortError`.

workers/scripts/skills.test.ts
25eliu: Added, and it is the check the docblock was arguing for — thank you for naming it that way.

Shape differs slightly from your suggestion because the README blocks are raw `---` fences rather than ```markdown blocks, so it matches on `^---\nname: <skill>\n…\n---$` and compares to `frontmatterOf(SKILL.md)`. Verified it catches drift by inducing some: appending a word to the SKILL.md description fails only that skill assertion, and the other two stay green.

On the wording — you are right, and it now says so. The docblock states that the YAML assertion was verified to fail on the pre-fix `tako-web-traffic/SKILL.md`, and that it is a convention check rather than a regression test for the other two, which is what the `>-` assertion carries. Nice independent verification, incidentally; your `bad indentation of a mapping entry (2:54)` matches what I get.

llms-full.txt
25eliu: Both updated and both now guarded.

`llms.txt` is wired into `assertLlmsFullCoverage`. It has no `### <tool>` sections, so the function degrades to exactly the right check for an index — every tool must be named somewhere in it. That immediately caught a second gap you did not have to look for: `llms.txt` never mentioned `tako_agent_start` / `tako_agent_wait` at all, so the split pair is documented there now too. Verified the guard bites by removing a mention and watching `registry:check` fail.

`agent.json` had the same stale sentence and is corrected. It is not machine-guarded — I did not want to invent a parity assertion for a file with no generator — but it is small and now consistent.

The rename trail is cleared: `_optional.ts:29`, `AGENTS.md:68` (which also pointed at `mcp.ts` rather than `_surface.ts`), and `index.test.ts:542`/`:617`. No `CHATGPT_DEFAULT_ON_TOOL_NAMES` references remain anywhere in the repo.

workers/wrangler.jsonc
25eliu: Both rewritten. You are right that the description had come loose from what the switch does, and that it matters most because it is the input to the "do not copy it to production yet" line.

`nativeCardProbeMeta` is deleted — it had no callers, and e3b8366 message did claim it was removed.

Both docblocks now say what setting it actually does: brings two public HTTP routes into existence, adds `_meta.native_card_url`, and adds a `connectDomains` entry for our own origin — plus the explicit note that we declare our origin and not the CDN, with the reason. The wrangler comment also lists what to re-read before promoting it to production: both routes are unauthenticated, both are now metered, the asset route is prefix-confined, and the embed route serves inert content types.

workers/wrangler.jsonc
25eliu: Flagging this one as not fully resolved, because I could not close it from here and would rather say so.

What I could check: `d1iyjvzoctsna.cloudfront.net` resolves and is a live CloudFront distribution fronting an S3 bucket — same shape as production `d12w4pyrrczi5e`, `AccessDenied` from S3 rather than a CloudFront error. So it is not a typo for a nonexistent host, and the length difference is not the signal.

What I could not check: whether it is STAGING distribution. That needs the console, and I have flagged it for @25eliu to confirm before the staging verification cycle.

On making the failure loud: I have not added the first-request assertion. The existing `console.warn` on `rewrites === 0` is still per-request-after-the-fact as you say, but with `rewriteCdnUrls` now scoped to the `archive/` prefix, a wrong distribution produces zero rewrites on the very first embed fetch and logs immediately — which for a staging-only experiment is the same signal one request earlier. Happy to add a smoke check if you would still prefer one; it seemed like the wrong thing to bolt on before the id itself is confirmed.

workers/src/embed_proxy.ts
25eliu: All three, and the last one you are simply right about.

Caching: `cf: { cacheEverything: true, cacheTtl }` added to both upstream fetches, matching `icons.ts` against this same distribution. No reason it would not apply — it was an oversight.

Path shape: confined to the `archive/` prefix. It costs nothing, which I could only establish by pulling a real embed page — 15 of 16 CDN references are under `archive/` and the sixteenth is a `previews/` OG image that only appears in `<meta twitter:image>` and `<noscript>`. And you are right that it makes the reflected-content-type question mostly moot; I kept the allow-list anyway as the second lock, since it is a few lines.

`timing-allow-origin`: the attribution was off. `access-control-allow-origin` on the line above is what lets the font load; `timing-allow-origin` only governs Resource Timing visibility. Comment corrected and the header kept for what it actually does, with the font test now asserting ACAO explicitly.

workers/src/index.ts
25eliu: Both routes now run through `FREE_TIER_RATE_LIMITER`, keyed per client IP via the existing `freeTierRateLimitKey`, failing open the same way every other use of that binding does. A 429 rather than a JSON-RPC result, since these are browser subresource fetches with no request id to answer.

We chose not to lean on the staging gate. It is real containment today, but it is containment that expires exactly when someone sets the binding in production — which is the moment this is easiest to forget. The wrangler comment now also lists it among the things to re-read before promoting.

workers/src/embed_proxy.ts
25eliu: Agreed — it now returns 502 when the strip missed AND the body still contains `csrfToken`.

Your framing is what settled it: the conservative-non-mangling argument is right for the document and was silently extending to the token, and the route stated fallback is that a failure costs an upgrade and nothing else. No card is a much cheaper failure than a forwarded credential.

Keyed on the token still being present rather than on `removedCsrf` alone, so an upstream that simply stops shipping the island is a success rather than a spurious 502. Two tests: an unparseable island with a token 502s and the token does not appear in the response body, and a page that never carried one still serves 200.

workers/src/embed_proxy.test.ts
25eliu: Half right, and the half that is wrong is interesting — so let me be precise, because I nearly "fixed" a non-bug.

`toBe(400)`: agreed for the cases that reach the handler, and changed. That test now covers the empty path plus everything outside the `archive/` tree (`user-images/`, `content-style/`, `previews/`, `static/`, and near-misses like `archivex/` and `Archive/`), all asserted `toBe(400)` with a spy confirming no upstream fetch.

`%2e%2e`: the bypass does not exist. The WHATWG path parser resolves `..` segments AND decodes `%2e` in order to recognise them, and it does so during `new URL(request.url)` — before the handler is entered. Measured in workerd: `/cdn-asset/%2e%2e/%2e%2e/x` arrives as `/x`, `.%2e/.%2e/x` as `/x`, and `/cdn-asset/archive/%2e%2e/%2e%2e/user-images/evil.html` as `/user-images/evil.html`. None still carries the route prefix, so all of them fail `startsWith` and fall through to the 404 — which also means a decline is the CORRECT outcome there, not a symptom of a broken prefix check.

I think the reading came from testing the upstream construction rather than `url.pathname`; I made the same mistake first and only caught it when the `toBe(400)` assertion failed on `../secrets`.

So there is no normalization to add — it already happened. What I did add is a test that pins the parser behavior explicitly, asserting the exact normalized pathname for each encoded form alongside the decline, plus a comment in the handler saying that the parser is the mechanism and the `includes("..")` guard is defense in depth. Your forward-looking concern still lands: it means a refactor that builds `assetPath` from a raw string instead of a parsed `URL` would make those guards load-bearing for real, and now it fails loudly instead. (`b%2Fc.js` does survive encoded, incidentally — that is an S3 key containing a slash, not traversal.)

.github/workflows/workers-ci.yml
25eliu: You are right, and the comment stated the weaker of the two readings. Both fixed.

`README.md` added to `paths:`, and it is now the input to two guards rather than one — `assertPinFormInDocs` already read it, and the new skills parity test reads it too. `llms.txt` added as well, now that it is guarded.

The comment now states the property that actually prevents a skipped guard — every file a guard READS must be a trigger path — flags it as the converse of what was written, and lists what reads what so the next person can check the invariant rather than trust the list.

workers/src/index.test.ts
25eliu: Added, and it caught nothing — which given "invisible in production" is property #1, is the result worth having pinned.

Four cases: `/embed-html/{pub_id}` 404s, `/cdn-asset/...` 404s, both preflights 404 rather than advertising CORS, and `/health` + `GET /mcp` still answer normally. That last one is deliberately in the same block: both handlers are awaited ahead of most of the router, so it is the end-to-end version of the `index.ts:54` blast-radius thread — a regression that made them throw rather than decline fails here too.

workers/wrangler.jsonc
25eliu: Resolved — and I should have just checked this instead of flagging it. `d1iyjvzoctsna` is staging's.

`https://staging.tako.com/` references it 26 times, and its own page-context island settles it outright:

```
"staticPrefix": "https://d1iyjvzoctsna.cloudfront.net/archive/ce645f901756abe2529af71c9e1eba759fadc1ce/"
```

`staticPrefix` is the exact value `rewriteCdnUrls` has to match, so this is a stronger check than the console would have been — it confirms the binding against the string the code actually operates on, not just against which distribution the account owns.

That also re-validates the `archive/` confinement on the environment that matters. On staging **every** CDN reference is under `archive/` — no `previews/` OG image, so not even the one benign exception production has.

Dropping the smoke-check question with it: a wrong distribution now produces zero rewrites on the first embed fetch and logs immediately, and the value is confirmed correct, so there is nothing left for a first-request assertion to catch that the existing warn does not.

robertabbott (approved):

PR #207, opened 2026-08-03
feat(oauth): per-zone session cookie name, and email/password sign-in

25eliu:
The auth half of #206, split out so it can land on its own — the widget/theme
work in that PR is unrelated to it and reviews differently. These commits
cherry-picked onto `main` cleanly.

> **Correction.** This originally claimed "zero file overlap between the two
> halves". That was true when the branch was split and is not true now: #206
> has since gained `cf5a583`, which also touches `workers/src/env.ts`, so
> rebasing #206 to drop these commits will conflict there. Trivial to resolve,
> just not zero. Caught in review.

## 1. `fix(oauth)`: send the session cookie under the name the target zone reads

Staging OAuth had been failing as **"Your Tako sign-in expired."** — with a
valid, fresh session.

The Stytch session cookies on the `.staging.tako.com` zone are
name-namespaced (`stytch_session_jwt_staging`), because the browser also
delivers prod's `Domain=.tako.com` cookie — carrying the *default* name — to
`staging.tako.com`, where it would shadow the staging session. Django resolves
the name per request host and reads **only** that name, so this Worker sending
the default name to staging meant Django saw no cookie at all, 403'd, and the
caller surfaced the misleading expiry copy.

- `sessionCookieName(base)` derives the name from the backend origin;
  `mintTakoApiKey` uses it instead of a hardcoded `stytch_session_jwt`.
- The rule mirrors tako's `cookie_name_contract.json`. That contract is the
  source of truth and lives in another repo, so **no shared CI can catch drift
  here** — the doc comment says so explicitly.
- The `unauthorized` branch in `/authorize` now `console.error`s the status.
  A 403 from a cookie-name misconfiguration and a genuinely expired session
  were previously indistinguishable, with nothing in `wrangler tail`. That is
  what let this hide for weeks.

## 2. `feat(oauth)`: email + password sign-in

The sign-in page offered Google and a magic link. Magic link is a poor fit for
an MCP consent flow — the user leaves the browser mid-dance to check email —
and it stranded anyone whose Tako account uses a password.

**`POST /login/password`**, a server-side form post (not the browser SDK), so
the password never passes through the CDN-loaded `js.stytch.com` script and the
session lands via the exact same cookie Google's redirect uses.
`completeStytchLogin` is now shared by both paths, so "how a user becomes
logged in" has one implementation rather than two places for the session claims
or TTL to drift.

Gate ordering is deliberate — method → config → same-site → limiter → field
presence → Stytch — so every check that can reject without a password check
runs before the expensive, attackable one.

- **Rate limiting, fail-closed.** `LOGIN_RATE_LIMITER` (8/60s) is metered on
  two axes per attempt: per client IP *and* per **hashed** email (per-IP alone
  misses a distributed spray at one account; per-email alone misses one host
  walking a list). Unlike the free-tier limiters — whose absence merely reverts
  to a 401 — an unbound or throwing limiter here **503s the endpoint** and
  leaves Google-only, because this is an unauthenticated password check on a
  public host. Declared in all three env blocks including the top-level one, so
  `wrangler dev` isn't silently reduced to Google-only.
- **No account enumeration.** Unknown-email and wrong-password both collapse to
  "Incorrect email or password." Only the two *actionable* states stay
  distinct: `no_user_password` ("this account signs in with Google" — otherwise
  a Google user is stranded on a page that no longer offers a magic link) and
  `reset_password`.
- **`?error=` is a closed set.** Codes map to copy server-side via
  `LOGIN_ERROR_COPY`; unknown codes render nothing. The param is
  attacker-controllable, so echoing free text would make the sign-in page a
  phishing surface. Copy is also `escapeHtml`'d on the way out.
- **MFA is named, not swallowed.** Stytch answers 200 with an
  `intermediate_session_token` and no usable `session_jwt` for second-factor
  accounts. This Worker has no second-factor UI, so that case is detected in
  the shared response parser and told to use Google, rather than falling into
  "something went wrong" and sending the user in circles.
- **Nothing sensitive leaks.** The redirect carries only a code (redirect URLs
  land in history, `Referer`, and proxy logs); logs carry the classification
  and status, never credentials; SDK error messages no longer surface raw
  Stytch exception text to the page.
- Deep links to Tako's real app for password reset and signup, verified against
  `routeTree.gen.ts` — a dead link is the first thing a plugin reviewer clicks.
- Login page restyled to Tako's palette (light + dark), with the mark served
  from the Worker's own `/icons/*` proxy.

## 3. `style(wrangler)`

The cherry-pick pasted the top-level limiter comment verbatim into the staging
and production blocks at top-level indentation, leaving `name` looking like it
sat outside its own object. Both now point at the one authoritative comment.

## Test plan

- [x] `npm run typecheck` — all **four** passes (src, test, scripts,
      test/widget); a bare `tsc --noEmit` misses three projects.
- [x] `npm test` — 1009 tests green (908 worker + 55 widget + 46 scripts),
      including 336 new lines of `handlers.test.ts` covering the limiter
      fail-closed path, a throwing limiter, cross-site rejection, each error
      classification, and `stytch.test.ts` covering the MFA and
      `no_user_password` branches.
- [x] `wrangler deploy --dry-run` on default, `staging`, and `production` —
      config parses and `LOGIN_RATE_LIMITER` is bound in all three.
- [ ] **Staging deploy:** confirm the cookie-name fix by completing a real
      connect against `mcp.staging.tako.com` (this is the bug that only shows
      up on the staging zone).
- [ ] **Staging deploy:** password sign-in end-to-end; a wrong password; a
      Google-only account (expect the "signs in with Google" copy, not
      "incorrect password").
- [ ] Confirm the `LOGIN_RATE_LIMITER` namespace ids (`1020`/`1021`/`1022`)
      don't collide with anything already allocated on the account.

## Note on #206

These commits are still in #206 as well. Once this lands, #206 should be
rebased onto `main` to drop them — say the word and I'll do it.


workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+  no_user_password:
+    "This account signs in with Google. Use “Continue with Google” above.",
+  reset_password:
+    "Your password needs to be reset. Use “Forgot password?” below to get a reset link.",
+  rate_limited: "Too many sign-in attempts. Wait a minute and try again.",
+  server_error: "Something went wrong signing you in. Please try again.",
+  // This Worker has no second-factor UI. Google carries the factor itself, so
+  // it is the way through for an MFA account — say so rather than implying the
+  // password was wrong.
+  mfa_required:
```
robertabbott: This copy sends the user to "Continue with Google", but Google goes `GET /oauth/stytch_callback` → `authenticateStytchToken` → the *same* `parseAuthenticateResponse` you just added the MFA branch to. If Stytch also answers the OAuth authenticate with an `intermediate_session_token` (MFA in Stytch is a policy on the user, not on the primary factor), the callback's `StytchError` branch fires and the user lands on `errorPage(502, "Stytch authentication failed (mfa_required). Please try signing in again.")` — a raw internal code plus an invitation to loop. That is precisely the "send the user in circles" outcome the docstring on the new branch says it prevents.

Has this been confirmed against a real MFA-enabled Stytch account, or should the copy point at Tako's own `/login/mfa` under `resolvePublicBase(env)` instead?

workers/src/oauth/handlers.ts
```diff
@@ -1322,6 +1599,25 @@ export async function handleStytchCallback(
     throw err;
   }
 
+  return completeStytchLogin(req, url, cfg, stytchResult);
```
robertabbott, line 1773: The new `mfa_required` errorType now flows out of `authenticateStytchToken` too, and the branch just above (`:1595`) interpolates it verbatim into a user-facing page: "Stytch authentication failed (mfa_required)." The password path went to the trouble of a closed-set copy table precisely so users never see internal `error_type` strings, but this path bypasses it entirely.

Should this branch route through `loginErrorCode(err)` + `redirectToLogin(url, code)`, so both paths give the same message for the same condition rather than one saying "use Google" and the other showing a 502?

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+  // login or a scanner. `Sec-Fetch-Site` is set by the browser and cannot be
+  // spoofed from page JS; a missing header (older browser, curl) is allowed
+  // through because the limiter below still bounds it.
+  const fetchSite = req.headers.get("sec-fetch-site");
+  if (fetchSite !== null && fetchSite !== "same-origin" && fetchSite !== "none") {
+    return new Response("cross-site form post rejected", { status: 403 });
+  }
+
+  // Fail-closed on the limiter — see `Env.LOGIN_RATE_LIMITER`.
+  const limiter = env.LOGIN_RATE_LIMITER;
```
robertabbott: `handleLoginPassword` 503s when this binding is absent, but `handleLogin` renders the password form unconditionally — it never reads `env.LOGIN_RATE_LIMITER`. In any deployment where the binding is missing (a fork of this config, pulling the `ratelimits` entry as an ad-hoc kill switch, `wrangler dev` against a modified config) the page shows email/password fields whose every submit returns the 503 error page.

It compounds: the inline script's fallback copy at `:1537` is now *"Google sign-in is unavailable. Use your email and password below."* — so an SDK failure plus an unbound limiter points the user at a form that points back at Google.

Should `handleLogin` gate the `<form>` block (and the "or" divider) on the same `typeof env.LOGIN_RATE_LIMITER?.limit === "function"` predicate this line uses, so the two cannot disagree?

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+  const body = await req.formData().catch(() => null);
+  const email = String(body?.get("email") ?? "").trim();
+  const password = String(body?.get("password") ?? "");
+
+  // Meter BEFORE validating fields, on both axes. Per-IP alone misses a slow
+  // distributed spray at one account; per-email alone misses one host walking
+  // a list. The email is hashed so the rate-limit key space never holds
+  // plaintext addresses.
+  const ip = req.headers.get("cf-connecting-ip") ?? "unknown";
+  const emailKey = email === "" ? "empty" : await sha256B64Url(email.toLowerCase());
```
robertabbott: The email axis is metered before field presence is validated, so an attacker does not need a plausible password — 8 junk POSTs a minute against a known address keeps that address's bucket empty and holds the victim in the `rate_limited` redirect, from arbitrary IPs, at essentially zero cost. The copy at `:1213` ("Too many sign-in attempts. Wait a minute and try again") tells them it was their own doing.

Was that lockout accepted deliberately as the price of catching a distributed spray at one account? If so, should the email bucket only be consumed by attempts that actually reach a credential check (meter IP → validate fields → meter email), or the per-email limit be looser than the per-IP one? They are both 8/60s today.

Smaller, on the comment above: an unsalted SHA-256 over an enumerable address space is pseudonymization, not anonymization. Is "the key space never holds plaintext addresses" meant to carry privacy weight, or is it only describing key hygiene?

robertabbott (commented):
Reviewed the auth split on its own. The security posture holds under scrutiny — I could not build a working forced-login (a cross-site POST is rejected by `Sec-Fetch-Site`, and on a header-less client the `SameSite=Lax` state cookie means `completeStytchLogin` bails before it appends the session `Set-Cookie`), `?error=` is a genuine closed set with no path from URL text to the document, no credentials reach logs or redirect URLs, and `sessionCookieName` matches `cookie_name_contract.json` case-for-case against `app/backend/auth/stytch/cookies.py`. So the headline claims check out.

The comments below are mostly about the reasoning the PR is built on rather than about exploits. Three things are worth reading first.

**The `mfa_required` advice may be a dead end** (inline on `handlers.ts:1217`). It points the user at Google, but Google goes through the same `parseAuthenticateResponse` you just added the MFA branch to.

**The limiter is weaker than the fail-closed decision assumes** (inline on `handlers.ts:1299`). `workers/README.md` §Measured behaviour reports a cold burst admits ~115 requests *regardless of the configured limit*, and the binding counts per colo — so an 8/60s bucket is a dampener, which `wrangler.jsonc:76` says outright while the handler comment claims the per-email axis covers a distributed spray. Given the endpoint refuses to serve at all when the binding is absent, it would help to have those two comments agree and to write down what the real bound is (which Stytch lockout, at what threshold).

**Two findings that are not inline, because the files are outside the diff:**

- `workers/src/freetier.test.ts:590,614` — the repo already has the mechanical check for the test-plan item you left unchecked, but `allNamespaceIds` is scoped to `/"name":\s*"FREE_TIER_[A-Z_]*"/` and asserts a hardcoded `toHaveLength(6)`, so 1020/1021/1022 are invisible to it and the suite stays green either way. Same for the `period === 60` guard in `allPeriods`. Can both regexes widen to every `ratelimits` entry, with the expected count derived from the parsed config rather than a literal? That turns "confirm the namespace ids don't collide" into a CI assertion instead of a manual step, and stops the *next* limiter being silently uncovered the same way.
- Docs the change left behind: `handlers.ts:15` (module-header flow diagram) still reads `user picks Google / magic-link`; `handlers.ts:1167` (`handleLogin` docstring) still says "email magic-link entry points"; `README.md:243` says "a sign-in page (Google or email magic-link)" and the `docs/images/mcp-tako-signin.png` screenshot two lines below still shows the old page; `workers/README.md:29` binding table omits `LOGIN_RATE_LIMITER`, and its rollout checklist has no "disabling password sign-in" entry — the free tier has a no-redeploy kill switch (`wrangler secret delete FREE_TIER_API_KEY`), this has none, and pulling the `ratelimits` entry instead leaves the form rendering with every submit 503ing (see the inline comment on `handlers.ts:1281`).

One PR-body claim does not hold: "Zero file overlap between the two halves" — `workers/src/env.ts` is touched by both `505771d feat(oauth)` and `cf5a583 feat(widget): a dev-only lever to bust the host's widget cache` in #206, so the rebase will conflict there. Trivial to resolve, just not zero.

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+        "Use “Continue with Google”.",
+    );
+  }
+
+  const body = await req.formData().catch(() => null);
+  const email = String(body?.get("email") ?? "").trim();
+  const password = String(body?.get("password") ?? "");
+
+  // Meter BEFORE validating fields, on both axes. Per-IP alone misses a slow
+  // distributed spray at one account; per-email alone misses one host walking
```
robertabbott: This says per-IP alone "misses a slow distributed spray at one account", but `wrangler.jsonc:76` states the binding counts **per colo**, and `workers/README.md` §Measured behaviour reports that a cold burst admits **~115 requests regardless of the configured limit** (`limit: 1` leaked ~118, the same as `limit: 10`). A distributed spray is distributed across colos by definition, so the per-email bucket gives roughly `8 × N_colos`/min plus a ~115-request cold leak per colo — the exact scenario this axis is claimed to cover is the one per-colo counting defeats.

`wrangler.jsonc` is the more candid of the two ("Stytch's own lockout is the real bound"), and the two comments now contradict each other. Since the endpoint **fails closed** on the premise that this limiter is load-bearing, could the rationale here be reconciled with that — and is the real bound (which Stytch lockout, at what threshold) written down anywhere?

workers/src/oauth/handlers.ts
```diff
@@ -1182,54 +1369,159 @@ function jsStringLiteral(s: string): string {
...
+  // `jsStringLiteral`, not HTML-escaped — they live inside <script>, not in
+  // HTML attributes or text. `errorMessage` is the reverse: it lands in HTML
+  // text, so it goes through `escapeHtml`. It is already a value from the
+  // closed `LOGIN_ERROR_COPY` set, so this is belt-and-suspenders.
+  const safeError = escapeHtml(errorMessage);
+  // Deep links into Tako's real web app, which owns account creation, password
+  // reset, and MFA. Routes verified against `routeTree.gen.ts`
+  // (`/login/forgot-password`, `/signup`) — a dead link here is precisely the
+  // kind of thing a plugin reviewer clicks first.
+  const webBase = resolvePublicBase(env);
```
robertabbott: This moves `/login` from "no env dependency" to "throws if the public origin is unresolvable": `resolvePublicBase` runs inside `loginPage`, `handleLogin` has no try/catch, and `validatePublicOrigin` throws on an absent value *and* on a trailing slash.

All three env blocks set `DJANGO_BASE_URL` today, so the fallback saves it. But a future `PUBLIC_BASE_URL: "https://tako.com/"` edit in the production block takes the entire sign-in page to a bare 500 — **including Google sign-in**, which is the one path the limiter's fail-closed design goes out of its way to preserve.

Should the two footer links be wrapped so a bad origin degrades to omitting "Forgot password?" / "Create an account", rather than killing the page?

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+ * The split mirrors Tako's own `PasswordSignInPage.tsx`:
+ * `invalid_credentials` deliberately absorbs unknown-email AND wrong-password
+ * so the page cannot be used to enumerate accounts, while `no_user_password`
+ * and `reset_password` stay distinct because they are the two cases where the
+ * user needs to be told to do something different.
+ */
+const LOGIN_ERROR_COPY: Record<string, string> = {
+  invalid_credentials: "Incorrect email or password.",
+  missing_fields: "Enter both your email and password.",
+  no_user_password:
```
robertabbott: The comment above this table says the split means the page "cannot be used to enumerate accounts", but `no_user_password` and `reset_password` are both returned by Stytch for an *existing* user with *any* password — so a single POST with a junk password separates "account exists (Google-only)" / "account exists (needs reset)" / "nothing here".

`PasswordSignInPage.tsx` can arguably afford that because it sits behind the product; this is an unauthenticated endpoint on the OAuth host. Should these two also collapse into `invalid_credentials` here — the user still recovers via "Forgot password?" and "Continue with Google", both already on the page — or is the enumeration acceptable on this surface for a reason I am missing?

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+  }
+  const cfg = readConfig(env);
+  if (cfg === null) return oauthDisabledResponse();
+  const url = new URL(req.url);
+
+  // CSRF: the form is same-origin, so a cross-site POST is either a forged
+  // login or a scanner. `Sec-Fetch-Site` is set by the browser and cannot be
+  // spoofed from page JS; a missing header (older browser, curl) is allowed
+  // through because the limiter below still bounds it.
+  const fetchSite = req.headers.get("sec-fetch-site");
```
robertabbott, line 1376: The comment says a missing header "is allowed through because the limiter below still bounds it" — but a limiter bounds volume, not forgery, and a forced login needs exactly one request.

Tracing it, what actually saves this is that `tako_oauth_state` is `SameSite=Lax` (`cookies.ts:48`), so a cross-site POST arrives without it and `completeStytchLogin` bails at the missing-state branch *before* it appends the session `Set-Cookie`. That ordering is load-bearing and nothing states or tests it — someone refactoring `completeStytchLogin` to build its response headers first would silently reintroduce forced login for header-less clients.

Two questions: should the comment name the real control, with a test asserting a header-less cross-site POST sets no session cookie? And since `same-site` is already rejected (so a `*.tako.com` subdomain is not trusted), is there a reason not to require a matching `Origin` header — which every browser capable of mounting this does send — instead of allowing the header-less case at all?

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+/**
+ * Map a Stytch failure to one of `LOGIN_ERROR_COPY`'s codes.
+ *
+ * Stytch reports wrong-password as `invalid_credentials`/401 and unknown-email
+ * as `email_not_found`/`user_not_found`/404; both collapse to
+ * `invalid_credentials` here. Only the two actionable states survive as
+ * themselves.
+ */
+function loginErrorCode(err: StytchError): string {
+  switch (err.errorType) {
```
robertabbott, line 1322: Of the nine outcomes in this switch, the tests reach five. `password_reset_required`, `user_not_found`, and `password_does_not_match` have no test, and the `default:` branch has none on either side of its ternary.

The 5xx side matters most: a Stytch outage is the failure users actually hit at scale, and nothing pins that a 500 from `/v1/passwords/authenticate` yields `error=server_error` rather than being misclassified as `invalid_credentials` and telling everyone their password is wrong during an upstream incident.

Also uncovered nearby: an absent `sec-fetch-site` (the deliberately permissive path at `:1275`), a non-form-encoded body (`req.formData().catch(() => null)`), and the route itself — `handleLoginPassword` is only ever called directly, so `index.test.ts` would not catch it being dropped from the dispatch chain.

workers/src/oauth/handlers.ts
```diff
@@ -1164,9 +1177,183 @@ export function handleLogin(req: Request, env: Env): Response {
...
+  }
+  return completeStytchLogin(req, url, cfg, stytchResult);
+}
+
+/**
+ * Bounce back to the sign-in page with a closed-set error code. Never carries
+ * the email or password — a redirect URL lands in browser history, `Referer`
+ * headers, and proxy logs.
+ */
+function redirectToLogin(url: URL, code: string): Response {
```
robertabbott, line 1493: `completeStytchLogin` still requires `tako_oauth_state`, whose TTL is `STATE_COOKIE_MAX_AGE_S = 10 * 60` (`cookies.ts:24`) and which is minted back at `/authorize`. Before this PR the only interactive step inside that window was a Google redirect.

Now the user can mistype, get `invalid_credentials`, retry, hit `rate_limited` and be told "Wait a minute and try again" — after which a *successful* password login can still land on `errorPage(400)` "the original authorization request was lost", discarding the session it just minted.

Was the 10-minute TTL re-checked against the new retry loop, and should `redirectToLogin` re-sign the state cookie with a fresh `exp` so the retry path cannot outlive it?

workers/src/oauth/handlers.test.ts
```diff
@@ -995,6 +996,359 @@ describe("/login", () => {
...
+      );
+      const html = await res.text();
+      expect(html).not.toContain(attempt);
+      expect(html).not.toContain("<img src=x");
+      // The error slot renders empty and CSS hides it (`.err:empty`).
+      expect(html).toContain('id="err" role="alert" aria-live="polite"></div>');
+    }
+  });
+
+  it("renders the mapped copy for a code it did issue", async () => {
```
robertabbott, line 1068: Three of the tests that pin this PR's headline safety properties pass for the wrong reason:

- **This one** asserts `html` contains `"Continue with Google"` — but that is the label of the Google button `loginPage` renders on every request (the test at `:1036` relies on the page always containing it). It passes with `LOGIN_ERROR_COPY` emptied entirely. Should it assert the distinguishing part, `"This account signs in with Google"`, inside the `id="err"` div?
- **`"fails closed when the rate limiter binding is missing"` (`:1124`)** asserts only `status === 503` and that `fetch` was not called. `oauthDisabledResponse()` satisfies both, as the test at `:1101` proves. Should it assert the body (`"Password sign-in is unavailable on this deployment"`) so a regression in `readConfig` cannot make it pass for the wrong reason?
- **`"meters on both the client IP and the email"` (`:1178`)** asserts `keys.length === 2`, that some key contains the IP, and that no key contains the plaintext address. `limit({ key: "login:email:constant" })` — a single global bucket that lets one attacker lock every account out of password sign-in — satisfies all three. Should it call the handler with two different emails and assert the two email keys differ, and that the same email twice yields the same key?

Smaller, `"blocks when the limiter itself throws"` (`:1147`) asserts the `location` header but not the status, unlike its sibling at `:1133`. A response carrying that header with a 200 is not a redirect at all.

workers/src/index.ts
```diff
@@ -158,6 +159,13 @@ export default {
     if (url.pathname === "/login") {
       return handleLogin(request, env);
     }
+    // Email + password sign-in. A server-side form POST rather than a
+    // browser-SDK call, so the password never passes through the CDN-loaded
+    // Stytch script and the session lands via the same cookie Google's
+    // redirect uses. Must be matched BEFORE any `/login` prefix handling.
```
robertabbott: "Must be matched BEFORE any `/login` prefix handling" — but this block sits *after* the `/login` check, and there is no prefix handling anywhere in this router: every OAuth route is an exact `url.pathname === ...` compare, and the earlier prefix matchers (`/icons/`, `/cdn-asset/`, `/embed/`) cannot match `/login/password`. The comment states an ordering constraint that neither exists nor is satisfied by the placement.

Drop it, or replace it with the real one — exact-match router, so any future `startsWith("/login")` handler must go below this line?

workers/src/oauth/handlers.ts
25eliu: Confirmed, and it was worse than a dead end — the callback branch at `:1595` rendered `errorPage(502, "Stytch authentication failed (mfa_required)")`, so the raw `error_type` reached the user.

Not verified against a real MFA-enabled Stytch account (I can't create one), so I fixed it in the way that's safe under either behaviour rather than betting on one: the copy now points at Tako's own `/login/mfa` under `resolvePublicBase`, which is the only place the second factor can actually be completed. If Stytch does report the factor on the OAuth path too, the advice still holds.

workers/src/oauth/handlers.ts
25eliu: Yes — done. That branch now returns `redirectToLogin(url, loginErrorCode(err))`, so both paths share one copy table and the same condition can't produce two different messages depending on which factor the user picked. No internal `error_type` reaches a page from either path now.

workers/src/oauth/handlers.ts
25eliu: Gated. `hasLoginLimiters(env)` is now the single predicate: `handleLoginPassword` fails closed on it, and `handleLogin` uses it to decide whether to render the form and divider at all, so the two can't disagree.

The compounding case you spotted is fixed too — `sdkFallback` is server-rendered, so when there's no form the SDK failure says "Please try again in a moment" instead of pointing at a form that points back at Google. Tests cover both directions (form hidden without the bindings, present with them).

workers/src/oauth/handlers.ts
25eliu: Not deliberate — you found a real hole, and the copy blaming the victim made it worse. Fixed both parts:

**Order is now** per-IP meter → validate fields → per-email meter. Only attempts that reach a real credential check can spend someone else's bucket. The IP axis is still charged first, so fieldless spam isn't free to whoever sends it. Test: a POST with an empty password charges the IP axis once and the email axis **zero** times.

**The email bucket is now looser than the IP one** — a separate `LOGIN_EMAIL_RATE_LIMITER` binding at 30/60s vs 8/60s, because a `ratelimits` limit is per-binding so one binding can't hold two limits. It's the axis an attacker can aim at a victim, so it gets the slack.

On the hash: you're right, and I've dropped the privacy claim. Unsalted SHA-256 over an enumerable address space is pseudonymization, not anonymization — the comment now says it's key hygiene (the key space doesn't hold addresses in the clear) and claims nothing more.

workers/src/oauth/handlers.ts
25eliu: You're right, and the contradiction was mine — `wrangler.jsonc` was the honest one. Reconciled rather than papered over.

The handler comment no longer claims the per-email axis covers a distributed spray. It now says what's true: `ratelimits` counts per colo, this repo's own measurements show a cold burst admitting ~115 regardless of the configured limit, so **both axes are dampeners plus a `wrangler tail` signal** — and the real bound on guessing one password is Stytch's per-user lockout. Same framing now in `env.ts` and a new `workers/README.md` section, so the three can't drift apart again.

On why it still fails closed: "dampened and logged" vs "wide open and silent" are different postures on an unauthenticated password check, which is the honest reason — not that the limiter is a hard bound. That's now written down as the rationale.

**What I could not write down is the threshold**, because it's configured in the Stytch dashboard, not in this repo — I have no way to read your project's policy. `workers/README.md` says explicitly that Stytch's lockout is the real bound and that anyone changing these numbers must check that policy too. If you can tell me the actual setting I'll put the number in the doc.

workers/src/oauth/handlers.ts
25eliu: Fixed — wrapped, and it degrades exactly the way you suggested.

`safePublicBase(env)` catches, logs, and returns `null`; `loginPage` then omits the two footer links and the whole page still renders. Test asserts `PUBLIC_BASE_URL: "https://tako.com/"` gives a 200 that still contains "Continue with Google" and no "Forgot password?".

Your framing is what made this worth doing: a trailing slash taking out **Google** — the one path the fail-closed design goes out of its way to preserve — would have been a bad way to find out.

workers/src/oauth/handlers.ts
25eliu: You're not missing anything — the claim in that comment was too strong, and I've corrected it. But I'm **keeping both codes distinct**, deliberately, so this is a judgement call rather than an oversight.

The trade-off, stated plainly in the new comment: with the magic link gone, a passwordless (Google-signup) account told "incorrect email or password" has no way to discover that Google is its way in. That stranding is the thing this PR exists to fix, and it's a certainty for real users; the enumeration is a leak of a fact that Tako's own sign-in page already exposes the same way. Both cases are also recoverable from the page as it stands ("Continue with Google", "Forgot password?").

What changed is the honesty of the comment. It no longer says the split means the page "cannot be used to enumerate accounts" — it now says the split **does** let one junk-password POST learn that an address exists, and why that's accepted.

Happy to collapse them if you'd rather have the enumeration resistance on this surface; it's a two-line change to `loginErrorFor`. Your call.

workers/src/oauth/handlers.ts
25eliu: Both done — and you traced the real control correctly.

**The comment now names it:** a limiter bounds volume, not forgery, and a forced login needs exactly one request. What actually saves it is that `tako_oauth_state` is `SameSite=Lax`, so a cross-site POST arrives without it and `completeStytchLogin` bails at the missing-state branch *before* appending the session `Set-Cookie`. The comment says the ordering is load-bearing and points at the test.

**The test exists now:** "allows a header-less POST but never lets it mint a session" — sends a POST with no `sec-fetch-site` and a *successful* Stytch response mocked, then asserts no session cookie comes back. That's the assertion that would catch the refactor you describe.

On requiring `Origin`: I left the header-less case allowed. Reasoning — the property that actually prevents forced login is the state cookie, which this test now pins, and every browser capable of mounting the attack sends `Sec-Fetch-Site` anyway, so an `Origin` requirement would mainly reject non-browser clients (curl, some corporate proxies that strip fetch metadata) for no security gain we don't already have. If you'd rather have belt-and-braces I'll add it — it's cheap, I just didn't want to claim it buys something it doesn't.

workers/src/oauth/handlers.ts
25eliu: All covered now. Added a table-driven test over every mapped `error_type` plus **both** sides of `default:`:

- `password_reset_required`, `user_not_found`, `password_does_not_match` — each pinned
- `default:` credential-ish side — unmapped type at 401 and at 404 → `invalid_credentials`
- `default:` outage side — **`internal_server_error`/500 and a body with no `error_type` at all/503 → `server_error`**, which is the row you flagged as mattering most: nothing previously stopped a Stytch outage from telling every user their password was wrong

Also added the three nearby gaps: absent `sec-fetch-site` (with the no-session-cookie assertion), a JSON body that `req.formData()` can't parse (→ `missing_fields`, no unhandled rejection), and the route wiring — `index.test.ts` now asserts `/login/password` doesn't fall through to the catch-all 404, and that a GET returns 405 before any config gate.

workers/src/oauth/handlers.ts
25eliu: Re-checked, and you're right that this page got easier to reach — it's now reachable after a **successful** sign-in, which is the part that makes it bad.

I did **not** re-sign the state cookie on retry. Re-signing would let one authorization request live indefinitely across repeated retries, which widens the window `STATE_COOKIE_MAX_AGE_S` exists to bound — a worse trade than the rough edge it fixes. The 10 minutes stays.

What changed is the copy, which was the actively misleading part. It now leads with the sign-in having worked, names the 10-minute window, says nothing is wrong with the account, and tells the user to restart the connection from the app they were connecting (noting they won't have to sign in twice). Test asserts all three of those and that the old "authorization request was lost" phrasing is gone.

Flagging the residual honestly: a user who burns 10 minutes in the retry loop still has to restart from the host. That's a known rough edge, not a fixed one.

workers/src/oauth/handlers.test.ts
25eliu: All four correct, and this was the most useful comment in the review — three tests that pinned the PR's headline safety claims while passing for the wrong reason.

1. **Error copy** — now asserts the distinguishing half *inside* the error slot: `/id="err"[^>]*>[^<]*This account signs in with Google/`. Fails with `loginErrorFor` emptied, which the old assertion didn't.
2. **Fail-closed** — now asserts the body contains "Password sign-in is unavailable on this deployment", so `oauthDisabledResponse()` can no longer satisfy it. Added a sibling that binds only the IP limiter, to pin that *either* missing binding trips it.
3. **Metering** — replaced with a helper that runs an attempt and returns the keys per axis. Now asserts two **different** emails produce two **different** keys, and the same email twice produces the **same** key (or the bucket never accumulates). Your `login:email:constant` counterexample fails it.
4. **Throwing limiter** — status assert added; a `location` header on a 200 isn't a redirect.

workers/src/index.ts
25eliu: Correct on both counts — the constraint didn't exist and the placement didn't satisfy it. Replaced with the real one:

> Every OAuth route here is an EXACT `pathname ===` compare, so this needs no particular position relative to `/login` above. Any future `startsWith("/login")` handler would have to go below this line.

robertabbott (approved):

PR #208, opened 2026-08-03
fix(widget): Claude renders the real interactive card, themed to the host

25eliu:
Everything here came out of dogfooding the widget on claude.ai. OAuth work is in #207; this branch touches no OAuth file.

## What was broken

**1. The card ignored Claude's theme.** `hostTheme()` read `window.openai.theme` and nothing else, which is `null` off ChatGPT, so both chart urls stayed on `dark_mode=auto` — resolved from `prefers-color-scheme` inside the widget iframe, i.e. from the **OS**. A light Claude on a dark Mac rendered a dark card on a light surface.

The comment shipped in #203 called this unfixable: *"the `ui/initialize` response carries none, and the spec defines none."* That's wrong. MCP Apps 2026-01-26 puts `hostContext.theme` on the `ui/initialize` **response** and defines `ui/notifications/host-context-changed` for updates. Confirmed live from a real session — claude.ai sends `{theme: "dark"}`.

Reading it asynchronously is safe because of handshake ordering: a host MUST NOT send notifications before our `initialized`, which we only send once the init response lands, so the theme is known before any `tool-result`. `window.openai.theme` keeps priority where both exist.

**2. White corners under the card.** The card paints its own surface at `border-radius: 8px` over a transparent `html`/`body`, so the corners fall through to the document backdrop. Measured in Chrome:

| frame | canvas |
|---|---|
| same-origin, nothing declared | composites over the parent — corners already correct |
| cross-origin / out-of-process | opaque **WHITE** ← claude.ai, the reported bug |
| document declares `color-scheme: dark` | opaque at **Chrome's #121212**, not transparent |

That third row is why the first attempt at this was wrong, and the reason there are two tiers rather than one:

1. **The host's own surface colour** from `hostContext.styles.variables`. Exact — the canvas becomes indistinguishable from the host's page. Safe anywhere, ungated.
2. **`color-scheme` from the declared theme.** Approximate, and **gated to the MCP Apps path**. It only beats WHITE, and white is the measured claude.ai behaviour; on a host that already composites transparently it would *add* a visible square.

Applied to both documents, since the native path replaces the widget's wholesale: on the widget's own `<html>`, and injected last into the native card's markup so it beats the page's own rules.

**ChatGPT is deliberately untouched.** It exposes a theme, but its frame's base isn't measured, and there the chart is a *nested* cross-origin iframe whose document we can't style anyway — so there's no upside worth gambling a visible box for. It keeps today's transparency and is still themed via the `dark_mode` rewrite on its iframe url, which is what actually colours its card. Two tests pin that. Unknown clients never get a widget at all, so they have no corners to expose.

**3. `tako_answer` rendered no chart at all.** Its citations *are* cards and their `card_id`s were in the output, but only search lifted the widget fields. Three things were missing: the fields (now via a shared `topCardChartFields` that `buildSearchOutput` also uses, so they can't drift again), `...autoChainShape` on the **advertised** schema (the SDK rebuilds it strict at the top level, so lifting without declaring is a no-op), and the `appUiResource`/`extraMeta`/`extraContentBlocks` hooks.

**4. Widget changes were invisible in testing.** Hosts cache the widget **by URI**, and claude.ai's cache outlives removing and re-adding the connector. Three rounds of the corner fix were tested against a bundle from before any of them — the console gave it away, logging `native card upgrading {bytes}` where the code logs `{bytes, scheme, surface}`. `WIDGET_URI_SUFFIX` mints a fresh URI on demand. Unset in staging and prod, where the URI must stay permanent.

## Costs and trade-offs

- Answer's advertised `outputSchema` grows **353 bytes (~88 tokens)** for the six widget fields. Measured, not estimated.
- **Widget hosts pay one added serial round-trip on every answer call.** Measured against prod, n=21 across 7 distinct cards: p50 **0.277 s**, p90 **0.374 s**, max 0.464 s. A cold render measured ~0.84 s (single sample). Nowhere near the 8 s `PNG_FETCH_TIMEOUT_MS` ceiling.
- **The render is cached per `pub_id`+`dark_mode`, so answer mostly hits.** Read from the backend rather than assumed: with no `width`/`height`/`hideFooter` the endpoint takes `ImageGenerationManager`'''s managed-cache path, and `ImageRequest.request_hash` keys on `pub_id:dark_mode:for_powerpoint:width:height:hide_footer:glanceable`. Search and answer both send `DEFAULT_DARK_MODE` and no dimensions, so their hashes are identical — a card search already charted is a hit, and a concurrent search+answer on the same card JOINS one render (`should_poll` on a matching in-flight hash) rather than starting two. Render rate therefore grows only by cards answer cites that search never charted.
- **Non-widget hosts (Cursor, Windsurf, Gemini CLI, plain API) now get an inline base64 PNG on every answer call** — `extraContentBlocks` fires when `ui === undefined` — typically ~170 KB plus up to 8 s of serial fetch, on the tool the description calls START HERE. Deliberate (an answer's citations are cards), and a larger number than the schema bytes.
- Tier 2 on an MCP Apps host that *would* have composited transparently trades nothing for a faint `#121212` square. Only claude.ai is measured, and there the alternative is white.
- A mid-conversation theme toggle re-themes from the next tool call onward, not retroactively: the native upgrade `document.open()`s the page, discarding our listeners.
- **Production is unaffected by the theme work**: `PUBLIC_CDN_URL` is staging-only, so the PNG is still the chart there and stays `dark_mode=true` until the native card is promoted.

## Review findings fixed in `f3aad74`

Found by re-reading this branch, not by tests failing:

- `mcpHostSurface` was declared *below* the function assigning it. `var` hoisting made it work, but the hoisted value is `undefined`, not `null` — so `!== null` would have read as a real colour and painted `undefined`. Declaration moved up; guards switched to truthiness.
- `safeCssColor` accepted `rgb()` with an empty body: passes the allow-list, assigns nothing, and *looks* like tier 1 succeeded while silently suppressing the fallback.

## Testing

- `vitest run` 883, widget suite 75, scripts 46 — all pass
- all four tsconfig projects clean; `schemas:check` and `registry:check` pass
- every fix written test-first, each new test confirmed RED first
- verified live through a tunnel against prod: `tako_answer` returns `pub_id`, `embed_url`, `native_card_url` on the request origin, the baked PNG, and a per-call `ui.resourceUri` — the same payload search returns
- the theme rewrite verified end to end in a browser with a local MCP-host harness: host declares light → widget requests `dark_mode=false`

### Not verified

- **Whether claude.ai sends `styles.variables`.** The live session showed `theme` only, so tier 1 doesn't fire there yet and you land on tier 2 — white becomes near-dark, with a faint square possible. If Claude never ships those variables, the follow-up that needs no host cooperation is to read the mounted card's own background and paint the canvas with it.
- ChatGPT was not re-tested on a live account; the claim is that it is unchanged, which the two gate tests assert structurally.


workers/src/tools/_chart_widget.ts
```diff
@@ -1560,19 +1742,48 @@ const WIDGET_HTML = `<!doctype html>
...
+    // \`dark_mode=auto\` in place.
     if (
       fromHost &&
       msg.jsonrpc === "2.0" &&
       msg.id === INIT_REQUEST_ID &&
       (msg.result !== undefined || msg.error !== undefined)
     ) {
+      if (msg.result && typeof msg.result === "object") {
+        mergeHostContext(msg.result.hostContext);
+        applyHostSurface();
```
robertabbott, line 1883: The ordering argument in the comment at 693-698 — and in the PR body — is that a spec-compliant host MUST NOT send notifications before our `initialized`, so `hostContext` is known before any `tool-result`. That establishes less than it looks like, because line 1825 sends `initialized` on a 200 ms timer regardless of whether the init response landed. So the precondition is not "the host is spec-compliant", it is "the host answers `ui/initialize` within 200 ms".

The comment above that timer says the window was shortened *because* claude.ai "appear[s] to start attempting tool-result delivery within ~hundreds of ms of widget mount" — i.e. this host was measured delivering on the same timescale as the handshake it is now racing. For a host answering in ~350 ms: we post `initialized` at 200, the host may legitimately post `tool-result` at 210, and its init response arrives at 350. postMessage preserves per-source order, so the tool-result is dispatched first and `render()` (1384) runs with `mcpHostTheme === null`.

That is not recoverable for the card: `render()` latches `rendered`, the image `load` handler calls `upgradeToNativeCard(withHostTheme(opts.nativeCardUrl), h)` at 1322, `withHostTheme` returns the url unchanged when `hostTheme()` is null, `nativeCardUrl()` emits no `dark_mode` at all, and `nativeStarted` blocks a retry. The init response landing at 350 then only re-runs `applyHostSurface()` — right corners, OS-derived card. That is TAKO-3781 again, on exactly the first-mount timing the 200 ms change was made for.

Would it work to hold the first `render()` for an outstanding init response until the 200 ms deadline, or to re-derive the native url once `hostContext` lands? At minimum, should the comment state the real precondition rather than "spec-compliant"?

workers/src/tools/_chart_widget.ts
```diff
@@ -973,11 +1122,41 @@ const WIDGET_HTML = `<!doctype html>
...
+        // exact surface colour is safe anywhere, while bare \`color-scheme\`
+        // (Chrome's own #121212) only beats WHITE and would read as a visible
+        // square on a host that composites transparently — so it is limited to
+        // the MCP Apps path, where the white base is the measured behaviour.
+        //
+        // In practice this path is MCP-only anyway: ChatGPT commits to the
+        // nested iframe and never reaches the native upgrade. The gate is
+        // written explicitly so it stays correct if that changes. Injected LAST
+        // so it wins over the page's own rules.
+        var scheme = mcpHostSurface ? hostTheme() : mcpHostTheme;
```
robertabbott: The host theme is read twice here, on opposite sides of the fetch, and nothing pins the two reads together. The card markup's theme is fixed at line 1322 by `withHostTheme(opts.nativeCardUrl)` before `fetch` is called; `scheme` and `surface` for the injected `<style id="tako-host-scheme">` are read here, inside the `.then()`, up to 7 s later.

If an init response or a `host-context-changed` lands in that window, `mergeHostContext` mutates `mcpHostTheme`/`mcpHostSurface` and the document written at 1161 carries a card body rendered for the old theme (or, in the race on the other comment, for no theme at all) wrapped in a `:root{background:…;color-scheme:…}` block for the new one. `nativeStarted` means there is no second fetch to reconcile it.

Should `scheme`/`surface` be snapshotted at 1322 next to the url so the written document is internally consistent, or should a context change during the fetch re-issue the upgrade?

workers/src/tools/_chart_widget.ts
```diff
@@ -650,23 +675,147 @@ const WIDGET_HTML = `<!doctype html>
...
+  //
+  // Both are no-ops when the host says nothing at all — that is the case where
+  // transparency already works (same-origin), and painting anything would
+  // introduce the opaque square the transparent surface exists to avoid.
+  function applyHostSurface() {
+    try {
+      var root = document.documentElement;
+      // Tier 1 — exact, and safe to apply on ANY host: the canvas becomes the
+      // host's own colour, so it cannot look like a box.
+      if (mcpHostSurface) {
```
robertabbott, line 858: Tier 1 is described as "exact, and safe to apply on ANY host", and it is the one tier with no gate. Two things make me want to push on that.

First, this repaints a surface that was deliberately unpainted. `1687dcd` ("transparent surface and host-matched theme") removed a `#0f1115` body background with this reasoning: it "was chosen so iframe-default white would not show through the chart card's rounded corners, and it solved that by filling the whole widget — which on a light-themed host is a dark box with square corners sitting inside the host's rounded container." Tier 1 is the same shape of fix, with the host's colour instead of a Tako colour. That is a real improvement, but it only holds if `--color-background-primary` is the colour immediately behind the widget. It is specified as the host's *page* background, and Claude mounts widgets inside a bordered container that is typically an elevated surface — and `readHostSurface` falls back to `--color-background-secondary`, treating the two as interchangeable. If either is a different shade from what is actually behind us, the square box from `1687dcd` is back, now on every host including ChatGPT.

Second, the PR is explicit that this tier has never been observed firing: "Whether claude.ai sends `styles.variables` … the live session showed `theme` only." So the ungated tier is the unverified one.

Given the log at 1761 already fires on the path that would answer this, would recording whether `styles.variables` arrived — and which candidate matched — turn the open question into something one dogfooding session settles? Right now it logs `{theme: mcpHostTheme}` only, which is the half already known.

workers/src/tools/_chart_widget.ts
```diff
@@ -650,23 +675,147 @@ const WIDGET_HTML = `<!doctype html>
...
+
+  // Spec: "Views merge received fields with their current context state rather
+  // than replacing it entirely." So a partial update that omits \`theme\` must
+  // leave the known theme alone rather than clearing it.
+  function mergeHostContext(hostContext) {
+    if (!hostContext || typeof hostContext !== "object") return;
+    var t = normalizeTheme(hostContext.theme);
+    if (t !== null) mcpHostTheme = t;
+    var surface = readHostSurface(hostContext);
+    if (surface !== null) mcpHostSurface = surface;
```
robertabbott, line 745: The merge semantics are right for `theme` — the spec language is quoted and there is a test for the displayMode-only update. Is the surface safe under the same rule, though? `mcpHostTheme` is theme-independent, but `mcpHostSurface` *is* a theme-dependent value, and `readHostSurface` returns null when a partial update carries no `styles`, so the old colour survives.

A host that toggles dark to light and sends `{theme: "light"}` with no `styles` therefore leaves the dark surface in `mcpHostSurface`. `applyHostSurface` then takes tier 1 and paints the dark colour with `color-scheme: light` — a dark canvas on a light host, which is the bug this PR is fixing, mirrored. Should a theme change that carries no new surface clear `mcpHostSurface` rather than keep it?

The theme-survives-a-partial-update case is tested at widget-dom.test.ts:1289; the mirror — a surface learned at initialize surviving a later theme-only update — does not appear to be.

workers/src/tools/_chart_widget.ts
```diff
@@ -650,23 +675,147 @@ const WIDGET_HTML = `<!doctype html>
...
+
+  // This value arrives over postMessage and is interpolated into a \`<style>\`
+  // block injected into the native card document, so it is an injection sink.
+  // Allow-list a colour grammar instead of trying to escape one: hex, and the
+  // functional notations, with no semicolons, braces, quotes, or \`url(\`.
+  // Anything else is dropped and we fall back to \`color-scheme\`.
+  function safeCssColor(v) {
+    if (typeof v !== "string" || v.length === 0 || v.length > 64) return false;
+    if (/^#[0-9a-fA-F]{3,8}$/.test(v)) return true;
+    return /^(rgb|rgba|hsl|hsla|oklch|oklab|lab|lch)\\(\\s*[0-9a-zA-Z.,%\\/\\s+-]+\\)$/.test(v);
```
robertabbott, line 783: `f3aad74` added the `rgb()` rejection because an empty functional colour "passes the allow-list, assigns nothing, and *looks* like tier 1 succeeded while silently suppressing the fallback". Does one space get back through? `\s` is inside the body class as well as the `\s*` prefix, so `rgb( )` satisfies `[0-9a-zA-Z.,%/\s+-]+` and returns true, where `rgb()` returns false.

The consequence is the one the original finding described: `mcpHostSurface` is truthy, `applyHostSurface` takes tier 1, `upgradeToNativeCard` emits `:root,body{background:rgb( )}` which the CSS parser drops, and tier 2's `color-scheme` never runs because it is in the `else` arm. Same for `rgb(url)`, `lab(image-set)` and `#1234567` — all shape-valid, none colours.

Since the property being checked is really "does this parse as a colour", would round-tripping the candidate through CSSOM — assign to a throwaway element's `style.background`, require a non-empty read-back — close the whole class rather than the one spelling? The allow-list would still gate the injection surface first.

workers/src/env.ts
```diff
@@ -190,6 +190,18 @@ export interface Env {
...
+   * `appUiResourceUri`). Hosts cache the widget BY URI and claude.ai's cache
+   * outlives removing and re-adding the connector, so during local development
+   * every widget change is otherwise invisible — you end up debugging a bundle
+   * that is not running. Set it to a timestamp to force a re-read.
+   *
+   * MUST stay unset in staging and production: the shipped URI is effectively
+   * permanent (Claude desktop caches it beyond connector lifecycle, so renaming
+   * it 404s previously-installed sessions).
+   */
+  WIDGET_URI_SUFFIX?: string;
```
robertabbott, line 204: Two questions on the lever, both about coverage rather than the mechanism.

The suffix reaches the static uri via `appUiResourceUri` at `_chart_widget.ts:2245`, but `APP_UI_TEMPLATE_URI_PATTERN` is a plain constant, so the dynamic variant registered at 2189 and returned at 2246 stays `ui://tako/embed/chart/{pub_id}` regardless. Resolution is fine either way — the static resource is registered under the same suffixed string it is read by. But the template path serves a different builder (`buildBakedWidgetHtml`) and is the uri written into the per-call `_meta.ui.resourceUri`, so it cannot be busted. It works on claude.ai today only because that host reportedly ignores the per-call override. Should `dynamic.uriPattern` and the resolver derive from `appUiResourceUri(env)` too, so the lever covers both bundles and any host that does honour the override?

Separately, and more consequential: this doc comment establishes that hosts cache the widget by uri beyond connector lifecycle, and that the shipped uri must stay permanent. Taken together, how is a bad widget recalled from production? A redeploy does not reach a session holding a cached bundle, and this var is documented as must-stay-unset there, so it is not the escape hatch either. Is the answer that the blast radius is bounded because `PUBLIC_CDN_URL` is absent from the production block, or is a real kill path worth having before the native card is promoted?

workers/src/tools/tako_answer.ts
```diff
@@ -325,6 +336,10 @@ const takoAnswer = {
             ),
           }
         : {}),
+      // Chart for the top cited card — shared with tako_search via
```
robertabbott: This makes answer render a chart, but the text that tells models so was not updated — and one line now actively says the opposite. The `DESCRIPTION` at line 60 ends "…alongside machine essentials (request_id, usage, guidance)" where `tako_search.ts:62` ends "…(request_id, usage, chart-widget fields)", and line 57 routes callers away: "Reach past it only for a different job: `tako_search` for breadth recon **and chart cards**".

That string is not only a prompt. `workers/scripts/gen-registry.ts` copies it verbatim into `registry/server.json` and `registry/lhm.plugin.json`, so it is the published discovery text on the MCP registry and the LobeHub listing. Because the string is unchanged the generated artifacts stay byte-identical and `registry:check` passes — the drift is in the content, not the codegen.

The same search-only framing is in `llms-full.txt:53`, `llms.txt:20-21`, `README.md:16` and the `tako_answer` row at `README.md:267`. `assertLlmsFullCoverage` (gen-registry.ts:105-140) only checks that each tool is mentioned and that its section names every *input* param, so none of this can fail CI.

Should these land in this PR? Beyond accuracy, the description is what stops a model re-posting `image_url` as a markdown image and double-rendering the card.

workers/src/tools/tako_answer.ts
```diff
@@ -325,6 +336,10 @@ const takoAnswer = {
...
+      // `topCardChartFields` so the two cannot render differently. Yields `{}`
+      // when nothing is citable, which is also the zero-card case.
+      ...topCardChartFields(cards, ctx.env),
```
robertabbott: "so the two cannot render differently" is true of the field construction, but `cards[0]` is not the same construct in the two tools, so "top card" can still mean two things.

`citesByPosition` at line 312 skips `orderCardsByUsefulness` whenever the prose carries a `[N]` marker — a deliberate and well-argued trade for citation correctness, but it now also decides which card gets charted. In that branch `cards[0]` is the backend's order; in `buildSearchOutput` it is always the usefulness order. Same card set, different chart, and the demoted-Overview-card case the comment at 303 names is exactly what would show up.

There is a second one that survives even without markers: answer orders the RAW cards at 314 and slims afterwards, while search slims first and orders the slimmed cards (`tako_search.ts:276`). `orderCardsByUsefulness`'s primary signal is `latestDataPoint`, which reads `content.dataset.rows`/`content.records`, and `slimCardContent` blanks those when `capRows === null` — i.e. on `include_contents: false`. So on that input search ranks every card as row-less and falls through to freshness, while answer still ranks on row dates.

Worth either sharing the ordering step too, or softening the comment to say the url construction is shared while top-card selection is per-tool?

workers/src/tools/tako_answer.ts
```diff
@@ -337,6 +352,33 @@ const takoAnswer = {
   slimStructured(output) {
     return slimAnswerStructured(output as unknown as AnswerFullOutput);
   },
+  // Widget plumbing, identical to tako_search's. Rationale for each gate lives
+  // there; the only reason it is duplicated rather than shared is that the
+  // hooks are per-tool fields on `ToolModule`.
+  async extraMeta(output, ctx) {
```
robertabbott, line 366: This adds a second serial upstream round-trip to every `tako_answer` call on a widget client, and the PR body quantifies only the schema bytes. `extraMeta` is awaited at `mcp.ts:1196` after `handler` resolves with nothing running alongside it, so end-to-end latency is now answer + this fetch.

The timeouts are not the interesting number: `image_url` points at `/api/v1/image/{pub}/`, which renders on demand, and the comment at `_chart_widget.ts:1884-1896` says explicitly that a `Range` header saves transfer, not render — so the ChatGPT branch pays the render too, just not the bytes. What is the measured p50/p99 of that render for a card `/api/v1/answer/` just returned, and what does it do to answer's distribution? The upstream answer call already carries `timeoutMs: 130_000` (line 266), so the worst case is ~138 s before anything is emitted — has that been checked against host tool-call budgets?

Backend side: the render rate on that endpoint goes from search QPS to search + answer QPS. Is `/api/v1/image/{pub}/` cached per `pub_id`+`dark_mode`, so an answer whose top card was already charted is a hit?

workers/src/tools/tako_answer.test.ts
```diff
@@ -198,6 +199,81 @@ describe("tako_answer handler", () => {
...
+      CTX,
+    );
+    const o = out as unknown as Record<string, unknown>;
+    expect(o.pub_id).toBe("abc123");
+    expect(String(o.embed_url)).toContain("/embed/abc123/");
+    expect(String(o.image_url)).toContain("/api/v1/image/abc123/");
+    expect(typeof o.height).toBe("number");
+  });
+
+  it("declares the same widget resource search does", () => {
```
robertabbott, line 232: The comment states an invariant — "requestOrigin must reach the resource: it is what the native-card URL and the connectDomains declaration are derived from" — that I do not think this fixture can exercise. `ENV` at line 30 is `{ DJANGO_BASE_URL }` only, so `PUBLIC_CDN_URL` is unset; `resolvePublicCdnBase(env)` is then undefined, which short-circuits `connectDomains` to `[]` (`_chart_widget.ts:2158-2165`) and drops the origin from `resourceDomains` (`:2176-2178`), and `nativeCardUrl` bails at `:212` before reading the origin at all. `html` is the static `WIDGET_HTML` either way.

Under this env, is there any compared field whose value differs between `appUiResource(ENV, ORIGIN)` and `appUiResource(ENV, undefined)` — i.e. would rewriting the implementation to drop the second argument turn this red? Separately, every assertion is relational (`a.x` vs `s.x`), so an identical regression inside the shared `buildChartAppUiResourceFromOutputPubId` passes on both sides. Would adding `PUBLIC_CDN_URL` to this env and pinning `a.uri` to the absolute `APP_UI_RESOURCE_URI` close both?

workers/src/tools/tako_answer.test.ts
```diff
@@ -198,6 +199,81 @@ describe("tako_answer handler", () => {
...
+    expect(a.uri).toBe(s.uri);
+    expect(a.html).toBe(s.html);
+    // requestOrigin must reach the resource: it is what the native-card URL
+    // and the connectDomains declaration are derived from.
+    expect(a.frameDomains).toEqual(s.frameDomains);
+    expect(a.resourceDomains).toEqual(s.resourceDomains);
+    expect(a.connectDomains).toEqual(s.connectDomains);
+  });
+
+  it("carries the widget hooks search carries", () => {
```
robertabbott, line 264: `typeof … === "function"` is the only coverage these hooks get anywhere — grepping `extraMeta|extraContentBlocks` across `workers/src` and `workers/test` returns just these two lines, so search's copies have no behavioural test either.

That means nothing would catch inverting `bakeImage: ctx.client !== "chatgpt"` at line 365, which is the gate the comment justifies as "a 64-byte ranged read instead of a ~170 KB render" — inverted, every ChatGPT answer call pays the full 8 s `PNG_FETCH_TIMEOUT_MS` budget for a payload the host discards. Nor would anything catch `extraContentBlocks` calling `fetchPngContentBlock(undefined)` instead of returning `[]` when `image_url` is absent.

Since the comment at 355 says these are duplicated rather than shared because the hooks are per-tool `ToolModule` fields, is there an assertion that would notice the two copies drifting — `takoAnswer.extraMeta(output, {...CTX, client: "chatgpt"})` with a stubbed fetch asserting no PNG body was requested, and `await takoAnswer.extraContentBlocks({}, CTX)` resolving to `[]`?

workers/src/tools/_search_results.ts
```diff
@@ -1015,22 +1015,40 @@ export function buildSearchOutput(
...
+ * Only the TOP card gets a chart — the widget renders one, and `pub_id` is
+ * singular in the output schema.
+ */
+export function topCardChartFields(
+  // `card_id` is nullable on the wire (`takoCardSchema`), so accept null here
+  // rather than making each caller narrow it — the `typeof` check below is the
+  // one place that decides what counts as renderable.
+  cards: readonly { card_id?: string | null | undefined }[],
+  env: Env,
+): Record<string, unknown> {
```
robertabbott: This helper exists so the two tools cannot drift, but typing it `Record<string, unknown>` gives up the check that would enforce that. Before the extraction these six fields were an inline literal inside `buildSearchOutput`, whose declared return type is `SearchOutput` — a renamed or mistyped key was a compile error. Spreading a `Record<string, unknown>` erases it: `{ ...base, ...topCardChartFields(cards, env) }` still typechecks as `SearchOutput` if this emits `pub_idd`, or a `height` that is a string.

It shows up downstream too. `AnswerFullOutput` (`_render_markdown.ts:117`) absorbs the six into its `[key: string]: unknown`, so answer's hooks have to write `(output as { image_url?: string }).image_url` at lines 362 and 371 where `tako_search.ts:309`/`:318` read `output.image_url` off a typed value. So a rename breaks search's build and silently stops answer rendering — the asymmetry the shared helper was meant to remove.

Since `autoChainShape` is already the single declaration of these fields for both advertised schemas, would tying the return type to it — `Partial<z.infer<z.ZodObject<typeof autoChainShape>>>`, or an explicit `Pick<SearchOutput, …>` — restore the link and let those casts go?

robertabbott (commented):
Reviewed the theme/handshake work, the `tako_answer` widget wiring, and the cache lever. Comments inline.

Cleared, so it is on the record and nobody re-derives it: I brute-forced `safeCssColor`'s emitted regex over U+0000..U+02FF in leading, body and trailing position plus hand-built `</style>`, `url(`, `expression(`, hex-escape and comment payloads — no injection, and the CSSOM path receives the same validated string, so it can do nothing the injected `<style>` cannot. `hostContext` arrives by structured clone, so there is no getter/Proxy TOCTOU between the check and the return. The postMessage gate is unchanged and covers both new branches identically to the pre-existing `tool-result` branch. `extraContentBlocks` is not ungated — `mcp.ts:1153` gates it on `ui === undefined`, the inverse of `extraMeta`'s gate, so ChatGPT and Claude never pay for a discarded PNG block. No committed artifact encodes any tool's `outputSchema`, so `...autoChainShape` needs no regeneration. No tests were deleted, weakened, skipped, or gated on env vars.

The two I would most want resolved before merge are the handshake-ordering one and the added per-call render round-trip.

workers/src/tools/_chart_widget.ts
noahjax, line 745: **[Issue] `mcpHostSurface` survives a theme flip, so the canvas keeps the old theme's colour.**

`mergeHostContext` never clears `mcpHostSurface`. When a host toggles theme and sends the spec-legal partial `{theme: "light"}` with no `styles`, the dark surface from initialize is retained and `applyHostSurface()` repaints it -- under a card the same update just flipped to light. Confirmed on both paths, driving the real bundle under the widget's jsdom harness:

```
widget canvas:  background = "rgb(25, 24, 23)"   colorScheme = "light"
native card:    :root,body{background:#191817}:root{color-scheme:light}
```

The comment above this function cites the spec's merge rule to justify not clearing `theme`. The mirror case needs the opposite: a surface colour is theme-dependent in a way a theme value is not, so retaining it across a flip is the one thing merge semantics must not do here.

The cost is not just a stale colour. Because `mcpHostSurface` stays truthy, tier 1 returns early and tier 2 -- which would be correct -- never runs. The `host-context-changed` handler's own comment says the corners must not stay wrong; with a stale colour they stay wrong and now confidently so.

Suggest: when an update changes `mcpHostTheme` and carries no surface of its own, null `mcpHostSurface` and fall through to tier 2.

workers/src/tools/_chart_widget.ts
```diff
@@ -650,23 +675,147 @@ const WIDGET_HTML = `<!doctype html>
...
+  }
+
+  // This value arrives over postMessage and is interpolated into a \`<style>\`
+  // block injected into the native card document, so it is an injection sink.
+  // Allow-list a colour grammar instead of trying to escape one: hex, and the
+  // functional notations, with no semicolons, braces, quotes, or \`url(\`.
+  // Anything else is dropped and we fall back to \`color-scheme\`.
+  function safeCssColor(v) {
+    if (typeof v !== "string" || v.length === 0 || v.length > 64) return false;
+    if (/^#[0-9a-fA-F]{3,8}$/.test(v)) return true;
```
noahjax: **[Suggestion] The hex branch still accepts values CSS rejects.**

`^#[0-9a-fA-F]{3,8}$` matches 5- and 7-digit hex, which are not colours. Confirmed: `#12345` passes `safeCssColor`, `mcpHostSurface` goes truthy, and `background` lands as `""`.

That is the same shape as the `rgb()` bug `f3aad74` closed -- tier 1 "succeeds" while painting nothing -- and `log("native card upgrading", {surface})` below then reports a surface that never rendered, which is the exact class of misleading log the PR body says cost three rounds of debugging.

Fix is `{3,4}|{6}|{8}`, or add `CSS.supports("background-color", v)` after the allow-list so what validates matches what the parser accepts. The allow-list still has to run first -- it is the injection guard -- but `CSS.supports` is what makes the "assigning it is a no-op" class impossible rather than enumerated.

workers/src/tools/_search_results.ts
noahjax: **[Suggestion] `Record<string, unknown>` drops the field-name check.**

These fields used to be written inline in `buildSearchOutput` and checked against `SearchOutput`. Spread from an index-signature type, a renamed or misspelled key is no longer a compile error.

That matters more here than usual: the stated reason for extracting this function is that the two tools cannot drift. Sharing the code stops one kind of drift and the widened return type reopens another -- a typo in `embed_url` now reaches the host as a silently missing field on both tools at once. A named return type with optional fields keeps the sharing and restores the check.

workers/src/tools/_chart_widget.ts
```diff
@@ -106,6 +106,31 @@ export const PNG_HEAD_FETCH_TIMEOUT_MS = 3_000;
...
+ * SHIPPED bundle ever needs cache-busting, register a new URI ALONGSIDE the old
+ * one, per the note above.
+ */
+export function appUiResourceUri(env: Env): string {
+  const suffix = env.WIDGET_URI_SUFFIX;
+  if (typeof suffix !== "string" || suffix === "") return APP_UI_RESOURCE_URI;
+  // Constrain the value: it becomes a URI the host stores and reads back, and
+  // a stray `/` or space would make a URI that never resolves.
+  const safe = suffix.replace(/[^A-Za-z0-9._-]/g, "");
+  return safe === "" ? APP_UI_RESOURCE_URI : `${APP_UI_RESOURCE_URI}-${safe}`;
```
noahjax, line 132: **[Suggestion] The cache-bust does not cover the per-call URI.**

Only `APP_UI_RESOURCE_URI` is suffixed; `APP_UI_TEMPLATE_URI_PATTERN` is not. The template URI self-busts across different charts because `{pub_id}` varies, but not when you re-test the SAME chart -- which is the loop this lever exists for, and the loop that produced the stale-bundle debugging session in the docstring.

Either suffix the pattern too, or state why it does not need it. Worth resolving in this PR rather than the next dev session that hits it.

noahjax (commented):
Reviewed the design and ran two probes against the branch under the widget's own jsdom harness rather than reasoning from the source. One confirmed correctness bug, three suggestions. Not a merge blocker: the bug needs `styles.variables`, which this PR reports the live claude.ai session never sent, so tier 1 cannot fire there today.

The design work is strong. The two-tier fix with the gate on tier 2 rather than tier 1 is the right shape, the measurement table earns its place, and 579 lines of tests that actually execute the bundle is the reason a finding like #1 is worth stating precisely instead of hand-waving.

One thing missing from the body's cost section. The `tako_answer` widget parity includes `extraContentBlocks`, which `mcp.ts` fires ONLY when `ui === undefined` -- Cursor, Windsurf, Gemini CLI, plain API clients. So every `tako_answer` call from a non-widget host now carries an inline base64 PNG (`MAX_INLINE_PNG_BYTES` = 4 MB cap, typical ~170 KB) plus up to `PNG_FETCH_TIMEOUT_MS` = 8 s of serial fetch, on top of answer's existing 130 s ceiling. That is a defensible call -- an answer's citations are cards -- but it is a larger number than the 353 bytes the body quantifies, and it lands on the tool the description calls START HERE.

One assumption worth writing down while you are in here: tier 1 is called safe anywhere because the canvas becomes the host's page colour. That holds only if the widget iframe sits on the host's PRIMARY background. Inside a message bubble or a tool-result card with its own surface, painting `--color-background-primary` produces exactly the visible square the tier-2 gate exists to prevent, ungated this time. `readHostSurface` tries primary before secondary; for a widget embedded in a chat bubble the reverse may be the better first guess.

workers/src/tools/_chart_widget.ts
25eliu: Fixed, and you were right that the precondition was the real issue rather than the wording.

The first tool-result is now HELD until the init response lands or a 400 ms grace expires (`releaseRenderHold`). One correction to the shape I first proposed: **the 200 ms timer is not the deadline to hold against** — in your scenario it fires at 200 and the tool-result arrives at 210, so holding to *that* deadline changes nothing. The hold needed its own, later window.

Gated on `initializedSent` as well: a tool-result arriving *before* our `initialized` comes from a host not following the handshake, so there is no response to wait for and holding would only delay a chart for no theme. That gate is also why the other 80 tests still pass — they all deliver without a handshake, which is the non-race order.

5 tests: held-then-released-by-response (asserting the native url carries `dark_mode=false`, i.e. the card is host-themed where it previously carried `auto`), released-by-grace, no-double-render when both fire, and not-held when the tool-result precedes `initialized`.

The comment now states the real precondition instead of "spec-compliant".

workers/src/tools/_chart_widget.ts
25eliu: Fixed — snapshotted at the url, as you suggested.

`upgradeToNativeCard` now captures `snapSurface`/`snapScheme` at entry, next to the url whose `dark_mode` the caller already fixed from the same values, and the `.then()` uses only the snapshot. Your read of the consequence was right and worse than a stale colour: `nativeStarted` blocks the second fetch, so the mismatch would have been permanent for that card rather than corrected on the next update.

workers/src/tools/_chart_widget.ts
25eliu: Two things here; I took the logging and left the candidate order.

**Logging: done.** `host context at initialize` now reports `surface`, `styleVarCount`, `hasPrimary` and `hasSecondary` alongside the theme. As you say, the theme half was already known — one dogfooding session now answers whether tier 1 can ever fire and which candidate matched.

**Candidate order: deliberately not changed yet.** You and noahjax both flagged that `--color-background-primary` may not be the colour immediately behind a widget mounted in a bordered container, and that falling back to `-secondary` treats them as interchangeable. I agree it is a real risk, and I would rather not guess a second time in the same PR — reversing the order is as speculative as the current order until the log shows a host actually shipping the variables. The new logging is what makes that decision evidence-based instead of another coin flip.

Noted in the PR body as an open question rather than a settled call.

workers/src/tools/_chart_widget.ts
25eliu: Fixed. A theme CHANGE carrying no replacement surface now nulls `mcpHostSurface` and falls through to tier 2.

Your framing is what made the fix correct: `mcpHostTheme` is self-describing, `mcpHostSurface` is theme-DEPENDENT, so the spec merge rule applies to one and must not to the other.

The test also caught that clearing the variable was only half of it — tier 2 has to actively UNPAINT tier 1, or the element keeps the old inline colour no matter what the tier decides. Three tests now: the mirror, a same-theme partial update that must KEEP the colour, and a flip that carries its own new surface.

workers/src/tools/_chart_widget.ts
25eliu: Fixed, and I took the CSSOM suggestion rather than patching the spelling.

Confirmed your finding and more: `rgb( )`, `rgb(url)`, `lab(image-set)`, `#12345` and `#1234567` all passed. Split into two gates — `allowListedColorShape` stays FIRST and unchanged as the injection guard (hex tightened to the four spellings CSS actually defines), then `paintsAsColor` uses `CSS.supports("background-color", v)` with a CSSOM round-trip fallback.

Your framing of the property being checked was the useful part: enumerating spellings only ever closes the next one, and each leak had the same signature — truthy surface, early return, and a log claiming a colour that never rendered.

workers/src/env.ts
25eliu: Split these: fixed the first, filing the second.

**Template uri: fixed.** `appUiTemplateUriPattern(env)` derives from `appUiResourceUri(env)`, and both the registration and the resolver go through it so they cannot disagree. Your point about it self-busting across different charts but not the same one is exactly the loop the lever exists for.

**Production kill path: filing separately, not solving here.** You are right that a redeploy does not reach a session holding a cached bundle and that this var is documented as must-stay-unset there, so neither is the escape hatch. The blast radius is bounded today because `PUBLIC_CDN_URL` is absent from the production block, but that is a coincidence of the rollout rather than a kill path — so it needs a real answer before the native card is promoted, and it is a bigger design question than this PR. Filing as its own issue with your framing.

workers/src/tools/tako_answer.ts
25eliu: Fixed, including the published artifacts.

`tako_answer`'s DESCRIPTION now ends "(request_id, usage, guidance, chart-widget fields)" and the "for breadth recon **and chart cards**" routing is gone. Regenerated `registry/server.json` and `lhm.plugin.json`. Also patched `llms.txt`, `llms-full.txt`, and both `README.md` spots.

Your last point was the one I had not considered and is the reason this belongs in this PR rather than a follow-up: the description is what stops a model re-posting `image_url` as a markdown image and double-rendering the card. It now says so explicitly.

workers/src/tools/tako_answer.ts
25eliu: Softened the comment rather than sharing the ordering, and documented both divergences you found.

The comment now says the URL CONSTRUCTION is shared — so the two cannot disagree about how a `card_id` becomes a chart — while top-card SELECTION is per-tool, and names both mechanisms: `citesByPosition` keeping the backend order when the prose carries `[N]` markers, and answer ordering RAW cards where search orders SLIMMED ones so that on `include_contents: false` search falls through to freshness while answer still ranks on row dates.

I did not share the ordering step because answer's citation ordering is a deliberate trade (citation numbering has to match the prose) and changing which card gets charted is a behavioural change I should not make quietly inside a rendering PR. If you would rather the chart always came from the usefulness order, that is a fair argument — it just wants its own change with the demoted-Overview case reasoned through.

workers/src/tools/tako_answer.ts
25eliu: Measured it. **0.50–0.87 s**, n=5 against prod, including a card `/api/v1/answer/` had just returned for the first time (0.84 s cold, 0.60 s on repeat).

So the added serial hop is well under a second, not near the 8 s `PNG_FETCH_TIMEOUT_MS` ceiling — the ~138 s worst case is real arithmetic but not the operating point. Numbers are now in the PR body next to the schema bytes; you were right that quantifying only the bytes was the wrong thing to have measured.

Two parts I cannot answer from here and am carrying as open questions rather than claiming:
- **p99.** n=5 gives a p50 and nothing credible past it.
- **Backend caching per `pub_id`+`dark_mode`.** The 0.84 → 0.60 s delta hints at some warmth but does not establish a cache, and I cannot see that endpoint's internals. Worth a backend answer before this lands, since it decides whether the render rate really goes from search QPS to search + answer QPS or mostly hits.

The parity argument still holds for me — an answer's citations are cards — but that is a product call, so flagging rather than assuming.

workers/src/tools/tako_answer.test.ts
25eliu: Correct, that test could not fail. Fixed.

Confirmed your reasoning: with `PUBLIC_CDN_URL` unset, `resolvePublicCdnBase` is undefined, `connectDomains` short-circuits to `[]`, `nativeCardUrl` bails before reading the origin, and `html` is the static bundle either way — so dropping the second argument would have kept it green.

Now: a `WIDGET_ENV` that sets `PUBLIC_CDN_URL`, `a.uri` pinned to the absolute `APP_UI_RESOURCE_URI` rather than only compared to search's (your second point — an identical regression inside the shared builder passed both sides), and a new test asserting `connectDomains` is `["https://mcp.example.test"]` with an origin and `[]` without, so the argument is genuinely observable.

workers/src/tools/tako_answer.test.ts
25eliu: Fixed — both assertions you proposed, near enough.

`extraMeta` with `client: "chatgpt"` now stubs fetch and asserts exactly one request carrying a `Range` header, so inverting `bakeImage` turns it red (the inverted branch issues an unranged GET for the full render). And `extraContentBlocks({}, CTX)` resolves to `[]` with fetch never called, so `fetchPngContentBlock(undefined)` cannot creep back.

You are right that search's copies still have no behavioural test of their own. I did not add them there to keep this PR's diff to the tool it changes, but the gap is real and worth closing — the two copies exist precisely because the hooks are per-tool `ToolModule` fields, so drift is the expected failure.

workers/src/tools/_search_results.ts
25eliu: Fixed — typed from `autoChainShape`, as you suggested.

`topCardChartFields` now returns `TopCardChartFields | undefined`, mapped from `z.infer<z.ZodObject<typeof autoChainShape>>`. Whole-or-undefined rather than `Partial<>` because of the spread site: under `exactOptionalPropertyTypes` spreading an optional widens it to `| undefined`, which `SearchOutput` rejects — so `buildSearchOutput` branches instead of spreading `?? {}`, and the mapped type uses `-?` plus `NonNullable` to drop both the optionality and the `| undefined` in the value.

The downstream half of your finding resolved itself: both casts in `tako_answer`'s hooks are gone, so it reads `output.image_url` off a typed value exactly like search does. That asymmetry — a rename breaking search's build while silently stopping answer rendering — was the sharpest part of the comment.

workers/src/tools/_chart_widget.ts
25eliu: Confirmed and fixed. Thank you for running it rather than describing it — the harness output made this unambiguous.

Took your suggested fix: a theme CHANGE carrying no surface of its own now nulls `mcpHostSurface` and falls through to tier 2. Your framing of why the spec merge rule applies to `theme` but must not apply to a surface colour is what made it obvious.

Writing the test surfaced a second half neither of us named: clearing the variable is not enough, because `applyHostSurface` had already set the colour as an inline style. Tier 2 now actively clears `background` on `<html>`/`<body>`, otherwise the canvas keeps painting the old theme's surface no matter what the tier decides. Your repro is now a test, plus two complements — a same-theme partial update that must KEEP the colour, and a flip that carries its own new surface.

workers/src/tools/_chart_widget.ts
25eliu: Confirmed `#12345` and `#1234567`, and took the stronger of your two suggestions.

Both: hex tightened to `{3,4}|{6}|{8}`, AND `CSS.supports("background-color", v)` after the allow-list, with a CSSOM round-trip fallback for hosts without it. As you say, the allow-list has to run first because it is the injection guard, but `CSS.supports` is what makes the whole no-op class impossible instead of enumerated — which mattered, because `rgb( )`, `rgb(url)` and `lab(image-set)` were also getting through and I would only have closed the two spellings named.

Your point about the misleading log landed hardest: `log(..., {surface})` reporting a colour that never rendered is the exact failure mode that cost three debugging rounds on this branch.

workers/src/tools/_search_results.ts
25eliu: Fixed — a named return type, keeping the sharing.

`TopCardChartFields`, mapped from the same `autoChainShape` both advertised schemas are built from. Whole-or-undefined rather than optional fields, because under `exactOptionalPropertyTypes` spreading an optional widens it to `| undefined` and `SearchOutput` rejects that.

Your closing line is the reason this was worth doing rather than annotating: sharing the code stopped one kind of drift and the widened return type reopened another, on both tools at once. The two casts in answer's hooks are gone as a side effect.

workers/src/tools/_chart_widget.ts
25eliu: Fixed — the pattern is suffixed too.

`appUiTemplateUriPattern(env)` derives from `appUiResourceUri(env)`, and both the `ResourceTemplate` registration and the per-call resolver go through it, so the registered string and the advertised one cannot diverge.

Your reasoning is why I did not just document it: the template self-busts across different charts but not when you re-test the same one, which is exactly the loop that produced the stale-bundle session in the docstring. Leaving that for the next dev session to rediscover would have been the same mistake twice.

workers/src/tools/tako_answer.ts
25eliu: Answered both parts from the code and a real sample — I should not have punted this to "needs a backend answer" when the backend is checked out locally.

**Yes, cached per `pub_id` + `dark_mode`.** The MCP worker sends no `width`/`height`/`hideFooter`, so `ImageView` takes the managed-cache path rather than `_generate_oneoff_image`: `ImageGenerationManager(config, variant=REGULAR if dark_mode else LIGHT_MODE)`, then `request_and_wait` → `request_image` → `_check_existing_image`, which returns `READY` with the stored `image_url` when status is `COMPLETED` and the hash matches. The key is `pub_id:dark_mode:for_powerpoint:width:height:hide_footer:glanceable` (`ImageRequest.request_hash`; timezone/locale are deliberately excluded so a prewarmed image is not re-rendered on a zone mismatch).

Search and answer both use `DEFAULT_DARK_MODE = true` and pass no dimensions, so **their request hashes are identical** — an answer whose top card search already charted is a hit.

**Concurrent calls join rather than double-render.** `_check_existing_image` also returns `should_poll: true` on `PENDING`/`PROCESSING` with a matching hash, so a search and an answer racing on the same card wait on one render.

So the render rate does **not** go from search QPS to search + answer QPS. It grows only by cards that answer cites and search never charted at the same params.

**Distribution, n=21 across 7 distinct cards, 3 passes each, against prod:**

```
min  0.186s
p50  0.277s
p90  0.374s
max  0.464s
```

Those are warm hits — the cards had just been charted by the `tako_search` calls that produced the pub_ids, which is itself the cache doing its job. A genuinely cold render measured ~0.84s in the earlier n=5 (single sample, so treat it as an order of magnitude, not a p99). Either way the operating point is a fraction of a second, not near the 8s `PNG_FETCH_TIMEOUT_MS` ceiling.

Updating the PR body to state this as resolved rather than open.

workers/src/tools/_chart_widget.ts
```diff
@@ -1560,21 +1865,91 @@ const WIDGET_HTML = `<!doctype html>
...
+      //
+      // \`initializedSent\` is part of the gate on purpose. The spec says a host
+      // MUST NOT send notifications before our \`initialized\`, so a tool-result
+      // that arrives first comes from a host not following the handshake —
+      // holding it to wait for a response that may never come would only delay
+      // a chart for no theme. The race being closed is the other order: we sent
+      // \`initialized\` on the 200 ms timer, the host replied to THAT before
+      // replying to \`ui/initialize\`.
+      if (initializedSent && !initResponseSeen && !renderHoldExpired) {
+        heldToolResult = params;
```
noahjax, line 1942: **[Issue] The hold inverts first-wins to last-wins inside its window.**

`heldToolResult = params` overwrites unconditionally, so a second tool-result arriving inside the 400 ms grace replaces the first. Before the hold, `render()`'s `rendered` latch made the FIRST tool-result win and every later one a no-op. Inside the hold window that is now reversed.

Confirmed by differential probe -- same two messages, only the hold differs:

```
fast init (no hold):  chart result, then chartless result  ->  src = data:image/png;base64,AAAA   (chart survives)
held (>200 ms):       chart result, then chartless result  ->  src = (none)                       (chart dropped)
```

The chartless payload takes `render()`'s no-chart short-circuit and collapses the widget, so a host that follows a chart notification with a chartless one within 400 ms loses the chart entirely. Narrow trigger, but the failure is total rather than cosmetic, and the hold is what introduces it.

One line: `if (heldToolResult === null) heldToolResult = params;` -- keeps the first, matching the latch semantics everywhere else in this file.

(The timer is already correct on this point: `if (renderHoldTimer === null)` means a second held result does not extend the grace.)

workers/src/tools/_chart_widget.ts
```diff
@@ -1560,21 +1865,91 @@ const WIDGET_HTML = `<!doctype html>
...
+    if (
+      fromHost &&
+      msg.jsonrpc === "2.0" &&
+      msg.method === "ui/notifications/host-context-changed"
+    ) {
+      mergeHostContext((msg.params || {}).hostContext);
+      // Re-apply: a theme toggle mid-conversation cannot re-theme the chart
+      // itself, but the corners must not stay wrong.
+      applyHostSurface();
+      log("host context changed", { theme: mcpHostTheme });
```
noahjax, line 1925: **[Nit] A `host-context-changed` during the hold does not release it.**

This branch calls `mergeHostContext` and `applyHostSurface` but not `releaseRenderHold`. So a host that answers `ui/initialize` late while sending the theme via a notification first has the theme known and still makes the held tool-result wait out the full 400 ms.

Probed: the output is correct (`dark_mode=false`), just later than it needs to be. Adding `releaseRenderHold("host context changed")` after `applyHostSurface()` costs nothing -- it early-returns when nothing is held, which is every case today.

Worth it mainly because the whole point of `RENDER_HOLD_MS` is that the theme is worth a short wait; once the theme is in hand, the wait is pure latency.

noahjax (approved):
Approving, conditional on the two comments below landing first -- the first one is a one-line change.

Re-ran my probes against `9b33b6b` under the jsdom harness. All three findings from my first pass are fixed, and I verified each rather than reading the diff:

| finding | probe result |
|---|---|
| stale surface across a theme flip | `background=""`, `colorScheme="light"` -- was `rgb(25, 24, 23)` / `light` |
| `safeCssColor` accepting non-colours | `#12345`, `#1234567`, `rgb( )`, `rgb(url)`, `lab(image-set)` all fall to tier 2; `#191817`, `#1918`, `rgb(...)`, `oklch(...)` still paint |
| `Record<string, unknown>` | `npm run typecheck` clean across all four projects, and answer's hooks read `output.image_url` off a typed value |

The unpaint half of the surface fix is the part I missed. Clearing `mcpHostSurface` was only half -- the element keeps the old colour unless tier 2 also resets `background`, and your test caught that where my finding did not.

The render-hold works. I confirmed the race it targets is actually closed (`initialized` at 200 ms, tool-result held, late init response at 350 ms, url comes out `dark_mode=false`), that a host answering inside 200 ms is never held, and that the 400 ms grace expires to `dark_mode=auto` rather than hanging.

Also verified: full suite green (85 widget, 46 scripts, main suite), `registry:check` and `schemas:check` clean, and the template uri now derives from `appUiResourceUri(env)` so the dev lever covers both bundles.

The `styleVarCount`/`hasPrimary`/`hasSecondary` log is the right call. It turns the one open question this PR cannot answer from a design argument into something the next dogfooding session settles.

PR #209, opened 2026-08-04
fix(widget): production never had the switch the theme fix needs

25eliu:
## The report

On prod, the chart in a `tako_answer` result renders as the real Tako card but is **not interactive** (no hover), and is **dark even when Claude is light-themed**. Both were believed fixed and tested in #208.

## Why the fix appeared not to work

It did work — on staging, which is where it was tested. #208 gated the whole native-card path behind `PUBLIC_CDN_URL`, and that var was set in the **staging block only**, with a comment saying it was deliberately absent from production.

So prod ran the new code with the feature switched off. Confirmed rather than assumed:

- `5b71fa9` (#208) is an ancestor of `origin/main`, `workers-deploy` on main succeeded, and `_chart_widget.ts` on `origin/main` is byte-identical to the branch. The code is live.
- `/cdn-asset/archive/x.js` → **502 on staging** (route exists, bogus path) vs **404 on prod** (route absent). That route only exists when `PUBLIC_CDN_URL` is set.

## One binding, both symptoms

`resolvePublicCdnBase` → undefined ⇒ `nativeCardUrl` → undefined ⇒ `_meta.native_card_url` never emitted ⇒ `upgradeToNativeCard` never runs. Prod stays on the baked PNG, and a PNG is:

1. **Not interactive** — a raster inside a click-through anchor. Exactly the "it's the real card but I can't hover it" report.
2. **Permanently dark** — `image_url` is built server-side at `dark_mode=DEFAULT_DARK_MODE` (`true`) before the host theme is knowable, and a raster can't be re-themed client-side. `withHostTheme` only rewrites the native-card and iframe urls (`_chart_widget.ts:1401`, `:1435`, `:1540`), so on prod the theme fix had nothing to act on.

Interactivity and theming were never two bugs. They were one switch, and it was off.

## The change

`PUBLIC_CDN_URL: "https://d12w4pyrrczi5e.cloudfront.net"` in the production block, plus corrections to four comments that now assert something false ("production is byte-identical", "INVISIBLE IN PRODUCTION", "the experiment off (production default)", "the production default").

**Distribution verified, because a mismatch fails silently** (zero urls rewritten → card mounts → no chart draws): prod's live embed page references `d12w4pyrrczi5e.cloudfront.net` 16 times, and never staging's `d1iyjvzoctsna`.

## Security posture

Unchanged from the staging review in `embed_proxy.ts`, but it is now a **production** surface, so restating it: `/embed-html/{pub_id}` and `/cdn-asset/{path}` become public and unauthenticated on the OAuth origin.

- Both metered through the free-tier per-IP limiter.
- `/embed-html/` answers `text/plain` under `Content-Security-Policy: sandbox`; `/cdn-asset/` is confined to the `archive/` prefix and reflects only inert content types. Neither serves a document the browser will execute.
- Proxied page has `csrfToken` and the GA bootstrap stripped.
- Handlers run ahead of the OAuth subsystem, so a malformed value declines rather than throwing (a throw would 500 `/authorize` and `/token`) — covered by tests.

## Test plan

- [x] `npm run typecheck` — all 4 tsconfig projects clean
- [x] `npm test` — 1,068 passed (937 + 85 + 46), 0 failures
- [x] `wrangler deploy --env production --dry-run` confirms the binding resolves: `env.PUBLIC_CDN_URL ("https://d12w4pyrrczi5e.cloudfront.net")`
- [x] `registry:check` / `schemas:check` clean
- [ ] **After deploy:** `/cdn-asset/archive/x.js` on prod returns 502 not 404 (route exists)
- [ ] **After deploy:** a `tako_search` / `tako_answer` chart on claude.ai web is hoverable, and renders light on a light-themed Claude

### DIFF
--- a/workers/scripts/smoke.ts
+++ b/workers/scripts/smoke.ts
@@ -390,20 +390,25 @@ try {
   // Reuses the card `tako_visualize` just created, so there is no hardcoded
   // pub_id fixture to rot.
   //
-  // 404 is NOT a failure, and that distinction is the point. The route is
-  // gated on `PUBLIC_CDN_URL`, so a 404 can mean the native-card path is
-  // switched off in this environment — a legitimate configuration (local dev,
-  // or a deliberate rollback), not a regression. Failing on it would make the
-  // smoke cry wolf about a choice. What DOES fail is 200-but-broken: the route
-  // is live and answering, yet the rewrite it exists to perform did not happen.
+  // A 404 FAILS on a deployed env, and only warns off one.
   //
+  // It was a warning everywhere in the first draft, which got this exactly
+  // backwards: a 404 on `mcp.tako.com` is precisely the outcome that means the
+  // binding never took effect — the whole subject of this check — and warnings
+  // exit 0, so it would have been reported by a green job whose log nobody
+  // opens. Both deployed envs now set `PUBLIC_CDN_URL`, so there is no
+  // legitimate 404 there to cry wolf about. Off a deployed env (a local
+  // `wrangler dev` with no binding) it stays a warning, because there the
+  // absence is the expected configuration rather than a regression.
+  const isDeployedTarget = /^https:\/\/mcp(\.staging)?\.tako\.com$/.test(baseUrl);
   // The two 404s are told apart by body, because conflating them would report
   // "feature off" for a live route: a declined route falls through to the
   // router's generic `not found`, while a live route whose pub_id did not
   // resolve upstream answers `chart not found` (`embed_proxy.ts`, the
-  // `response.status === 404` branch). Both stay warnings — the second is most
-  // likely lag between creating the card and its embed page existing, which is
-  // not something to redden a deploy over.
+  // `response.status === 404` branch). The latter stays a warning even on a
+  // deployed env — it means the route IS serving, and the likeliest cause is lag
+  // between creating a card and its embed page existing, which is not something
+  // to redden a deploy over.
   const nativeRes = await fetch(
     `${baseUrl}/embed-html/${encodeURIComponent(tvStructured.card_id)}`,
   );
@@ -416,11 +421,23 @@ try {
           `between creating it and its embed page existing).`,
       );
     } else {
-      console.warn(
-        `[warn] /embed-html/ → 404 "${nativeBody.trim()}": the native-card ` +
-          `path is OFF here (PUBLIC_CDN_URL unset). Charts on claude.ai ` +
-          `render as a static dark PNG, not an interactive host-themed card.`,
-      );
+      // Deliberately does NOT name a cause. An ABSENT `PUBLIC_CDN_URL` and a
+      // MALFORMED one produce byte-identical responses: `resolveProxyOrigins`
+      // catches the throw from any resolver and declines, so the router answers
+      // the same generic body either way. Since promoting this feature is
+      // exactly "type a new URL into the production block", "set but broken" —
+      // a single trailing slash will do it — is the likelier of the two, and
+      // reporting it as a deliberate configuration choice would send the reader
+      // the wrong way. The worker logs the real cause; the smoke cannot read
+      // worker logs, so it points there instead of guessing.
+      const why =
+        `/embed-html/ → 404 "${nativeBody.trim()}": the native-card path is ` +
+        `not serving here. PUBLIC_CDN_URL is either unset or malformed (both ` +
+        `decline identically) — grep the worker log for "native-card proxy ` +
+        `disabled" to tell which. Charts on claude.ai render as a static dark ` +
+        `PNG, not an interactive host-themed card.`;
+      if (isDeployedTarget) fail(why);
+      console.warn(`[warn] ${why}`);
     }
   } else {
     assert(
@@ -438,14 +455,24 @@ try {
     // The rewrite is what makes the card renderable: Card.js is a
     // `type="module"` script, always a CORS fetch, and the CDN reflects CORS
     // for tako.com alone — so every asset url must be pointed back at
-    // `/cdn-asset/`. Zero occurrences means the distribution in
+    // `/cdn-asset/` on THIS origin. Zero occurrences means the distribution in
     // `PUBLIC_CDN_URL` does not match the one this page references.
+    //
+    // Origin-qualified, not a bare `/cdn-asset/` substring. `rewriteCdnUrls`
+    // writes `${widgetOrigin}${CDN_ASSET_PREFIX}` and `widgetOrigin` prefers the
+    // `PUBLIC_MCP_URL` binding over the request origin — so a prod deploy
+    // carrying a staging value would emit `https://mcp.staging.tako.com/cdn-asset/…`,
+    // satisfy a bare substring check, and serve every asset in the widget from
+    // the wrong worker. That is the same copy-paste class as the bug this PR
+    // fixes, so the check names the origin it expects.
+    const expectedPrefix = `${baseUrl}/cdn-asset/`;
     assert(
-      nativeBody.includes("/cdn-asset/"),
-      `/embed-html/ returned 200 but rewrote no CDN urls — PUBLIC_CDN_URL ` +
-        `does not match the distribution this env's embed page references ` +
-        `(prod: d12w4pyrrczi5e, staging: d1iyjvzoctsna). The card will mount ` +
-        `and no chart will draw. See the worker log for the 0-rewrite warning.`,
+      nativeBody.includes(expectedPrefix),
+      `/embed-html/ returned 200 but rewrote no CDN urls to ${expectedPrefix} — ` +
+        `either PUBLIC_CDN_URL does not match the distribution this env's embed ` +
+        `page references (prod: d12w4pyrrczi5e, staging: d1iyjvzoctsna), or ` +
+        `PUBLIC_MCP_URL names a different worker. The card will mount and no ` +
+        `chart will draw. See the worker log for the 0-rewrite warning.`,
     );
     // Belt-and-braces, stated honestly: the handler already refuses to serve a
     // body whose `csrfToken` it failed to strip (`embed_proxy.ts`, the
@@ -458,6 +485,48 @@ try {
       `/embed-html/ leaked a csrfToken into a body bound for a third-party sandbox`,
     );
     ok(`/embed-html/ → 200 text/plain, CDN urls rewritten, no csrfToken`);
+
+    // Then actually FETCH one, because everything above is still only a claim
+    // about a string. `embed_proxy.ts` calls these two routes "one feature" —
+    // the page is useless without its assets — and until something retrieves an
+    // asset, a syntactically valid, page-matching `PUBLIC_CDN_URL` behind a
+    // distribution that does not serve `archive/` passes every check here and
+    // still draws no chart.
+    //
+    // It also converts the count-flattened-to-a-boolean weakness above into
+    // something with teeth. `ASSET_PATH_PREFIX` confinement is MEASURED, not
+    // structural, so a frontend build that moves the entry bundle out of
+    // `archive/` while leaving the fonts in would keep `rewrites > 0` and a
+    // green check. Fetching the JS specifically — Card.js is the one asset whose
+    // absence means no chart at all — is what notices.
+    const assetUrls = [
+      ...nativeBody.matchAll(
+        new RegExp(`${expectedPrefix.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}[^"'\\s)]+`, "g"),
+      ),
+    ].map((m) => m[0]);
+    assert(assetUrls.length > 0, `no rewritten asset urls to fetch`);
+    // Prefer the JS entry bundle; fall back to the first asset when the naming
+    // changes, so this degrades to "fetched something" rather than to a skip.
+    const probeAsset = assetUrls.find((u) => u.endsWith(".js")) ?? assetUrls[0]!;
+    const assetRes = await fetch(probeAsset);
+    const assetPath = probeAsset.slice(baseUrl.length);
+    assert(
+      assetRes.status === 200,
+      `rewritten asset ${assetPath} → ${assetRes.status}. The page rewrote its ` +
+        `CDN urls, but this origin cannot actually serve them: PUBLIC_CDN_URL ` +
+        `names a distribution that does not serve the "archive/" tree, or the ` +
+        `asset moved out of it. The card mounts and no chart draws.`,
+    );
+    // `ACAO: *` is not cosmetic here: a module script is always a CORS fetch, so
+    // without this header the widget's Card.js load fails even though the bytes
+    // arrived. This is the exact header the proxy exists to add.
+    assert(
+      assetRes.headers.get("access-control-allow-origin") === "*",
+      `rewritten asset ${assetPath} lacks "access-control-allow-origin: *" — a ` +
+        `type="module" script is always a CORS fetch, so the widget cannot load ` +
+        `it. Adding that header is this route's entire reason to exist.`,
+    );
+    ok(`/cdn-asset/ → 200 with ACAO:* for ${assetPath.slice(0, 72)}`);
   }
 } finally {
   await client.close().catch(() => {
--- a/workers/scripts/smoke.ts
+++ b/workers/scripts/smoke.ts
@@ -20,6 +20,13 @@
  *        d. `get_credit_balance`          — `details.credit_balance`
  *                                           must be a number or numeric string (read-only)
  *        e. `tako_visualize`              — creates a card (charges 1 credit)
+ *   6. Native-card proxy (`/embed-html/`) on the card step 5e just created —
+ *      404 soft-warns (path gated off via `PUBLIC_CDN_URL`), 200 hard-asserts
+ *      inert content type, a nonzero CDN rewrite, and no leaked `csrfToken`
+ *
+ * The step-6 rewrite assertion is the one check here that exists to catch a
+ * failure with NO error signal: a `PUBLIC_CDN_URL` naming the wrong CloudFront
+ * distribution answers 200, mounts the card, and draws no chart.
  *
  * Excluded by design:
  *   - `tako_agent` / `tako_agent_start` / `tako_agent_wait` — long-running;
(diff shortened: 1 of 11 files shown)


workers/wrangler.jsonc
```diff
@@ -183,6 +177,31 @@
...
+        // theme is knowable, and a raster cannot be re-themed in the widget — so
+        // a light-themed Claude got a dark, unhoverable image of a chart
+        // (TAKO-3781's user-visible half). `withHostTheme` only ever rewrites the
+        // native-card and iframe urls, so the theme fix has nothing to act on
+        // until this var exists. Interactivity and correct theming are the same
+        // switch.
+        //
+        // What it exposes, reviewed before promoting: `/embed-html/{pub_id}` and
+        // `/cdn-asset/{path}` become public and unauthenticated on the OAuth
+        // origin. Mitigations already in `embed_proxy.ts` — both metered through
```
robertabbott, line 225: This block cites "both metered through the free-tier per-IP limiter" first among the mitigations. `embed_proxy.ts:240` keys that meter with `freeTierRateLimitKey(CF-Connecting-IP)` against `FREE_TIER_RATE_LIMITER` — the same function, same binding and same key that `freetier.ts:313` uses for the anonymous `tools/call` quota, and production configures it at `{limit: 10, period: 60}` (line 221-224).

Two consequences follow from that sharing. First, `embed_proxy.ts:117-119` measures a real prod embed page at 15 CDN references under `archive/`, plus the chunks named at `:718-719`, so one cold render is >=16 metered requests against a bucket of 10 — and `rateLimited` runs before the `cf: {cacheEverything}` upstream fetch, so edge caching does not reduce the metered count. Second, those asset fetches now drain the same bucket that gates a free-tier user's tool calls from that IP, so rendering one card can spend their tool quota.

`README.md:300-307` already prescribes the instrument: "confirm against the DEPLOYED Worker with a burst of more than `limit` metered calls inside one period." Has that burst been run against a deployed env, or is the ~115-per-colo leak the README documents what has been carrying staging?

robertabbott (commented):
Reviewed the promotion rather than the diagnosis — the diagnosis is solid, and the one-line binding change is the right fix. Comments are on the safety story around it: the mitigation this PR leans on hardest (the per-IP limiter), one unauthenticated amplification vector that the limiter does not contain, and the fact that the new post-deploy check soft-warns on the exact outcome that would mean the binding never took effect.

Nothing here contradicts the root-cause analysis. Happy to re-review.

workers/wrangler.jsonc
```diff
@@ -183,6 +177,31 @@
...
+        // the free-tier per-IP limiter, `/embed-html/` answers `text/plain` under
+        // `Content-Security-Policy: sandbox`, `/cdn-asset/` is confined to the
+        // `archive/` prefix and reflects only inert content types, and the
+        // proxied page has its `csrfToken` and GA bootstrap stripped. Neither
+        // route serves a document the browser will execute.
+        "PUBLIC_CDN_URL": "https://d12w4pyrrczi5e.cloudfront.net",
```
robertabbott, line 231: `embed_proxy.ts:665` builds the upstream as `` `${cdnBase}/${assetPath}${url.search}` ``, so an attacker-controlled query string reaches CloudFront and lands in the cache key for the `cf: {cacheEverything: true, cacheTtl: 3600}` subrequest — defeating the mitigation `:155-160` says the route depends on.

`/cdn-asset/archive/<hash>/vite_dist/assets/Card.js?<nonce>` (the hash is public — `icons.ts:29` documents how) is ~1.48 MB and under `MAX_JS_REWRITE_BYTES`, so every nonced request pays a full origin fetch, an `arrayBuffer()`, a `TextDecoder` decode, and a whole-body `split`/`join` copy, then 1.48 MB egress with `ACAO: *` — from a ~200-byte request. Measured on staging: 0.11s cached vs 0.32s / 0.44s per fresh nonce, so the query does reach upstream and does defeat the cache.

The only meter is the bucket in the comment above, which fails open (`:244-249`) and counts per colo. These handlers run at `index.ts:55-64`, ahead of the whole OAuth subsystem, so CPU and memory pressure here degrades `/authorize`, `/token` and `/register` on the same worker.

Content-hashed assets should not need a query string, and `rewriteCdnUrls` never emits one. Is dropping `${url.search}` safe here, or is there an asset that relies on it?

workers/scripts/smoke.ts
```diff
@@ -367,6 +374,91 @@ try {
...
+  // router's generic `not found`, while a live route whose pub_id did not
+  // resolve upstream answers `chart not found` (`embed_proxy.ts`, the
+  // `response.status === 404` branch). Both stay warnings — the second is most
+  // likely lag between creating the card and its embed page existing, which is
+  // not something to redden a deploy over.
+  const nativeRes = await fetch(
+    `${baseUrl}/embed-html/${encodeURIComponent(tvStructured.card_id)}`,
+  );
+  const nativeBody = await nativeRes.text();
+  if (nativeRes.status === 404) {
```
robertabbott, line 416: Both 404 branches are `console.warn` only, and the script signals failure exclusively through `fail()` -> `process.exit(1)` (`:87-90`), so a 404 exits green with no annotation and no run-summary entry — a human sees it only by opening the log of a passing job.

The PR description treats that same status as the failure signal: "After deploy: `/cdn-asset/archive/x.js` on prod returns 502 not 404 (route exists)". So the one outcome that means this PR did not take effect is the one outcome the new check downgrades to a warning.

Compounding it, production deploys fire from `release: published` (`workers-deploy.yml:18-19`), while `workers-smoke.yml:28-31` admits only `workflow_dispatch` or `workflow_run.event == 'push'` — a `release`-triggered deploy matches neither, so a prod deploy runs no smoke at all and step 6 reaches production only if someone dispatches it by hand.

Should the 404 branch `fail()` when `SMOKE_BASE_URL` names a deployed env — both of which now set the binding — and keep the soft warning for local only?

workers/scripts/smoke.ts
```diff
@@ -367,6 +374,91 @@ try {
...
+    if (nativeBody.includes("chart not found")) {
+      console.warn(
+        `[warn] /embed-html/ → 404 "chart not found": the route is LIVE, but ` +
+          `card ${tvStructured.card_id} did not resolve upstream (likely lag ` +
+          `between creating it and its embed page existing).`,
+      );
+    } else {
+      console.warn(
```
robertabbott: This message asserts "PUBLIC_CDN_URL unset", but a **malformed** value produces a byte-identical response: `resolveProxyOrigins` (`embed_proxy.ts:197-220`) catches a throw from any of the three resolvers and returns `undefined`, so the handler declines and the router answers the same `"not found"` body it answers for an absent binding. A single trailing slash is enough — pinned as a BAD case at `embed_proxy.test.ts:651`, and the same holds for a malformed `PUBLIC_BASE_URL` (`:670-677`).

Given this PR's entire content is typing a new URL into the production block, "set but broken" is the live risk here, and it is the one this line reports as a deliberate configuration choice. The worker does log the real cause (`embed_proxy.ts:214`), but the smoke does not read worker logs.

Would keying the message off the body alone — or dropping the causal claim and pointing at the `native-card proxy disabled` log line — be more honest than naming a cause the check cannot distinguish?

workers/scripts/smoke.ts
```diff
@@ -367,6 +374,91 @@ try {
...
+      nativeType.includes("text/plain"),
+      `/embed-html/ must answer text/plain on the OAuth origin, got ${JSON.stringify(nativeType)}`,
+    );
+    // The rewrite is what makes the card renderable: Card.js is a
+    // `type="module"` script, always a CORS fetch, and the CDN reflects CORS
+    // for tako.com alone — so every asset url must be pointed back at
+    // `/cdn-asset/`. Zero occurrences means the distribution in
+    // `PUBLIC_CDN_URL` does not match the one this page references.
+    assert(
+      nativeBody.includes("/cdn-asset/"),
```
robertabbott: Three gaps in what this substring proves.

It never pins the origin. `rewriteCdnUrls` writes `${widgetOrigin}${CDN_ASSET_PREFIX}` (`embed_proxy.ts:494-498`), and `widgetOrigin` prefers `PUBLIC_MCP_URL` (`env.ts:341-360`). A prod deploy carrying a staging value emits `https://mcp.staging.tako.com/cdn-asset/...`, the substring is present, this passes, and every asset in the widget loads from the wrong worker — the same copy-paste class of bug this PR exists to fix.

It flattens a count to a boolean. The handler's own signal is the count (`:500-508`), and the rewrite is scoped to `${cdnBase}/archive/` where `ASSET_PATH_PREFIX` is measured rather than structural (`:117` — "Measured against a real production embed page"). A frontend build that moves the entry bundle out of `archive/` while leaving the fonts in yields `rewrites > 0`, a green check, and a `Card.js` that still dies on the CDN's CORS allow-list.

And nothing ever fetches a `/cdn-asset/` URL, so a syntactically valid, page-matching `PUBLIC_CDN_URL` behind a distribution that does not actually serve `archive/` still passes — even though `embed_proxy.ts:193-195` calls the two routes "one feature."

Would asserting `` nativeBody.includes(`${baseUrl}/cdn-asset/`) `` and then fetching the first rewritten URL close all three, and automate the second unchecked box in the PR description at the same time?

workers/src/tools/_chart_widget.ts
```diff
@@ -256,7 +256,9 @@ export async function buildChartExtraMeta(
   },
 ): Promise<Record<string, unknown> | undefined> {
   // Native-card field rides along on whatever else this returns. Empty unless
-  // `PUBLIC_CDN_URL` is set, so production is byte-identical.
+  // `PUBLIC_CDN_URL` is set — which both deployed envs now do, so this field is
```
robertabbott, line 259: This PR corrects four comments on the grounds that leaving them "would mislead the next reader," which is the right call. A sweep of the branch finds five more it did not reach, in three files outside the diff:

- `env.ts:52` — "EXPERIMENT SWITCH, and unset everywhere by default"
- `env.ts:82` — "Leave unset in production. Set it on staging to run the experiment."
- `index.ts:54` — "route does not exist at all in production"
- `index.test.ts:63` — "'invisible in production' is property #1 of the whole feature"
- `test/widget/widget-dom.test.ts:984` — "production default"

`env.ts:82` is the load-bearing one: it is the binding's own declaration-site docstring, the first place a reader looks up what `PUBLIC_CDN_URL` is for, and it now reads as an instruction to revert this PR. Worth folding all five in while the reasoning is fresh?

workers/src/embed_proxy.test.ts
```diff
@@ -5,9 +5,11 @@
  *
  * Three properties carry real weight here, in this order:
  *
- *   1. INVISIBLE IN PRODUCTION. Gated on `PUBLIC_CDN_URL`; unset means the
- *      handler declines and the router 404s, so no new surface exists. A
- *      MALFORMED value declines too — these handlers run ahead of the whole
+ *   1. ABSENT WITHOUT ITS BINDING. Gated on `PUBLIC_CDN_URL`; unset means the
+ *      handler declines and the router 404s, so no new surface exists. Both
```
robertabbott, line 9: This comment now says the properties below are "load-bearing rather than theoretical" — and the two most load-bearing ones are the two without coverage in this file.

The rewrite inside `handleEmbedProxy` (`embed_proxy.ts:489-509`) is never asserted through the handler. `rewriteCdnUrls` has five direct tests (`:350-398`) and the JS path is covered through `handleCdnAssetProxy` (`:483-508`), but the 200-body test at `:202-230` never checks the body for `/cdn-asset/`, even though the fixture already references the CDN under `/archive/` (`UPSTREAM_HTML:48`). That wiring is exactly what smoke step 6 exists to validate, so today it is verified only post-deploy — by the check in the thread on `smoke.ts:411`.

`FREE_TIER_RATE_LIMITER` appears nowhere in this file, so the deny path (`:250`), the fail-open catch (`:244-249`), and the call *ordering* are untested. Ordering is the one with teeth: the limiter runs before the upstream fetch on both routes (`:420-421`, `:662`), and a refactor that moves either below its fetch makes the meter cosmetic with nothing going red — for the mitigation this PR cites first.

Should both get unit coverage here now that these routes are live in both deployed envs?

workers/src/embed_proxy.ts
```diff
@@ -43,9 +43,11 @@
  *     what claude.ai reports to the user as "There was a problem displaying
  *     content from tako." Removing it removes a guaranteed false alarm.
  *
- * Gated on `PUBLIC_CDN_URL`: with the experiment off (production default) this
- * route 404s exactly like any unknown path, so the surface is unchanged. A
- * MALFORMED value gates it off too rather than throwing — see
+ * Gated on `PUBLIC_CDN_URL`, which is now set in BOTH staging and production —
+ * so on the deployed envs these routes are LIVE, and this file is the public
```
robertabbott, line 51: Not exploitable today, flagged because the docstring claims more than the code delivers.

`:630-646` says `url.pathname` normalization plus the checks below confine the path, and that those checks "still hold if this is ever called with a hand-built path instead of a parsed Request." `/cdn-asset/archive/x/%2e%2e%2fuser-images/evil.html` defeats both: `%2f` is not a segment delimiter, so WHATWG parsing leaves `%2e%2e%2f` intact as a single segment, and the resulting `assetPath` starts with `archive/`, contains no literal `..`, no leading `/` and no `\`. Verified on staging — that path returns 502 (the upstream fetch was issued) where the equivalent `..%2f` returns 400 (the guard fires).

The escape does not currently work: CloudFront/S3 treat the encoded dots as a literal key (403 against a path that returns 200 unencoded), so confinement today rests on S3 key literalness rather than on this code. That is worth knowing precisely because `ASSET_PATH_PREFIX`'s own note at `:129-133` names the payoff if it ever slips — a presigned-upload `.html` under `user-images/` served as a document on the OAuth origin — and any normalizing hop added later (a CloudFront Function, a URI-rewrite behavior, a move off the S3 origin) would arm it with no test noticing.

Would re-running the five guards over a `decodeURIComponent`'d copy, or rejecting `%` outright, make the docstring true?

workers/wrangler.jsonc
25eliu: Confirmed, and it was worse than a fairness question — I ran the burst you pointed at. 14 unique `/cdn-asset/` requests against staging:

```
statuses: 502 502 502 502 502 502 502 502 502 502 429 429 429 429
```

Requests 1-10 served, 11-14 all 429. So the bucket does enforce, and one card render (9 distinct rewritten urls on a measured prod page + `/embed-html/` itself = 10, before Card.js's lazy chunks) sits exactly at the ceiling. **The feature was throttling its own assets** — the chart would fail to draw, which is the symptom this PR exists to fix. To answer your question directly: no, the burst had not been run, and what was carrying staging is the per-colo under-count, not headroom.

Fixed with its own binding, `NATIVE_CARD_RATE_LIMITER` at 200/60 (~12-20 renders per minute per IP), declared in all three blocks. Still fails open.

One correction to the quota-draining half: for claude.ai those two axes have *different* keys — asset fetches carry the user's browser IP, `tools/call` originates from Anthropic's servers. It does apply to anonymous callers hitting `mcp.tako.com` directly. The self-throttling half stands on its own and was the serious one.

workers/wrangler.jsonc
25eliu: Dropped. Your measurement reproduced: `${url.search}` reaches CloudFront and lands in the `cacheEverything` cache key, so each nonce buys a full ~1.48 MB origin fetch, decode, whole-body rewrite copy and 1.48 MB of `ACAO: *` egress from a ~200-byte request.

To your question — yes, safe: `rewriteCdnUrls` never emits a query, and these are content-hashed assets under `archive/<sha>/`, so the hash *is* the cache-buster. I checked rather than assumed: 0 of 9 rewritten urls on a real production page carry one. Test pins that the nonce no longer reaches upstream.

workers/scripts/smoke.ts
25eliu: Yes to your suggestion, and thank you for catching the workflow half — that was the bigger finding.

The 404 now `fail()`s when `SMOKE_BASE_URL` matches a deployed host and warns only off one. You are right that it had it backwards: a 404 on `mcp.tako.com` is the failure, and a warning exits 0.

On the trigger: confirmed, prod was never smoked at all. `workers-smoke.yml` now accepts `release` and derives the target from the upstream event using the *same* mapping `workers-deploy.yml` uses to pick it (push → staging, release → production), so the two cannot disagree about which env was deployed. Upstream `workflow_dispatch` stays excluded — its env comes from an input `workflow_run` does not carry, and guessing is what the old comment rightly warned produces a misleading badge.

Worth flagging: this makes the smoke's `tako_visualize` call a per-release cost on prod (one credit, one card).

workers/scripts/smoke.ts
25eliu: Right, and the causal claim is gone. The message now says PUBLIC_CDN_URL is "either unset or malformed (both decline identically)" and points at the worker's `native-card proxy disabled` log line for which — verified that string exists at the `resolveProxyOrigins` catch.

Your framing is the part I had backwards: since this PR's entire content is typing a new URL into the production block, "set but broken" is the *likely* case here, not the edge one — and a trailing slash is enough.

workers/scripts/smoke.ts
25eliu: All three closed, and yes — your suggested shape is what I implemented.

1. **Origin pinned.** Now asserts `` `${baseUrl}/cdn-asset/` `` rather than a bare substring, so a prod deploy carrying a staging `PUBLIC_MCP_URL` fails instead of passing while every asset loads from the wrong worker.
2. **Assets actually fetched.** It retrieves a rewritten url — preferring the JS entry, since Card.js's absence means no chart at all — and asserts 200 plus `ACAO: *`, the header the route exists to add. That also gives the measured-not-structural `ASSET_PATH_PREFIX` confinement teeth: a build moving the entry bundle out of `archive/` while leaving the fonts in would otherwise keep `rewrites > 0` and stay green.
3. And it automates the second unchecked box in the PR description, as you noted it would.

workers/src/tools/_chart_widget.ts
25eliu: All five folded in. You are right that `env.ts:82` was the load-bearing one — "Leave unset in production. Set it on staging to run the experiment" read as an instruction to revert this PR, at the binding's own declaration site.

Fixed: `env.ts:52` ("EXPERIMENT SWITCH, unset everywhere by default"), `env.ts:82`, `index.ts:54`, `index.test.ts:63`, and `widget-dom.test.ts` (the text is at :993, not :984). Also added the new limiter to the README binding table so the operator-facing list is not the next thing to drift.

workers/src/embed_proxy.test.ts
25eliu: Both added, and you were right that the ordering one has teeth.

**Rewrite through the handler** — asserts the 200 body contains `https://mcp.tako.com/cdn-asset/archive/abc/` and that the CDN origin is gone from the markup. Origin-qualified for the reason in the `smoke.ts:444` thread.

**`NATIVE_CARD_RATE_LIMITER`** (renamed, per the C1 thread) — deny path on both routes with `expect(fetchMock).not.toHaveBeenCalled()`, the fail-open catch, the absent-binding case, and call ordering.

I mutation-tested the ordering assertion rather than trusting it: relocating the asset-route `rateLimited` call below its fetch turns it red (`expected [ 'fetch' ] to deeply equal [ 'limit', 'fetch' ]`), and reverting turns it green. It would have caught exactly the refactor you described.

workers/src/embed_proxy.ts
25eliu: Reproduced and fixed. Confirmed your staging measurement: `%2e%2e%2f` → 502 (the fetch was issued) where `..%2f` → 400 (guard fires).

The guards now also run over a `decodeURIComponent`'d copy, and a malformed escape is refused outright. I went with decoding rather than rejecting `%` because the docstring defends an encoded slash as a legitimate S3 key — banning `%` would have made the comment true by making the code wrong in a different way. There is a test that an encoded slash still serves.

Your framing is the reason I treated this as more than hygiene: confinement rested on S3 key literalness rather than on this code, so any normalizing hop added later would have armed it with nothing going red.

robertabbott (approved):

PR #210, opened 2026-08-04
fix: OAuth consent deny + account switch, plus ChatGPT app-review compliance

25eliu:
**This PR carries two independent batches.** They share no files; they are together because both were ready and the second one blocks the OpenAI app submission. Reviewable separately, commit by commit.

| | Commit | What |
| --- | --- | --- |
| **Part 1** | `09a0d4a` | OAuth consent: deliver `access_denied` on Cancel, and let users switch accounts |
| **Part 2** | `c9e404f` | ChatGPT app-review compliance: remove upsells, internal ids, and iframe analytics; disclose public chart creation |

---

# Part 1 — OAuth consent (`09a0d4a`)

Two gaps on the OAuth consent page, found while answering a question about why connecting the staging MCP skips the login form and lands straight on "Signed in as …". That part is correct — it's standard OAuth SSO, and the session-cookie check in `handleAuthorize` does exactly the right thing. These two are what turned up alongside it.

Neither is tracked: TAKO-2679 shipped the consent page and closed 2026-05-01, and none of its follow-ups (TAKO-2698–2702) cover them.

### 1. Cancel never told the client anything

```html
<button type="button" onclick="window.history.back()">Cancel</button>
```

`access_denied` appeared **nowhere** in the codebase. Declining a connector left the host waiting for a callback that never arrived, and in a fresh tab or popup with no history the button did nothing at all.

Both buttons are now real submits carrying `action=allow|deny`, and a denial goes to the client via the redirect URI per **RFC 6749 §4.1.2.1** — using the `redirectAuthError` helper that already served the other error paths in the same file.

Two deliberate choices:

- **Deny sits above the session check.** `redirect_uri` is already known-registered by that point, so a refusal needs no live session. Had it stayed below, an expired cookie would have returned `401 session expired` and the client would still hang — the exact failure being fixed.
- **A missing `action` is a 400, not an implicit allow.** An absent field must never read as consent.

**Allow is first in DOM order.** With two submit buttons, Enter activates whichever comes first in tree order, and that default must never be deny. `flex-direction: row-reverse` keeps Cancel reading on the left.

### 2. No way to change accounts

The page stated `Signed in as <email>` with only Cancel and Allow — no switcher. Signed into the wrong Tako account meant leaving the flow, logging out on the web app, and starting over. That bites hardest when juggling accounts across staging and prod.

Adds a "Not you? Sign in with a different account" link that sets **`prompt=login`** (OIDC Core §3.1.2.1), forcing a fresh sign-in and clearing the Worker's own session cookie. Using the standard parameter means any client already sending it gets the behavior for free.

`prompt` stays **off** the form action — it's GET-only, and on the POST it would bounce Allow back to `/login` instead of issuing a code.

`prompt=none` is deliberately **not** honored: doing it properly means returning `login_required` via the redirect URI rather than rendering anything, and nothing sends it today. Half-implementing it would be worse than ignoring it.

### Two details worth a reviewer's eye

**The redirect needed a real `Headers` instance.** `forceLogin` sets two `Set-Cookie` headers (state + cleared session) and the previous object literal would have silently kept only one. There's a test asserting both survive.

**Clearing the session is host-scoped.** `buildSetCookie` deliberately sets no `Domain`, so this signs the user out of this Worker only — not trytako.com, not Google. On the email path they can enter a different address; **Stytch's Google button may still silently re-select the same account.** Making that reliable needs `prompt=select_account` passed through to the Stytch OAuth start, which I left out because I can't confirm their SDK exposes it. Flagged below as a follow-up.

### Tests

The 14 existing `handleAuthorize` consent POSTs gained the body the handler now requires. Rewritten by script after confirming both call shapes matched exactly (13 + 1 = 14, every other POST in the file already carried a body), rather than hand-editing 14 sites.

8 new tests:

| | |
|---|---|
| denial delivery | 302 to `redirect_uri`, `error=access_denied`, `state` preserved, no `code` |
| denial without a session | 302, not 401 — the dead-end guard |
| no `action` | 400, and no redirect, so no code could issue |
| DOM ordering | `value="allow"` precedes `value="deny"`; no `history.back` remains |
| `prompt=login` | 302 `/login`, state cookie set **and** session cookie cleared (`Max-Age=0`) |
| `prompt=login` | does not render consent |
| no `prompt` | still renders consent (regression guard) |
| switch link | carries `prompt=login` + validated params; form action carries no `prompt` |

---

# Part 2 — ChatGPT app-review compliance (`c9e404f`)

Five findings from OpenAI's review of the ChatGPT app submission. Independent of Part 1 — no shared files — but batched here so the whole approval blocker moves as one review.

Common thread: everything below is either **model-visible text** or **the surface OpenAI reviews**. Each item states what changed and why it had to.

### 2.1 Four upsells removed from model-visible errors

The anonymous-tier messages ended with "get an API key at `https://tako.com/account/` …":

| Constant | Was | Now |
| --- | --- | --- |
| `FREE_TIER_LIMIT_MESSAGE` | "Try again in a minute, or get an API key at … for higher limits." | "Rate limit reached for anonymous access. Try again in a minute." |
| `FREE_TIER_GLOBAL_LIMIT_MESSAGE` | "Try again shortly, or get an API key at … for dedicated access." | "Anonymous access is at capacity right now. Try again shortly." |
| `FREE_TIER_BATCH_MESSAGE` | "… or get an API key at … for full access." | "Batch requests are not supported for anonymous access. Send one JSON-RPC request per POST." |
| `FREE_TIER_CREDITS_MESSAGE` | "Get an API key at … to continue with your own account." | "Anonymous access is temporarily out of shared capacity. Try again later." |

**Why it matters more than it looks.** These are not HTTP error bodies a human reads — `freeTierLimitResponse` deliberately delivers them as an **HTTP 200 JSON-RPC tool result**, because a 429 makes the SDK client throw and the model never sees the text. So the host feeds each one straight to the model as tool output. That is an upsell delivered *by the app*, which OpenAI's commerce policy forbids for digital services, subscriptions, tokens, and credits. Existing paid-account functionality is untouched — an authenticated connection behaves exactly as before, and `get_credit_balance` (opt-in, not on the ChatGPT surface) still answers account questions when a user asks one. Only advertising it from a tool response is gone.

`FREE_TIER_GLOBAL_LIMIT_MESSAGE` was **not** in the original report — it was found by grepping the module rather than trusting the list.

**Enforcement inverted, not just edited.** A guard test previously *required* the account URL in all four (`expect(message).toContain("https://tako.com/account/")`). It now bans URLs and account/pricing/upgrade wording across every message the module can emit, so a future message is covered by adding one line to `ALL_FREE_TIER_MESSAGES`.

Incidental finding: the path those messages pointed at was already **stale**. The API-key page is `/console/api-keys`, now recorded in `workers/README.md` and in the test comment — documentation, not tool output.

### 2.2 `request_id` out of both response channels

`tako_search` / `tako_answer` advertised `request_id` in `outputSchema` and printed it in the markdown footer. OpenAI's review names request/trace/session/debug identifiers as things a tool response should not carry unless strictly necessary.

**The advertised schema is the real gate.** Removing the footer line alone would have left the id in `structuredContent`; `pickDeclared` in `mcp.ts` strips keys the schema does not declare, so deleting the field from the advertised shape is what actually removes it from the wire. The handlers still receive it from the backend.

**Why a log line had to be added at the same time.** Nothing logged a *successful* call's `request_id` — only wire-guard failures did. Removing it from responses without that would have made "this answer looked wrong" untraceable, since a user could no longer quote an id. `logToolRequestId` (`tools/_log.ts`) now records `tool` + `client` + `request_id` per call. Runbook: `wrangler tail <worker> --search "request_id="`.

**Kept deliberately:** `run_id` and `thread_id` on the agent tools. A caller cannot poll a run to completion without `run_id`, or continue a conversation without `thread_id` — those are the caller's control flow, not our debugging. The new guard test encodes that distinction rather than banning every `*_id`.

**Surface-wide enforcement:** `mcp.conformance.test.ts` reads the real `tools/list` response and rejects `request_id` / `trace_id` / `correlation_id` / `session_id` / `debug_id` on *any* tool — including one added later by someone who never read this PR.

### 2.3 `card_id` off `tako_visualize`'s output

It carried the identical string to `pub_id`, which the widget needs to resolve the chart. A second copy under an internal-sounding name was pure duplication in front of the model. Callers get `webpage_url` (share), `embed_url` (embed), `pub_id` (render).

### 2.4 The footer is now cost-only, and absent when there is no cost

With `request_id` gone, cost was the footer's only remaining member — so an unmetered call would have rendered a bare `_ _`. `renderFooter` returns `undefined` there and both renderers skip the block.

### 2.5 Two tool descriptions were left claiming otherwise

`tako_search` and `tako_answer` both told the model its rows ride in `structuredContent` "alongside machine essentials (**request_id**, usage, …)". After 2.2 that is a false statement about the payload, in text the model reads on every session. Corrected in both, and the generated `registry/server.json` + `registry/lhm.plugin.json` regenerated to match.

Found by review after the code change, not before — which is the argument for regenerating the registry as part of the same commit rather than letting `registry:check` catch it at deploy time.

### 2.6 `tako_visualize` now says it publishes

The live description said only "embeddable" — the mechanism, with the consequence left out. It now leads with:

- the card is **PUBLIC** and **PERSISTENT**;
- the supplied data is **sent to and stored by Tako**;
- **anyone with the returned link can view it** without signing in;
- confirm the user wants a public chart **before** calling;
- **never** put passwords, API keys/tokens, payment-card or bank details, health information, government identifiers, precise home addresses, or third-party personal data in `components` — aggregate, anonymize, or decline.

**Why the wording is aimed at the model.** The model assembles the `components` payload, so it is the only party positioned to refuse — and a model that does not know the output is public cannot warn the user before pasting their data into it. This now agrees with the annotations the tool already shipped (`readOnlyHint: false` plus ChatGPT `openWorldHint: true`, which makes the call confirmation-worthy) and with the `tako_visualize` justifications in `chatgpt-app-submission.json`.

### 2.7 Google Tag Manager out of the reviewed iframe

On ChatGPT the widget commits to a real cross-origin `<iframe src="https://tako.com/embed/…">`, and that page loaded GTM — a third-party beacon **inside the frame OpenAI's iframe policy reviews**.

Fixed with `disable_tracking=true`, an **already-supported flag** on Tako's embed route: it gates out the GTM bootstrap (`templates/embed/Card.html`) and excludes the load from Tako's impression counters (`embed_impressions.should_track_impression`). No change needed in the Tako web repo, so nothing here waits on another deploy.

**Applied at the three `iframe.src` assignment points, not to `embed_url` itself.** `embed_url` is also the shareable link the model hands the user, and a person clicking it in their own browser is an ordinary tako.com visit that should keep counting. Scoping the flag to loads the *widget* performs means only agent-driven renders stop counting — the accepted trade — while the click-through is unaffected. The `/embed-html/` proxy fetch (Claude's native-card path) sends it too; its existing GA strip stays as the half that does not depend on an upstream template honouring a query flag.

**Verified against production**, not assumed: the same embed page fetched with the flag references only Tako's own CDN and the S&P attribution links — zero `googletagmanager`. Asserted by `chatgpt-path.test.ts` (`renderedIframe`) and `widget-dom.test.ts` (`EMBED_IFRAME_SRC`), both exact compares so a regression that drops the flag fails.

### 2.8 `docs/chatgpt-app-review.md`

OpenAI asks for prose, not just metadata: why interactivity requires the iframe, and what is inside it. The doc records the justification (hover values, series toggling, interval switching, drill-in — a PNG cannot answer "what was Q3 exactly?"), the single pinned `frameDomains` origin with no wildcards, the PNG fallback for hosts that block frames, and the **measured** contents of the framed page: three external origins, no ads, no checkout, no signup, no upgrade UI, zero `<a>` elements. Each policy item is paired with the test that enforces it, so the next person answering a review question does not re-derive any of this.

### Not in this PR — needs counsel

Recorded as §5 of that doc:

- **Privacy policy** (`app/frontend/src/pages/privacy-policy/` in the Tako web repo) states no retention periods and does not cover OAuth account identifiers, search/answer queries, IP/rate-limit data, public chart data and its retention, Stytch and Cloudflare as processors, or chart deletion / connection revocation. It also needs to reconcile the site's zero-data-retention claim with `tako_visualize`, which creates persistent public charts by design.
- **Age policy:** OpenAI requires suitability for 13–17; Tako's terms require 18+ and the privacy policy says the service is not intended for under-18s.

---

## Test plan

Both parts, run together on the merged branch:

- [x] `npm run typecheck` — clean on all four tsconfig projects
- [x] `npm test` — **1,079 pass** (948 + 85 + 46 across the three vitest configs)
- [x] `npm run registry:check` / `npm run schemas:check` — no drift
- [x] Production embed page fetched with and without `disable_tracking=true`, confirming GTM is the only third-party beacon and that the flag removes it

Part 1 — manual, staging:

- [ ] Connect the MCP, click **Cancel**, confirm the host reports a denial rather than hanging
- [ ] Click **"different account"**, confirm `/login` renders a fresh form on the email path
- [ ] Same via **Continue with Google** — expected to re-select the same account (see above); confirms whether the `prompt=select_account` follow-up is needed
- [ ] Press **Enter** on the consent page and confirm it approves rather than denies

Part 2 — manual, staging:

- [ ] Anonymous call over the rate limit, confirm the relayed text carries no link or account mention
- [ ] ChatGPT chart render, confirm the iframe src carries `disable_tracking=true` and the chart is still interactive
- [ ] Claude chart render, confirm the native card still upgrades (the proxy fetch changed too)

## Follow-up, not in scope

- `prompt=select_account` through to the Stytch OAuth start, so Google-path account switching is reliable
- `prompt=none` → `login_required` via the redirect URI, if any host ever sends it
- Privacy policy and the 18+ vs 13–17 age conflict — counsel, tracked in §5 of `docs/chatgpt-app-review.md`
- A narrower embed flag that suppresses only third-party analytics while first-party impressions keep counting, so agent-driven renders stay measurable


workers/src/oauth/handlers.ts
```diff
@@ -829,9 +903,10 @@ function consentPage(args: {
 <p>${safeName} is requesting access to your Tako account. Approving will let it call the Tako MCP server on your behalf.</p>
 <div class="who"><span class="who-dot"></span> Signed in as <strong>${safeEmail}</strong></div>
 <form method="POST" action="${safeAction}">
-  <button type="button" onclick="window.history.back()">Cancel</button>
-  <button type="submit">Allow</button>
+  <button type="submit" name="action" value="allow">Allow</button>
+  <button type="submit" name="action" value="deny">Cancel</button>
```
robertabbott, line 913: Making Cancel a `type="submit"` also makes it match `button[type=submit] { background: var(--accent); color: var(--on-accent); border-color: var(--accent); }` two lines up. `git log -S` puts that selector in #54, where it existed precisely because Allow was the *only* submit and Cancel was `type="button"` — so both buttons now render as identical filled accent buttons on the consent screen.

`row-reverse` compounds it: Tab order is Allow-then-Cancel while visual order is Cancel-then-Allow, so a keyboard user's first Tab lands on the right-hand button. The new DOM-ordering test asserts `value="allow"` precedes `value="deny"`, but nothing asserts the two are visually distinguishable, which is why CI is green on it.

Should Cancel carry a `.secondary` class, or the rule narrow to `button[value=allow]`?

workers/src/tools/_chart_widget.ts
```diff
@@ -1537,7 +1565,11 @@ const WIDGET_HTML = `<!doctype html>
     var useIframe = hasOpenAiRuntime() && validEmbed;
 
     if (useIframe) {
-      var themedUrl = withHostTheme(url);
+      // Tracking-free, then themed. This is the ChatGPT path: a committed
+      // cross-origin iframe of the embed page, and the surface OpenAI's iframe
+      // policy reviews — so it must not load a third-party beacon. See
+      // \`withoutTracking\`.
+      var themedUrl = withHostTheme(withoutTracking(url));
```
robertabbott, line 1572: The no-image `validEmbed` fallback further down (`} else if (validEmbed) {` → `if (frame.src !== url) frame.src = url; armEmbedOrigin(url);`) is the one iframe load that does not go through `withoutTracking`.

It looks unreachable today — `buildSearchOutput` and `buildChartUrls` emit `embed_url` and `image_url` together, so `validEmbed && !validImage` needs a non-`http(s)` `image_url` the server never produces — and no test covers it. But the `probeUrl` comment above and `docs/chatgpt-app-review.md` §4 both state the invariant absolutely ("every iframe load the widget performs"), and `withoutTracking` is idempotent and free.

Worth closing so the invariant holds by construction rather than by the mutual-presence property continuing to hold?

workers/src/mcp.conformance.test.ts
```diff
@@ -192,6 +192,30 @@ describe("published outputSchema conformance", () => {
...
+    //
+    // Deliberately NOT banned: `run_id` and `thread_id` on the agent tools.
+    // Those are strictly necessary — `run_id` is the only way to poll a run to
+    // completion and `thread_id` the only way to continue a conversation, so a
+    // caller cannot use the tool without them. The test is about identifiers
+    // that exist for OUR debugging, not the caller's control flow.
+    const banned = /^(request|trace|correlation|session|debug)_id$/;
+    const offenders: string[] = [];
+    for (const [name, schema] of published) {
+      for (const key of Object.keys(schema.properties ?? {})) {
```
robertabbott: `Object.keys(schema.properties ?? {})` only walks the top level, so this is shallower than the guarantee the doc states ("rejects any `request_id` / `trace_id` / … on *any* tool").

`agentRunSchema` already nests `result.request_id`. That is harmless today because `slimAgentRunStructured` drops `result` and `agentRunSlimOutputShape` never declares it — but it is exactly the shape that would slip past this test if a future tool advertised a nested result object, which is the case the test is written to catch.

Should the key walk recurse into `properties` and `items`?

robertabbott (commented):
Reviewed both commits. Three inline questions, plus one on the commit metadata.

**Commit type vs. the published contract.** Releases here are conventional-commit driven (`release: 0.17.0` + generated CHANGELOG), and both commits are `fix:`. `c9e404f` removes `request_id` — previously a **required** declared field — from two published `outputSchema`s, plus `card_id` from `tako_visualize`. A programmatic MCP client reading `structuredContent.request_id` breaks silently on deploy. Is a `BREAKING CHANGE:` footer wanted so the version reflects it, or is the reasoning that no programmatic consumer reads those fields?

For the record, things I checked and found correct rather than assumed: `pickDeclared` really is the gate (it iterates `Object.keys(shape)`, so dropping the zod key is what removes the field from `structuredContent`); all five `FREE_TIER_*_MESSAGE` constants are covered by `ALL_FREE_TIER_MESSAGES`, and `git grep` finds no remaining account/API-key/pricing copy anywhere under `workers/src`; the deny branch sits above the session check but below client + `redirect_uri` validation, which is the order that makes it safe; `prompt` is deliberately absent from `StateCookieClaims`, so `completeStytchLogin`'s rebuilt `/authorize` cannot re-trigger `forceLogin` in a loop; and the consent page is `no-store`, so Back after "different account" re-issues the GET rather than showing a stale Allow.

workers/src/oauth/handlers.ts
25eliu: Narrowed the rule to `button[value="allow"]` — keyed on the decision rather than the element type, so Cancel is outlined again and Allow keeps the accent. You were right about the provenance: that selector only worked because Allow was the sole submit, and making Cancel a real submit is exactly what swept it in.

Your point about *why* CI stayed green is the more useful half, so that's the part I tested. Ordering and appearance are different properties and only the first was asserted. The new test asserts the accent rule cannot match Cancel (`not.toMatch(/button\[type=submit\][^}]*var\(--accent\)/)`), and that both buttons remain `type="submit"` — so a "fix" that reverted Cancel to `type="button"` would break denial and fail here too. Confirmed it fails against the old selector before keeping it.

On Tab order: leaving it. Tab follows DOM order (Allow first) while `row-reverse` puts Cancel on the left, so the first Tab does land on the right-hand button. That's inherent to wanting Enter to mean Allow — there's no HTML way to set a form's default submit other than tree order, so the alternative is the keyboard default meaning *deny*. I'd rather have the odd Tab landing than that. Happy to revisit if you disagree.

workers/src/tools/_chart_widget.ts
25eliu: Closed — that branch now goes through `withoutTracking` like the others.

Your reachability read matches mine: it needs a non-`http(s)` `image_url` beside a valid `embed_url`, and `buildSearchOutput` / `buildChartUrls` always emit the pair, so there's no path to it today. Fixed anyway for the reason you gave — §4 states the invariant absolutely, and "holds because two fields happen to always ship together" is a weaker guarantee than "holds by construction."

Added a jsdom test that actually reaches it (no `window.openai`, valid `embed_url`, `image_url: "ftp://…"` so the image path rejects) and asserts the exact src with `&disable_tracking=true`. **Confirmed it fails without the fix** — stashed the change and watched it go red — so it's covering the branch rather than passing incidentally.

One thing I deliberately did not change: the branch still doesn't apply `withHostTheme`, unlike the ChatGPT path. The invariant you cited is about tracking, and theming an unreachable fallback is a behavior change beyond the comment. Say the word if you'd rather it match.

workers/src/mcp.conformance.test.ts
25eliu: Yes — it now recurses through `properties` and `items` (the latter handling both a single schema and a tuple), reporting dotted paths like `t.result.request_id` / `t.cards[].trace_id`.

Ran it against the real `tools/list` first, since recursing could have surfaced a genuine offender: it found none, so no schema changed. Your read on `agentRunSchema` holds — `slimAgentRunStructured` drops `result` and the slim shape never declares it, so the nested `request_id` never reaches the published surface.

Which is exactly why I didn't stop at "still green." Against today's schemas the recursive and top-level walks are indistinguishable, so passing proves nothing about the mechanism. Three tests now pin it on shapes the old walk provably missed: a `request_id` nested one level down, a `trace_id` inside array items, and — the one I'd want if someone later widens the pattern — `run_id`/`thread_id` staying unflagged at depth, so the recursion can't quietly start catching the caller's control flow.

workers/src/mcp.conformance.test.ts
```diff
@@ -145,6 +145,40 @@ function dummyFor(schema: unknown): unknown {
...
+    items?: unknown;
+  };
+  const found: string[] = [];
+  for (const [key, child] of Object.entries(s.properties ?? {})) {
+    const childPath = `${path}.${key}`;
+    if (banned.test(key)) found.push(childPath);
+    found.push(...bannedKeysIn(child, childPath, banned));
+  }
+  // `items` may be a single schema or a tuple of them.
+  const items = Array.isArray(s.items) ? s.items : s.items === undefined ? [] : [s.items];
```
robertabbott: The recursion covers `properties` and `items`, but not `anyOf` — and `anyOf` is how a nullable object reaches the published schema. The proof is in this same file: `dummyFor` at line 118 has an explicit `anyOf` branch, commented "Nullable fields arrive as anyOf:[T, null]".

Live on the two headline tools. `searchSlimOutputShape.usage` and `answerSlimOutputShape.usage` are `usageAdvertisedSchema` (`_search_results.ts:611` — an object) `.nullable()`, so they publish as `anyOf: [{properties: {total_cost_usd}}, {type: "null"}]`. That wrapper has no `properties` and no `items`, so `bannedKeysIn` returns `[]` without entering the object at all. Same shape on `agentRunSlimOutputShape.error` and the `entity` field at `_render_markdown.ts:495`.

This is the original point one layer up: the walk is still shallower than the rule it states, and it passes for a property of today's schemas rather than of the check. The three new mechanism tests all build plain `{properties}` / `{items}` shapes, so none of them would catch it either.

Worth adding `anyOf` (and its `oneOf` / `allOf` siblings) to the descent, with a mechanism test that wraps the nested object in one?

robertabbott (approved):
All four addressed. One non-blocking follow-on inline — same family as the sweep-depth point, one layer up.

Things I verified rather than took on trust, in case it saves you re-deriving them:

- **The 0.18.0 claim holds.** `release-please-config.json` sets `"bump-minor-pre-major": true` and the manifest is at `0.17.0`, so `BREAKING CHANGE:` lands as 0.18.0, not 1.0.0.
- **The separate-commit reasoning holds too.** main preserves per-commit history — 0.17.0's CHANGELOG carries `4cefe2c`, `8daa0d4` and `27283f1` from one branch as three entries — so the footer reaches release-please intact without an amend.
- **`button[value="allow"]` is specificity 0-1-1** against the bare `button` rule's 0-0-1, so Cancel genuinely falls back to the transparent/`--border` outline in both colour schemes. Pinning both buttons as real submits in the same test is the assertion I'd have asked for — it makes a revert-to-`type="button"` "fix" fail there instead of silently breaking denial.
- **The new widget test reaches the branch** rather than asserting near it: `ftp://` `image_url` beside a valid `embed_url` clears the no-chart short-circuit, fails both image validators, and with no `window.openai` falls through `useIframe` → `validImage` → `validEmbed`.

One thing I raised that you correctly did not change: Tab order still runs Allow→Cancel against the visual Cancel→Allow. Now that the two are visually distinct again, reversing two sibling buttons to keep Enter off the destructive choice is the conventional trade.

workers/src/mcp.conformance.test.ts
25eliu: You're right, and the `dummyFor` citation is what made it undeniable — that comment is sitting eleven lines above the walk I wrote.

Verified before touching it. `tako_search.usage` publishes as exactly what you described:

```json
{"anyOf":[{"type":"object","properties":{"total_cost_usd":{"type":"number"}},
  "required":["total_cost_usd"],"additionalProperties":{}},{"type":"null"}]}
```

No `properties`, no `items` on the wrapper, so `bannedKeysIn` returned `[]` without entering it. I probed the live `tools/list` to size the blindness rather than reason about it: **7 `anyOf` branches carrying `properties`** on the published surface today. Live, not hypothetical.

**So the keyword list is gone.** It's been wrong twice, and adding `anyOf`/`oneOf`/`allOf` to it would just be a third guess. The walk now descends the whole tree and treats any `properties` map it finds as property names, whatever nests it — unions, `not`, `if`/`then`/`else`, `additionalProperties`, `$defs`, `dependentSchemas`, and whatever a future serializer emits. That makes coverage a property of the function instead of a property of today's schemas, which was your point all three times.

Two positions stay special-cased, both because their keys aren't property names: `enum`/`const` hold instance data (an object in there may legitimately have a `properties` key, and reading it as a subschema would invent an offender — there's a negative test), and `patternProperties` keys are regexes.

Ran the deeper sweep against the real `tools/list` before anything else, since it could have surfaced a genuine offender: none, so no schema changed.

Four new tests, three confirmed to fail against the previous walk — the verbatim nullable-`usage` shape, `oneOf`/`allOf`, and `$defs` as a keyword the function was never taught. And you were right that the earlier three couldn't have caught it: they pass under both walks, since they build plain `{properties}`/`{items}` shapes.

`b5cffbb`. 1,088 tests, typecheck clean.

PR #211, opened 2026-08-04
fix(tools): accept a stringified array where hosts serialize one

25eliu:
## What broke

OpenBB Copilot called `tako_answer` with `sources` as the JSON **text** `["data","web"]` instead of an array. The MCP SDK validates tool arguments against our zod schema before any handler runs, so the call died there. Reproduced byte-for-byte against prod:

```
MCP error -32602: Input validation error: Invalid arguments for tool tako_answer:
  [{"expected":"array","code":"invalid_type","path":["sources"],
    "message":"Invalid input: expected array, received string"}]
```

That whole error string is ours. The tell: the host's own argument panel rendered `sources` as a two-element array (`0 → data`, `1 → web`), so it JSON-parsed for display what it put on the wire as text.

Nothing was wrong with what the model meant. It picked the right tool, the right query and the right sources, and we rejected it on serialization. What a model does next after that error is retry with a differently-wrong shape or abandon Tako for a generic web search. Both are billed to the user. Copilot did the second.

## Why the coercion belongs on our side

OpenBB's [documented MCP integration](https://docs.openbb.co/workspace/developers/ai-features/mcp-tools) renders our JSON Schema into prose in the system prompt, reading only `type` and `description` per param and dropping `items` entirely. The single function actually exposed to the model is an untyped passthrough:

```python
"parameters": { "type": "object", "description": "The arguments to pass to the tool",
                "additionalProperties": True }
```

with `gpt-4o` on the legacy `functions=` API. No schema reaches the model natively, so no grammar constraint and no client-side validation can apply anywhere in that chain. **This server is the first component that knows `sources` is an array of enums.**

The LLM half of this is an ecosystem-wide class, not one vendor's quirk: [qwen-code #379](https://github.com/QwenLM/qwen-code/issues/379), [koog #2017](https://github.com/JetBrains/koog/issues/2017), [python-sdk #820](https://github.com/modelcontextprotocol/python-sdk/issues/820), [fastmcp #932](https://github.com/PrefectHQ/fastmcp/issues/932). Server-side coercion is established practice: [FastMCP](https://gofastmcp.com/servers/tools) defaults to `strict_input_validation=False` to accept string representations of scalars. It stops short of parsing stringified JSON for structured params, which is exactly the case here.

Worth noting the blast radius: the code above is OpenBB's **reference implementation**, so every Workspace custom agent built from that doc inherits the behavior.

## The change

`looseArray()` (`workers/src/tools/_loose_array.ts`) reshapes a string, then hands off to the real array schema:

Accepted on **every** wrapped field:

| sent | accepted as |
|---|---|
| `'["data","web"]'` | `["data","web"]` (the observed case) |
| `'data'` | `["data"]` |

Opt-in per field, because each is unsafe on the wrong item domain:

| option | form | enabled on | why not everywhere |
|---|---|---|---|
| `commaSeparated` | `'data,web'` → `["data","web"]` | `sources` (closed enum) | A comma is legal in an opaque string. On `urls` it split one wikipedia URL into two that both validate, billing two fetches for the wrong pages. |
| `jsonObjectAsItem` | `'{"id":"a"}'` → `[{"id":"a"}]` | `components` (object items) | On a primitive domain it can never validate, and it replaces `expected array, received string` with a per-item error about element 0 of an array never sent. |

JSON-shaped input is parsed or passed through untouched, never comma-split.

Applied to all 8 array-typed inputs: `sources` + `node_ids` on `tako_answer` / `tako_search`, `sources` on `tako_agent` / `tako_agent_start`, `urls` on `tako_contents`, `components` on `tako_visualize`. Each coercion logs `[mcp] input coerced field=<tool>.<field> from=<form> items=<n>` — the form and count, never the value — so post-deploy incidence is measurable rather than inferred.

Two things it deliberately does not do:

- **It does not widen the contract.** `sources: "bing"` still 400s, and every item type, `min` and `max` still applies. A string that cannot be reshaped (`''`, `'[not json'`) is passed through unchanged so the caller gets the honest "expected array, received string" for what they actually sent.
- **It does not change the advertised schema.** `z.preprocess` serializes to the wrapped schema verbatim under the SDK's `z.toJSONSchema(…, { io: "input" })`, so well-behaved clients still see `type: "array"` with the item enum, `minItems`, `default` and `description`. A `z.union` would publish `anyOf: [array, string]`, which both advertises the sloppy form as legal (teaching every other client to send it) and trips hosts that reject `anyOf` in tool parameters. Leniency belongs on the accepting side only, never in the contract. There is a test asserting that published-schema equality.

## Tests

- `_loose_array.test.ts` — the accepted forms, the rejected ones, default preservation, and published-schema equality against the unwrapped array.
- A registry-walking coverage guard: it derives array-typed fields from each tool's **published** JSON Schema and fails by name on any future array input someone forgets to wrap. Currently 8 fields.
- Three end-to-end tests in `mcp.test.ts` over the real server on an in-memory transport, proving the coercion survives `registerTool` (where the SDK rebuilds our `.shape` into its own `z.object` and validates), and that the upstream request body carries the right per-source keys.

**Scope:** top-level array inputs only. The five arrays nested inside `components[]` are unwrapped, and do not need wrapping for the observed failure — a host that stringifies the whole array yields correctly typed nested values once it parses. The test suite states that boundary as an assertion instead of leaving it implied.

`npm run typecheck` (all 4 projects), `registry:check`, `schemas:check` and all three vitest configs pass: 1000 + 85 + 46. Every guard is mutation-tested: re-enabling the comma split on `urls`, dropping a wrap while still compiling, and swapping the preprocess for a union each fail the assertion that claims to cover it.

## Test plan

- [ ] CI green
- [ ] After deploy, confirm on prod that `{"sources": "[\"data\",\"web\"]"}` on `tako_answer` returns an answer rather than -32602, and that `{"sources": "bing"}` still returns -32602
- [ ] After deploy, confirm `tako_contents` with `urls: "https://en.wikipedia.org/wiki/Washington,_D.C."` fetches ONE page and bills once
- [ ] Watch `[mcp] input coerced` in Workers Logs for a few days: which fields, which forms, and whether OpenBB traffic recovered
- [ ] Ask OpenBB to send a real array (their agent should also recover from a validation error instead of pivoting to web search); this fix means their current build works against us either way

## Follow-on, not in this PR

The sibling gap is still open: the same host on the same prose-flattened path can send `preview_rows: "5"` or `strict: "true"`, and those still 400. FastMCP coerces exactly those by default. Same helper shape, small change, worth doing once we decide how lenient we want scalars to be.


workers/src/tools/_loose_array.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (trimmed.startsWith("[") || trimmed.startsWith("{")) {
+    try {
+      const parsed: unknown = JSON.parse(trimmed);
+      return Array.isArray(parsed) ? parsed : [parsed];
+    } catch {
+      // Looks like JSON but isn't. Don't guess — let the schema report it.
+      return value;
+    }
+  }
+  if (trimmed.includes(",")) {
```
robertabbott: Does the comma split need to be skipped for `urls`? `ContentsRequest.shape.url` is a bare `z.string()` (`generated/schemas.ts:117`) with no URL format check, so `urls: "https://en.wikipedia.org/wiki/Washington,_D.C."` splits into `["https://en.wikipedia.org/wiki/Washington", "_D.C."]`, both halves pass `.min(1)`, and the handler fans out two independently BILLED subrequests (`tako_contents.ts:493`) — handing the model a different city's page as the payload. The comment at line 33 justifies the split with a "not a valid URL" error that nothing in this path can raise, and the unwrapped legacy `url` field accepts the same string intact (`tako_contents.ts:482` makes `urls` win). `node_ids`, whose items are also plain `z.string()`, is the same class.

workers/src/tools/_loose_array.ts
```diff
@@ -0,0 +1,75 @@
...
+ * blanket type error on the whole field.
+ */
+export function coerceStringToArray(value: unknown): unknown {
+  if (typeof value !== "string") return value;
+  const trimmed = value.trim();
+  if (trimmed === "") return value;
+  if (trimmed.startsWith("[") || trimmed.startsWith("{")) {
+    try {
+      const parsed: unknown = JSON.parse(trimmed);
+      return Array.isArray(parsed) ? parsed : [parsed];
```
robertabbott: Should the `[parsed]` object wrap be limited to object-item arrays like `components`? For the six primitive-item fields it can never validate, and it degrades the very error the model is meant to self-correct from: `sources: '{"data":{},"web":{}}'` — a plausible guess, since the descriptions discuss the nested per-source wire shape — used to report `sources: expected array, received string`, which names the actual mistake. It now reports an invalid enum value at `sources[0]`, inviting the model to fix element 0 of an array it never sent.

workers/src/tools/_loose_array.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  const arrayFields = (tool: (typeof TOOL_REGISTRY)[number]): string[] => {
+    const published = z.toJSONSchema(z.object(tool.inputSchema.shape), {
+      io: "input",
+    }) as { properties?: Record<string, { type?: unknown }> };
+    return Object.entries(published.properties ?? {})
+      .filter(([, schema]) => schema.type === "array")
+      .map(([name]) => name);
+  };
+
+  for (const tool of TOOL_REGISTRY) {
```
robertabbott: Should this assert that the guard generated any tests at all? If `arrayFields` returns `[]` for every tool, the block registers zero `it()`s and the suite stays green with the coverage claim void — and two things make that reachable: the `as { type?: unknown }` cast on line 91 discards zod's literal-union type for `type`, so a typo in `=== "array"` is no longer a compile error, and `published.properties ?? {}` yields `{}` rather than failing if the root ever moves behind a `$ref`. An `expect(count).toBe(8)` outside the loop would make it fail closed.

workers/src/tools/_loose_array.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Coverage guard. Wrapping is per-field and therefore forgettable: a new tool
+ * (or a new field) that publishes `type: "array"` without `looseArray` brings
+ * the -32602 back for exactly the hosts this exists for. This walks the real
+ * registry, finds every array-typed input field from the PUBLISHED JSON Schema
+ * (the same `z.toJSONSchema(…, { io: "input" })` the MCP SDK publishes — no
+ * zod internals), and fails naming the field.
+ */
+describe("every array-typed tool input accepts a stringified array", () => {
```
robertabbott: Does "every array-typed input field" hold, when `arrayFields` reads only top-level `properties`? Nested arrays are invisible: `components[].config` requires `datasets` (`tako_visualize.ts:155`), `columns`/`rows` (`:186,188`), `items` (`:194`) and `datasets[].data` (`:95,109`), none wrapped. A future `.nullable()` or union array is skipped silently too — zod's `nullableProcessor` emits `anyOf` with no top-level `type`, so the filter is false and no test is generated for it. Narrow the claim, or wrap at the `mcp.ts:651` shape chokepoint instead of per field?

workers/src/tools/_loose_array.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      io: "input",
+    }) as { properties?: Record<string, { type?: unknown }> };
+    return Object.entries(published.properties ?? {})
+      .filter(([, schema]) => schema.type === "array")
+      .map(([name]) => name);
+  };
+
+  for (const tool of TOOL_REGISTRY) {
+    for (const field of arrayFields(tool)) {
+      it(`${tool.name}.${field}`, () => {
```
robertabbott: Is there a test that a stringified `components` payload actually validates? This assertion proves only that no top-level type complaint fired, not that coercion produced a usable value: `'["x"]'` fails the discriminated union with `invalid_type` at path `["components", 0]`, which the `path.length === 1` filter on line 109 drops — so the guard passes on a parse that failed. `components` is the only wrapped field where coercion must reshape into a 20-member union, and neither its JSON-array-text nor its single-object form is exercised anywhere.

workers/src/tools/_loose_array.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const inner = z
+      .array(z.enum(["data", "web", "tako"]))
+      .min(1)
+      .default(["data", "web"])
+      .describe("Source(s) to ground in.");
+    const strict = z.toJSONSchema(z.object({ sources: inner }), { io: "input" });
+    const loose = z.toJSONSchema(z.object({ sources: looseArray(inner) }), {
+      io: "input",
+    });
+    expect(loose).toEqual(strict);
```
robertabbott: Should this compare the shipped fields rather than a schema rebuilt inside the test? This is the load-bearing claim — that leniency never reaches the contract — yet it is asserted on a local `inner`, and the registry guard checks only `type === "array"`, so drift in `items`, `minItems`, `default`, `description` or `required` on the eight real fields is unguarded. It also covers only the `.default()` shape; `node_ids` and `urls` are `.optional()`, which is exactly where `required` would move if `looseArray` were ever refactored to a plain `z.pipe` (which inherits `optin` from the transform, not from the wrapped schema).

robertabbott (commented):
Inline findings below: one issue (the comma split on `urls`), the rest suggestions concentrated on the coverage guard.

Two claims I tried to break and could not, noted because they are the ones that would have been expensive to get wrong:

- The published-schema-unchanged claim holds at every serialization site — SDK publication via `toJsonSchemaCompat(…, { pipeStrategy: "input" })`, and both `io` modes in `gen-registry.ts` — including the `.optional()` fields (`node_ids`, `urls`) that the equality test does not cover. It works because `$ZodPreprocess` re-derives *both* `optin` and `optout` from `def.out` (`zod/v4/core/schemas.js:2026-2030`), overriding the `$ZodPipe` inheritance. That makes the wrap safe in zod 4.4.2 specifically — worth knowing on a future zod bump, and the reason for the comment on the equality test below.
- No checked-in artifact needs regeneration. `registry/server.json`, `registry/lhm.plugin.json` and `_registry.ts` are the only generated files carrying input JSON Schema, and `registry:check` covers all three.

One pre-existing defect found on the way, out of scope for this PR and not inline-able since the file is not in the diff: `gen-registry.ts:565` omits `io: "input"`, so `z.toJSONSchema` defaults to `"output"` and computes `required` from `optout` — publishing every defaulted field as required to LobeHub. `registry/lhm.plugin.json:88` lists `sources`, `include_contents`, `preview_rows`, `country_code`, `locale` and `strict` in `required`, while `tako_contents` omits `url`. `registry:check` only diffs the generator against its own output, so it cannot see this. Happy to file it separately.

workers/src/mcp.test.ts
```diff
@@ -1347,3 +1347,113 @@ describe("SERVER_INSTRUCTIONS", () => {
...
+  // The contract does not move: a value the array schema rejects is still
+  // rejected, coerced or not.
+  it("still rejects an unknown source name", async () => {
+    const result = await callTool("tako_answer", {
+      query: "x",
+      sources: "bing",
+    });
+
+    expect(result.isError).toBe(true);
+    expect(result.content[0]?.text).toContain("-32602");
```
robertabbott: Does asserting only on `-32602` pin this test to the right failure? `InvalidParams` is also what the SDK returns for an unknown tool name and for output-validation failures, and `McpError` prefixes all of them with `MCP error -32602` — so a typo in the tool name on line 1452 would still pass. The assertion never names `sources` or `bing`. Related: this test alone leaves `fetch` unstubbed, so if coercion ever did widen the enum it would issue a live request to `staging.trytako.com` before failing. Assert the message text and zero fetch calls?

workers/src/tools/tako_answer.ts
```diff
@@ -66,11 +67,15 @@ const inputSchema = z.object({
     .string()
     .min(1)
     .describe('Natural-language question to answer (e.g. "What was US GDP in 2024?"). Website-traffic data is keyed by domain — ask about "openai.com monthly visits", not "OpenAI website visits".'),
-  sources: z
-    .array(z.enum(["data", "web", "tako"]))
-    .min(1)
-    .default(["data", "web"])
-    .describe('Source(s) to ground in. Default ["data","web"] (both) — keep BOTH enabled unless you have a confirmed reason to narrow. Narrow to ["data"] only once `tako_available_data` has confirmed the proprietary data exists (web is the fallback when it does not). Narrow to ["web"] only for content a data graph cannot hold (news articles, page text, qualitative claims) — never because a metric merely feels web-native: website traffic, app usage, and similar digital metrics ARE in the proprietary data graph. ("tako" is a legacy synonym for "data".)'),
+  // looseArray: hosts that stringify the array they meant to send (observed
+  // from OpenBB Copilot) get it coerced instead of a -32602. See _loose_array.ts.
```
robertabbott: Is there a reason not to log when a coercion fires? `mcp.ts:1682-1733` was added in `bdec340` specifically so argument rejections surface in Workers Logs "instead of only as user complaints" — coercing silently removes that signal for exactly the class this PR targets. After deploy there is then no way to tell whether real OpenBB traffic recovered, or which hosts send which shape, beyond the manual prod probe in the test plan. A `console.warn` naming the tool and field would also size the sibling scalar gap before you build it.

workers/src/tools/_loose_array.ts
25eliu: Fixed, and confirmed exactly as described before changing anything:

```
urls: "https://en.wikipedia.org/wiki/Washington,_D.C."
  → ["https://en.wikipedia.org/wiki/Washington", "_D.C."]
```

Both halves pass, `targets.map(fetchOne)` bills two subrequests, and nothing downstream can detect it. You are right that the justifying comment described an error that path cannot raise — there is no url format check to raise it.

Comma splitting is now opt-in (`commaSeparated`) and enabled only on `sources`. The rule I settled on: split only where the item domain provably excludes commas, which a closed enum does and an opaque string does not. So `urls` and `node_ids` no longer split; a bare string still wraps intact, which is the form that actually shows up.

Cost of that, stated plainly: `node_ids: "n1,n2"` now becomes `["n1,n2"]` and fails as one bad id rather than resolving two. That is the right trade on a billed fan-out — "usually right" is what produced this bug.

Guarded on the shipped field, not a rebuilt one: `tako_contents.urls` with that wikipedia URL must parse to exactly one entry. Re-enabling `commaSeparated` there fails that test (mutation-tested).

workers/src/tools/_loose_array.ts
25eliu: Fixed. Reproduced the degradation first: `sources: '{"data":{},"web":{}}'` reported `invalid_value` at `["sources", 0]` where it used to report `invalid_type` at `["sources"]`, so the model was being pointed at element 0 of an array it never sent.

`jsonObjectAsItem` is now opt-in, enabled only on `tako_visualize.components`. A JSON object reaching any other field passes through untouched and gets the plain `expected array, received string` back.

One thing your example surfaced that I would have missed: with comma splitting still on for `sources`, that same payload would have fallen through to the split and produced two bogus per-item errors — worse than either. JSON-shaped input is now parsed or passed through, never comma-split, and there is a test for that specific string.

workers/src/tools/_loose_array.test.ts
25eliu: Added, as an exact-set assertion rather than a count:

```
expect(found.map(f => `${f.tool.name}.${f.field}`).sort()).toEqual([
  "tako_agent.sources", "tako_agent_start.sources", "tako_answer.node_ids",
  "tako_answer.sources", "tako_contents.urls", "tako_search.node_ids",
  "tako_search.sources", "tako_visualize.components",
])
```

The set costs nothing over `toBe(8)` and says which field went missing rather than that one did. Bumping it is now a deliberate edit in the same commit that adds the field.

Also closed the detection hole you pointed at in the next comment, since it is what would make this assertion silently satisfiable.

workers/src/tools/_loose_array.test.ts
25eliu: Narrowed the claim, and fixed the `anyOf` half.

Confirmed your zod point: `z.array(z.string()).nullable().optional()` publishes `{"anyOf":[{"type":"array",…},{"type":"null"}]}` with no top-level `type`, so the filter now also matches array branches inside `anyOf`/`oneOf`. Without that, the exact-set assertion added above would have been silently satisfiable by a field that simply stopped being detected.

Not moving to the `mcp.ts:651` chokepoint in this PR, deliberately. It needs schema-walking introspection and would apply coercion to every input on every tool, and the nested case it would buy is not an observed shape: in the failure we actually have, the whole `components` array arrives as one JSON string, so parsing it yields correctly typed nested values. A host that stringifies only `components[0].config.datasets` is hypothetical.

So the scope is now written down instead of implied — in the module docstring, in the describe name (`registry guard: top-level array inputs`), and with all five unwrapped nested fields named (`components[].config.datasets`, `.columns`, `.rows`, `.items`, `datasets[].data`) plus the reason. If that hypothetical shows up, the chokepoint is the right answer and the note says so.

workers/src/tools/_loose_array.test.ts
25eliu: Right, the guard was passing on a failing parse. `'["x"]'` fails with `invalid_type` at `["components", 0]`, which `path.length === 1` dropped.

Two changes:

- The generic per-field assertion now requires `result.success` against a valid JSON-text payload per field, so a per-item failure cannot hide.
- `components` gets its own block exercising both real forms — the JSON-array text and the single-object text — each asserting a successful parse and one resulting item, plus a case proving `'["x"]'` still fails.

For the record both forms already worked; the coverage claim for them was the thing that was hollow.

workers/src/tools/_loose_array.test.ts
25eliu: Rebuilt this against the shipped fields. `looseArray` now records `wrapper → inner` in a WeakMap, and the guard asserts, for each of the 8 real fields:

```
z.toJSONSchema(z.object({ [field]: wrapped }), { io })
  === z.toJSONSchema(z.object({ [field]: inner }), { io })
```

for `io` of both `"input"` and `"output"`, so `gen-registry.ts`'s mode is covered too, and `required` is part of the compared object — which is what makes the `.optional()` fields (`node_ids`, `urls`) real coverage rather than assumed.

The WeakMap is also how it avoids poking zod internals to find the inner schema. Your `$ZodPreprocess` note (re-deriving both `optin` and `optout` from `def.out`, overriding the `$ZodPipe` inheritance) is now in the docstring as the reason this test exists, with the 4.4.2 pin called out, since a zod bump is the realistic way it breaks.

Mutation-tested: swapping the preprocess for `z.union([schema, z.string().transform(...)])` fails the equality on the shipped fields in both modes.

workers/src/mcp.test.ts
25eliu: Both fixed. Confirmed the code is overloaded — `ErrorCode.InvalidParams` covers `Tool X not found`, `Tool X disabled`, and both output-validation branches in `server/mcp.js`, so the old assertion would have passed on a typo'd tool name.

It now asserts the message names `tako_answer`, carries `invalid_value`, names the `"sources"` path and matches the option list `"data", "web", "tako"`. Note zod reports the allowed values rather than the received one, so there is no `bing` in the message to assert on — the option list is the specific part.

The fetch point was the better catch: that test was the only one leaving `fetch` unstubbed. It now stubs an empty sequence and asserts `not.toHaveBeenCalled()`, so a widened enum fails locally instead of billing a live `staging.trytako.com` request.

workers/src/tools/tako_answer.ts
25eliu: Added. Every coercion now emits, from `_loose_array.ts`:

```
[mcp] input coerced field=tako_answer.sources from=json-array items=2
```

`from` is one of `json-array | json-object | comma-list | bare-value`, which is what actually answers "did OpenBB traffic recover, and which hosts send which shape". A value that needs no reshaping logs nothing, so the line count is the incidence rate.

Two deliberate choices:

- The form and item count are logged, never the value. A coerced field can hold user URLs or query terms, and neither is needed to size the problem.
- `field` is a required option on `looseArray` rather than derived, because a zod preprocess cannot see the tool it was registered under. To stop that from rotting, the guard asserts each label's suffix matches its real property key and its prefix is a registered tool name.

It lives in `_loose_array.ts` rather than `_log.ts` to keep the module self-contained. And yes — sizing the sibling scalar gap off this log is exactly the plan before building it.

robertabbott (approved):

PR #212, opened 2026-08-04
feat(oauth): managed-OAuth support for Foundry — non-expiring partner client_ids and offline_access

25eliu:
## Why

Microsoft's Foundry partner onboarding asks us for a **managed OAuth app**: one `client_id` that means "Azure AI Foundry", which they embed in their own configuration and use for every customer they onboard, so no developer has to register their own OAuth app to use Tako.

We currently can't safely give them one. `mcp.tako.com` is a pure Dynamic Client Registration server, and every registration expires after `REGISTRATION_TTL_S` (1 year, `handlers.ts:101`). That TTL is correct for a consumer host — Claude.ai and ChatGPT silently re-register on first use after expiry — and wrong for a catalog partner, where the value is pasted into config once and never touched again. Handing Microsoft a 1-year client means Foundry auth breaks for every Tako customer simultaneously, 12 months later, with no warning and nobody on either side looking at it.

## What changed

An authenticated partner path on `/register`, gated by a new optional secret `OAUTH_PARTNER_REGISTRATION_TOKEN`, presented as the `X-Tako-Partner-Token` header. Clients minted through it carry **no `exp`** — `isExpired` already treats an absent `exp` as never-expiring (`jwt.ts:94`) — plus a `partner: true` claim so a decoded `client_id` is self-describing in logs and during an incident.

**Why a dedicated header instead of `Authorization`:** hosts routinely retry `/register` with a stale access token still attached. Overloading `Authorization` would force a choice between rejecting those (breaking ordinary DCR for real clients) and ignoring a mistyped partner secret (silently minting a 1-year client for a partner — the exact failure this path exists to prevent). Presence of the dedicated header unambiguously means "mint a partner client", so any bad value is a loud 401, including the case where no partner token is configured for the environment. Comparison is constant-time.

**Also:** `client_id_expires_at` is now reported on every registration. RFC 7591 §3.2.1 reads an omitted field as "never expires", which actively misdescribed our 1-year public path. On the partner path the `0` is also the operator's confirmation that the gate engaged rather than quietly falling back to public DCR.

The public path is otherwise untouched: same TTL, same redirect-URI validation, stray `Authorization` headers still ignored.

## Second commit: `offline_access`

Found while researching Foundry's actual requirements, and it would have broken the integration on day one.

Foundry's [setup guidance](https://learn.microsoft.com/en-us/azure/foundry/agents/how-to/mcp-authentication) tells operators to include `offline_access` for automatic token refresh, and its troubleshooting table names a *missing* `offline_access` as the fix for "Your session has expired". Our `SUPPORTED_SCOPES` held only `mcp` and `/authorize` validates fail-closed — so Foundry's own recommended configuration got a `400` at setup time, before anyone reached a tool call. Confirmed with a failing test before the fix.

`offline_access` grants nothing here: we already issue a refresh token on every grant, which is exactly what the scope asks for. The resource server never needed a change — `access.ts` checks `scopes.includes("mcp")`, so `mcp offline_access` already passed there.

Requesting `offline_access` *alone* is now rejected at `/authorize` rather than minting a token that `/mcp` would refuse with `insufficient_scope` on the first tool call, which would surface the mistake far from its cause.

## Deliberately not changed

`ACCESS_TOKEN_TTL_S` stays at 15 minutes. There are open reports of Foundry mishandling OAuth refresh against custom MCP servers ([redirectUrl: null when a Refresh URL is set](https://learn.microsoft.com/en-us/answers/questions/5806486/azure-ai-foundry-custom-mcp-tool-with-oauth-identi), [no refresh token sent, 401 after access-token expiry](https://techcommunity.microsoft.com/discussions/azure-ai-foundry-discussions/new-ai-foundry-not-sending-refresh-tokens-to-mcp-401-after-access-token-expirati/4515123)). If those bite us, agents will start 401ing ~15 minutes into a session. But that's a bug on their side and loosening our token lifetime to paper over it is the wrong first move — asking Microsoft directly during onboarding is. Flagged in the runbook so it's the first thing checked if Foundry sessions die early.

## Sharp edges, documented in `docs/partner-oauth-clients.md`

- **Adding a redirect URI later mints a new `client_id`.** The URIs are baked into the signed JWT and `/authorize` matches by exact string equality — there's no server-side record to amend. So we must collect Microsoft's complete redirect-URI list up front, staging and per-region callbacks included.
- **Rotating `OAUTH_PARTNER_REGISTRATION_TOKEN` revokes nothing.** It's consulted only at mint time.
- **`OAUTH_SIGN_KEY` rotation is the only revocation**, and it takes every public registration with it.

## Test plan

- [x] 12 new tests in `handlers.test.ts` — non-expiring mint, still-valid-at-two-years (clock advanced past the public TTL), wrong token → 401, wrong-length token → 401, header with no token configured → 401, redirect-URI validation still enforced on the partner path, partner client accepted end-to-end by `/authorize`, unregistered redirect_uri still rejected, `client_id_expires_at` correct on both paths, stray `Authorization` header ignored.
- [x] Written test-first; confirmed all 5 new behavioral tests failed before implementation.
- [x] 3 further tests for `offline_access`: accepted alongside `mcp`, rejected alone, advertised in discovery. Both behavioral ones confirmed failing first.
- [x] `npm test` — 950 + 85 + 46 = **1081 passed**, 0 failed (all three vitest configs).
- [x] `npm run typecheck` — clean across all four tsconfig projects.

## Deploy note

`OAUTH_PARTNER_REGISTRATION_TOKEN` must be set before a partner client can be minted:

```bash
openssl rand -base64 32 | npx wrangler secret put OAUTH_PARTNER_REGISTRATION_TOKEN --env production
```

Unset, `/register` behaves exactly as it does today — the feature is inert.


workers/src/oauth/handlers.ts
```diff
@@ -386,6 +415,26 @@ export async function handleRegister(
...
+  // — including the case where no partner token is configured for this
+  // environment, which would otherwise quietly hand back a 1-year client.
+  const partnerHeader = req.headers.get(PARTNER_TOKEN_HEADER);
+  const isPartner = partnerHeader !== null;
+  if (isPartner) {
+    const expected = env.OAUTH_PARTNER_REGISTRATION_TOKEN;
+    if (
+      typeof expected !== "string" ||
+      expected.length === 0 ||
+      !constantTimeEqual(partnerHeader, expected)
```
robertabbott: Should both sides be trimmed before comparison? The runbook's own mint command — `openssl rand -base64 32 | npx wrangler secret put` (`docs/partner-oauth-clients.md:22`) — stores a trailing newline, and HTTP strips trailing whitespace from header values, so the stored secret is one byte longer than anything a client can present and the length check in `constantTimeEqual` returns false permanently. This repo already paid for that lesson: `freetier.ts:104` trims `FREE_TIER_API_KEY` because a "piped `wrangler secret put` newline ... would otherwise 401 every anonymous request downstream — indistinguishable from a wrong key", and `workers/README.md:28` records it in the binding table. Is the partner path different?

workers/src/oauth/handlers.ts
```diff
@@ -386,6 +415,26 @@ export async function handleRegister(
...
+  // non-expiring client, so anything other than an exact match fails loud
+  // — including the case where no partner token is configured for this
+  // environment, which would otherwise quietly hand back a 1-year client.
+  const partnerHeader = req.headers.get(PARTNER_TOKEN_HEADER);
+  const isPartner = partnerHeader !== null;
+  if (isPartner) {
+    const expected = env.OAUTH_PARTNER_REGISTRATION_TOKEN;
+    if (
+      typeof expected !== "string" ||
+      expected.length === 0 ||
```
robertabbott: Should this require a minimum length rather than only rejecting the empty string? Set to something short in staging — `test`, say — the partner path is live and a 4-character guess mints a `client_id` that never expires, against an endpoint with no app-level rate limit (`handlers.ts:334`) and no log line on failure. Failing closed below ~32 bytes would make a weak value behave as "not configured" instead of as a live oracle. Relatedly: is the `expected.length === 0` arm reachable from a test? `envWith()` only ever leaves the secret `undefined`, so today only the `typeof` arm is covered.

workers/src/oauth/handlers.test.ts
```diff
@@ -372,6 +372,224 @@ describe("/register (DCR)", () => {
...
+      const twoYearsMs = 2 * 365 * 24 * 60 * 60 * 1000;
+      Date.now = () => realNow() + twoYearsMs;
+      const claims = await verifyJwt<ClientIdClaims>(client_id, SIGN_KEY);
+      expect(claims).not.toBeNull();
+    } finally {
+      Date.now = realNow;
+    }
+  });
+
+  it("rejects a wrong partner token with 401 rather than silently downgrading", async () => {
```
robertabbott, line 506: The XOR loop in `constantTimeEqual` is never executed with a mismatching byte — could one negative case use an equal-length wrong token? `PARTNER_TOKEN` is `"partner-secret-value"` (20 chars) and the two rejection tests pass `"wrong-token"` (11) and `"short"` (5), so both return early at `if (ab.length !== bb.length)`. Replacing the loop body with `diff |= 0` keeps all 12 partner tests green — i.e. any 20-character value would mint a non-expiring partner `client_id`. This is the same shape as the "guards that passed either way" round in 27283f1, which is why it seemed worth naming rather than assuming.

workers/src/oauth/handlers.ts
```diff
@@ -386,6 +415,26 @@ export async function handleRegister(
...
+      !constantTimeEqual(partnerHeader, expected)
+    ) {
+      return jsonError(
+        "invalid_client",
+        "invalid partner registration token",
```
robertabbott, line 482: Should this 401 and the successful mint at `:508` each emit a log line? Neither does today, and `/register` has no app-level rate limit by design (`:334`), so someone probing the partner header leaves no trace at all. It also leaves no record of what was minted: `types.ts:46` documents the `partner: true` claim as existing so a decoded `client_id` is "self-describing in logs and in an incident", but no code path ever writes one, so "which partner client_ids exist, minted when, for which redirect URIs" has no answer six months out. This worker logs comparable events deliberately — `access.ts:44` on every `/mcp` reject, `handlers.ts:1548` on login rate-limiting because "the status code carries no signal for abuse monitoring". A line naming the branch taken (never the value) would also separate "wrong `--env`" from "mistyped secret" during the one-time mint, which the deliberately opaque 401 body cannot.

workers/src/oauth/handlers.ts
```diff
@@ -450,14 +499,22 @@ export async function handleRegister(
...
   const client_id = await signJwt(claims, cfg.signKey);
 
   return Response.json(
     {
       client_id,
       client_id_issued_at: claims.iat,
+      // RFC 7591 §3.2.1: 0 means "never expires". Omitting the field means
+      // the same thing, which would misreport the 1-year public path — so
+      // state it either way rather than leaving clients to guess.
+      client_id_expires_at: claims.exp ?? 0,
```
robertabbott, line 585: RFC 7591 §3.2.1 does not define `client_id_expires_at`. It defines `client_id`, `client_secret`, `client_id_issued_at`, and `client_secret_expires_at`, and the "or 0 if it will not expire" wording belongs to `client_secret_expires_at`. The name is also absent from OIDC Registration 1.0 §3.2 and from the IANA DCR metadata registry. So an omitted field never meant "never expires" — it meant nothing — and no client is obliged to read `0` that way. Reporting the value still looks right, and it is a good engaged-or-not signal for the operator. Should this comment and `docs/partner-oauth-clients.md:41`, which calls `0` "the 'never expires' signal from RFC 7591 §3.2.1", present it as a Tako convention instead of as spec?

docs/partner-oauth-clients.md
```diff
@@ -0,0 +1,91 @@
...
+```bash
+openssl rand -base64 32 | npx wrangler secret put OAUTH_PARTNER_REGISTRATION_TOKEN --env production
+```
+
+Then mint, passing **every** redirect URI the partner will ever use:
+
+```bash
+curl -sX POST https://mcp.tako.com/register \
+  -H 'content-type: application/json' \
+  -H "x-tako-partner-token: $OAUTH_PARTNER_REGISTRATION_TOKEN" \
```
robertabbott: The runbook cannot succeed as written: line 22 pipes `openssl rand` straight into `wrangler secret put`, so the value is never captured, and this line then expands `$OAUTH_PARTNER_REGISTRATION_TOKEN` from a shell that never had it. An empty value is curl's documented header-*removal* form, so the request arrives carrying no partner header at all — `isPartner` is false and the operator silently receives a 1-year public `client_id`, which is the exact outcome the gate comment at `handlers.ts:418` exists to prevent. The `client_id_expires_at: 0` check below would catch it, but only if they read that far. Should this capture first, e.g. `TOKEN=$(openssl rand -base64 32)` then pipe `printf %s "$TOKEN"` into `wrangler`?

docs/partner-oauth-clients.md
```diff
@@ -0,0 +1,91 @@
...
+configuration time, before anyone reached a tool call.
+
+`offline_access` on its own is rejected at `/authorize`: it would mint a token
+that `/mcp` later refuses with `insufficient_scope`, surfacing the mistake at a
+place that cannot explain it.
+
+## Operational cautions
+
+- **`OAUTH_SIGN_KEY` rotation is the only revocation mechanism**, and it
+  invalidates every public registration too. Treat it as a break-glass action
```
robertabbott: Rotating `OAUTH_SIGN_KEY` invalidates considerably more than "every public registration" — the same key signs auth codes (`handlers.ts:814`), access tokens (`:1218`), refresh tokens (`:1219`), state cookies (`:677`) and session cookies (`:1875`). Rotating it signs out every user and voids every live 14-day refresh token on top of dropping every registration. Should that be stated here? A responder reading only this line during an incident would size the break-glass action as "partners re-register" rather than "everyone is logged out at once".

workers/src/oauth/handlers.test.ts
```diff
@@ -1682,6 +1900,65 @@ describe("/authorize hardening", () => {
     expect(res.status).toBe(200);
   });
 
+  it("accepts `mcp offline_access` (Azure AI Foundry's documented scope)", async () => {
```
robertabbott, line 2012: This asserts only `status === 200` on GET `/authorize`, so nothing proves `mcp offline_access` survives into an issued token — worth a `/token` round-trip case? The sibling empty-scope test at `:1962` makes exactly that assertion (`expect(body.scope).toBe("mcp")`). `scope: parsed.scope ?? "mcp"` (`handlers.ts:806`) carries the raw string into the auth code and `access.ts:123-127` splits it and checks `includes("mcp")`, but every token in `access.test.ts` uses a single-valued `scope: "mcp"`. Refactoring that check back to `claims.scope !== REQUIRED_SCOPE` would keep the whole suite green while returning `insufficient_scope` on the first tool call for every token issued through Foundry's documented scope — the failure this commit exists to pre-empt.

workers/src/oauth/handlers.ts
```diff
@@ -386,6 +415,26 @@ export async function handleRegister(
...
+  const partnerHeader = req.headers.get(PARTNER_TOKEN_HEADER);
+  const isPartner = partnerHeader !== null;
+  if (isPartner) {
+    const expected = env.OAUTH_PARTNER_REGISTRATION_TOKEN;
+    if (
+      typeof expected !== "string" ||
+      expected.length === 0 ||
+      !constantTimeEqual(partnerHeader, expected)
+    ) {
+      return jsonError(
```
robertabbott, line 480: Is the off-spec error shape here deliberate? RFC 7591 §3.2.2 defines exactly four registration error codes — `invalid_redirect_uri`, `invalid_client_metadata`, `invalid_software_statement`, `unapproved_software_statement` — and says to return 400 unless otherwise specified, so `invalid_client` at 401 is neither, while every other failure in this handler already uses a legal code (`:470`, `:477`). It is also the only 401 in the worker without a `WWW-Authenticate` challenge (`mcp.ts:1772`, `:1808` send one via `wwwAuthenticate`), and `jsonError` has no parameter to add one. If the reasoning is that only an operator with curl ever reaches this path, that is a fine answer — worth a line in the comment above?

docs/partner-oauth-clients.md
```diff
@@ -0,0 +1,91 @@
+# Partner OAuth clients (managed OAuth)
```
robertabbott, line 1: Is this doc reachable from anywhere an operator would actually look? Its only inbound reference is the TSDoc block in `env.ts:147`. Every other operator-facing worker secret has a `workers/README.md` section — `FREE_TIER_API_KEY` gets a binding table, a numbered rollout checklist, a kill switch and a rotation note (`README:148`), `LOGIN_*` gets one at `:251` — and neither `workers/README.md` nor `AGENTS.md` mentions this secret or this file. A pointer from the README would save the next responder a grep through `env.ts`.

robertabbott (commented):
Reviewed the partner gate, the scope change, the docs, and the tests. The design reasoning holds up — the dedicated header over `Authorization`, the absent-`exp` mechanism, and the `must include mcp` rule all check out, and I confirmed nothing else in the worker relies on `client_id` expiry as a safety net.

The substantive findings are inline. The one I would fix before minting anything is the missing `.trim()`: as documented, the runbook cannot produce a working partner client. Two of the findings are on the runbook itself rather than the code.

workers/src/oauth/handlers.ts
25eliu: Fixed — you were right, and this was a blocker: the partner path could not have worked in production at all.

Both sides are now trimmed via a `partnerSecret(env)` helper, with the TSDoc pointing at `resolveFreeTierConfig` as the precedent. Added a unit test that configures the secret with a trailing `\n` and presents it without.

Also verified outside the test suite, since every unit test passes a clean in-memory string and would not have caught this either. Ran `wrangler dev` with `OAUTH_PARTNER_REGISTRATION_TOKEN` stored with an embedded trailing newline and curled `/register` with the header carrying no newline:

```
client_id_expires_at: 0
decoded claims: {"type":"client_id","client_name":"Microsoft Foundry",
                 "redirect_uris":["https://foundry.example.com/cb"],
                 "iat":1785872692,"partner":true}
```

No, the partner path was not different. Same failure, same fix.

workers/src/oauth/handlers.ts
25eliu: Done. `partnerSecret` now returns `null` below `PARTNER_TOKEN_MIN_LEN` (32 chars, after trimming), so a weak value behaves as "not configured" rather than as a live oracle. 32 sits comfortably under the 44 that the runbook's `openssl rand -base64 32` produces, so a correctly minted secret is never caught by it.

On coverage: correct, the `length === 0` arm was unreachable from any test. There are now three cases — empty string, `"test"` (too short), and whitespace-only — all asserting the path stays closed.

This pairs with the logging on the sibling thread: failing closed is only diagnosable if the operator can tell "below the floor" from "mistyped", which the 400 body deliberately will not say.

workers/src/oauth/handlers.test.ts
25eliu: Caught a real hole. Both rejection cases used wrong-*length* tokens (11 and 5 against 20), so both returned early at the length guard and the comparison loop never executed.

Added two cases: an equal-length wrong token, and one differing only in its final byte. Confirmed by running your exact mutation — replacing the loop body with `diff |= 0`:

```
× rejects an equal-length wrong token (exercises the XOR compare)
× differs from the real token in only its final byte and is rejected
  Tests  2 failed | 104 passed (106)
```

Before this commit that mutation was green. Also bumped `PARTNER_TOKEN` in the tests to 37 chars so it clears the new minimum-length floor.

workers/src/oauth/handlers.ts
25eliu: Added. Both outcomes now log one greppable line, never the value:

```
[oauth] /register minted partner client_id name=Microsoft Foundry redirect_uris=https://foundry.example.com/cb
[oauth] /register partner token rejected (configured=yes)
```

Both captured from a live `wrangler dev` run. The `configured=` flag is what separates "wrong `--env`" from "mistyped secret" during the one-time mint, which matters more now that a sub-32-char secret also fails closed.

You were right that `types.ts:46` was writing a cheque no code path cashed. It does now.

workers/src/oauth/handlers.ts
25eliu: You are right and I checked the RFC rather than taking it on trust. §3.2.1 defines `client_id`, `client_secret`, `client_id_issued_at` and `client_secret_expires_at` only; `client_id_expires_at` appears nowhere in RFC 7591, and the "or 0 if it will not expire" wording belongs to `client_secret_expires_at`. I cited a spec for something it does not say, in three places.

Kept the field, relabelled it as a Tako convention in all three: the code comment now spells out which fields §3.2.1 actually defines and that this one echoes `client_secret_expires_at`'s shape; the two test comments; and the runbook, which now says "a Tako convention, not a spec field" and leans on it as the engaged-or-not signal you identified rather than as an obligation on clients.

workers/src/oauth/handlers.ts
25eliu: Deliberate is the wrong word for it — it was not considered. Went with the spec shape instead of documenting the deviation: registration failures on this path now return **400 `invalid_client_metadata`**.

That also disposes of the `WWW-Authenticate` problem rather than commenting around it. Worth adding to your point: RFC 9110 §15.5.2 makes the challenge header mandatory on a 401, so the old shape was a protocol violation and not only an off-spec code — which made extending `jsonError` the only other honest option, for a path with one human caller.

The comment above the gate now records why the code is what it is. Tests updated to assert 400 + `invalid_client_metadata`, and confirmed live:

```
HTTP 400
{"error":"invalid_client_metadata","error_description":"invalid partner registration token"}
```

workers/src/oauth/handlers.test.ts
25eliu: Fixed, and your hypothetical was exactly right. Two additions:

- A `/token` round-trip in `handlers.test.ts` asserting `body.scope === "mcp offline_access"`, that the same value survives inside the decoded access token, and that a refresh token is issued.
- Three cases in `access.test.ts`: `mcp offline_access`, `offline_access mcp` (position independence), and `mcpx` rejected, so the split-on-whitespace behaviour is pinned rather than a substring match.

Confirmed with your mutation — narrowing the check to `claims.scope !== REQUIRED_SCOPE`:

```
× accepts a multi-valued scope containing 'mcp' (mcp offline_access)
× accepts 'mcp' in any position within a multi-valued scope
  Tests  2 failed | 122 passed (124)
```

Green before this commit. The regression the commit exists to pre-empt is now actually defended.

docs/partner-oauth-clients.md
25eliu: Fixed — and I reproduced the trap before fixing it, because the failure is silent. Curling `/register` with an unset variable in the header:

```
client_id_expires_at: 1817408709   # 365 days, not 0
```

An ordinary public client, no error, exactly as you described.

The runbook now captures first, using your suggested shape:

```bash
TOKEN=$(openssl rand -base64 32)
printf %s "$TOKEN" | npx wrangler secret put OAUTH_PARTNER_REGISTRATION_TOKEN --env production
```

`printf %s` over `echo` so the stored bytes match what was pasted, though the worker now trims anyway per the other thread. Also promoted the `client_id_expires_at: 0` check from a trailing sentence to its own bolded step, since as you note it only helps if they read that far — and it now says what a non-zero value means and that `wrangler tail` will show which of wrong-`--env` / mistyped / too-short it was.

docs/partner-oauth-clients.md
25eliu: Corrected. It now reads that one key signs auth codes, access tokens, refresh tokens, DCR client_ids, and the state and session cookies — so rotating it signs out every logged-in user, voids every live 14-day refresh token, and drops every registration simultaneously, with a browser sign-in needed for each human afterwards.

Your framing of the risk was the useful part: a responder reading the old line would have sized a full-outage action as a partner inconvenience. Also propagated to the `env.ts` TSDoc, which had the same understatement.

docs/partner-oauth-clients.md
25eliu: Added. `workers/README.md` now has a `## Partner OAuth clients (OAUTH_PARTNER_REGISTRATION_TOKEN)` section in the same shape as the `FREE_TIER_API_KEY` one: binding table with the trim and minimum-length notes, what the unset state does (feature inert — the correct state for any env without partner onboarding in flight), the mint-time-only nature of the credential, the real revocation cost, the `wrangler tail` grep for the new log lines, and a link out to the runbook.

`docs/partner-oauth-clients.md` now links back to the README section, and the `env.ts` TSDoc points at both.

robertabbott (approved):

PR #216, opened 2026-08-05
fix(available_data): ask the graph whether the entity actually has the metric

25eliu:
## The problem

`tako_available_data`'s lookup path (`metric` supplied) resolves its two halves with **two independent `graph/search` calls** and never asks whether they are connected. That was enough for `found: true` plus a pinned, priced `next_call`.

Measured on prod, that claim is wrong **23% of the time** — the emitted handle retrieves on 15 of 22 pairs. The failures all resolve a metric whose *name* fits perfectly, so no amount of string matching catches them:

```
Lockheed Martin / backlog   ->  pins `Backlog`   ->  0 cards
                                (Lockheed has `12 Month Backlog`, not `Backlog`)
JPMorgan Chase / NIM        ->  pins `Net Interest Margin`  ->  0 cards
```

Only the graph knows, and asking it is free.

The advertised schema was also lying about it: it claimed `found` meant *"at least one match has live data coverage — not mere node resolution"* while the lookup path set it `true` with `matches: []` and no check performed. The honest wording existed only on the internal schema, which never reaches the wire.

## The change

After the existing parallel resolution, **one** call:

```
GET /api/beta/graph/related?node_id=<entity>&relation=metrics&q=<filter>&limit=20
```

and a `verified` discriminator that reports what was *checked*, separately from `found`, which stays the outcome:

| value | meaning |
|---|---|
| `coverage` | a coverage list was drilled (discovery path) |
| `pair` | the metric is on the entity's own metric list |
| `unlinked` | the list was checked and holds nothing matching → **the emitted `next_call` drops the pin** |
| `resolution` | no pair evidence (skipped or failed) → behaves exactly as today |

`resolution` deliberately collapses probe-failed and probe-skipped: the caller's action is identical in both. A distinct value only earns its place when the recommended action differs.

## Results (prod, 24 pairs, `scripts/available-data-truth.ts`)

| | handles that land |
|---|---|
| today (always pinned) | **15/22 (68%)** |
| this PR | **17/22 (77%)** |

Two rescues — Lockheed/backlog `0 → 1`, JPMorgan/NIM `0 → 2` — and **zero regressions**: no case where dropping the pin lost cards.

## Two things measured and then deliberately removed

**A re-pin branch** that chose a node off the entity's list when rank 0 failed the name test. It fired 4 times and 2 were catastrophic, because `graph/related` is *not* a curated list of an entity's own metrics:

```
Netflix / "paid subscribers"  ->  Disney Core Paid Subscribers
Walmart / "total revenue"     ->  5G Telco Edge IoT Revenue Total Revenue
```

Both returned from the **queried entity's** metrics relation, both passing `confidentMatch`. (`Bombshells Same Store Sales` is on Starbucks'.) Every firing was where the tool correctly emits *no* handle today, so the branch converted "no handle" into a confidently wrong **priced** call. A separate arm confirmed it would not have helped anyway: pinning an entity-backed node recovers 1 of 11 dead handles.

**A second filter variant.** Changed the verdict on 0 of 24 pairs while doubling `graph/related` load. Now exactly one filter — the resolved node's own name when confident, the query's longest token otherwise.

## Cost

Base resolution ~200ms; the probe adds one `graph/related` (~0.4–1.5s observed, the graph API is variable), bounded by a 2.5s deadline after which the verdict degrades to `resolution`. It runs only when both an entity and a rank-0 metric resolved — with no node to check, the probe cannot supply one.

## What this does NOT fix

An edge is not a card. Five pairs come back `verified: pair` and still retrieve nothing (Carnival/passenger cruise days, Apple/P&E, Tesla/vehicle deliveries, Netflix/paid subscribers, Starbucks/same store sales). That is backend card coverage, not resolution, and no MCP-side change closes it.

## Test plan

- [x] 1065 + 86 + 46 tests pass across all three vitest configs
- [x] all four `tsc` projects clean; `schemas:check` and `registry:check` in sync
- [x] end-to-end on prod before and after the single-call simplification: identical 17/22
- [x] failure isolation: probe throws → output byte-identical to today (`verified: resolution`)
- [x] no-veto invariant: `unlinked` never forces `found: false` and never withholds the handle
- [x] regression guard: `reconcilePair` can never return a node (the Disney case)
- [ ] re-run `scripts/available-data-truth.ts` post-deploy against staging

## Tooling added

- `scripts/available-data-live.ts` — drive the real handler for one pair, print the markdown a model reads plus `structuredContent`. Free.
- `scripts/available-data-truth.ts` — 24 pairs: run the handler, execute its own `next_call`, report whether it retrieves. Priced (~$0.17).
- `scripts/pair-confirm-eval.ts` — bulk verdict/latency sweep. Carries a warning: it *reimplements* the resolution logic, and an earlier version omitted the exact-name promotion (`sameTokens`) and understated accuracy by 19 points. Never measure this tool by re-deriving it.


workers/src/tools/_pair_confirm.ts
```diff
@@ -0,0 +1,194 @@
...
+  if (globalMetric !== null && confidentMatch(metricQuery, globalMetric)) {
+    const linked = scoped.some((n) => sameNode(n, globalMetric));
+    return { verified: linked ? "pair" : "unlinked", entityMetricMatches };
+  }
+
+  // Rank 0 is absent or unvetted, and the entity's list holds nothing that
+  // vouches for it either. The caller still sees rank 0 as an unvetted
+  // candidate with its alternates; what this adds is that the entity's OWN
+  // list was checked too, which is firmer ground for reporting a gap.
+  return { verified: "unlinked", entityMetricMatches };
```
robertabbott: `reconcilePair`'s fallthrough returns `unlinked` without ever reading `scoped`. On the diagnose path (`rank0Confident === false`) `metricFilter` sends the query's longest token, so `scoped` is frequently non-empty — `_pair_confirm.test.ts:124` and `tako_available_data.test.ts:1203` both construct exactly that state, queueing `Operating costs and expenses` back from Pfizer's own list, and the verdict is still `unlinked`.

The advertised enum says `unlinked` means "the entity's list was checked and holds nothing matching", and `buildPairSummary` renders "Pfizer's own metric list was also checked for \"R&D expense\" and holds nothing matching" while two matching entries sit in `entityMetricMatches`. Nothing about rank 0's linkage was established on this path — should it return `resolution`, leaving `unlinked` to mean only what the schema advertises?

workers/src/tools/tako_available_data.ts
```diff
@@ -377,6 +390,53 @@ const tako_available_data = {
...
+        );
+        const parsed = relatedShape.safeParse(raw);
+        if (!parsed.success) {
+          // Same rule as the drill: a wire-guard failure degrades the caller's
+          // answer, so it is ALWAYS logged rather than silently suppressing the
+          // pair evidence.
+          logWireGuardFailure("tako_available_data", "pair-confirm", parsed.error, raw);
+          return null;
+        }
+        return parsed.data.relation?.items ?? [];
```
robertabbott: `pairProbe` returns `relation.items` and discards `relation.total` / `total_capped`, which `coverageProbe` reads off the same parsed shape 50 lines up.

With `PAIR_PROBE_LIMIT = 20` and a confident-path filter that is the resolved node's own name — often generic (`Revenues`, `Operating Income`) — a large entity's filtered metric list can exceed 20 entries, and a target paged out of view yields `unlinked`. That drops the pin and prints "the graph holds no edge … If it is also empty, Tako has no card for this pair: report the gap rather than rephrasing further."

Should `relation.total > items.length` (or `total_capped`) degrade to `resolution` instead of asserting absence?

workers/src/tools/_available_data.ts
```diff
@@ -520,7 +544,29 @@ export function buildPairSummary(input: {
...
-    return `Resolved the entity, but NO metric confidently matches "${metricQuery}" — the closest names Tako holds are shown below and are probably NOT what you asked for. Pick one deliberately (pin its node_id with strict:true), or conclude Tako does not track this measure.`;
+    // Two signals agreeing is a firmer basis for stopping than one. When the
+    // entity's OWN list was checked and came up empty, say so — that is the
+    // difference between "we could not vet the name" and "this entity does not
+    // have it", and only the second justifies reporting a gap.
+    const checked =
+      verified === "unlinked"
+        ? ` ${oneLine(pair.entity.name)}'s own metric list was also checked for "${metricQuery}" and holds nothing matching.`
+        : "";
+    return `Resolved the entity, but NO metric confidently matches "${metricQuery}" — the closest names Tako holds are shown below and are probably NOT what you asked for.${checked} Pick one deliberately (pin its node_id with strict:true), or conclude Tako does not track this measure.`;
```
robertabbott: `entityMetricMatches` reaches `buildPairSummary` on the diagnose path, but this branch only uses `checked` — the `near` clause that prints the names exists solely in the `verified === "unlinked"` branch below, which is unreachable when `metricConfident` is false.

`metricFilter`'s diagnose mode picks the query's longest token specifically because "no handle is emitted on this path, so the near-miss list is the whole payload and recall is what matters", and `_pair_confirm.test.ts:145` calls it "the genuinely useful half of an unlinked verdict". As written, Pfizer's real R&D metrics are fetched, sliced to 3, and dropped. Should this branch append `near` as well?

workers/src/tools/_pair_confirm.ts
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * Is this scoped node the same metric the global search resolved?
+ *
+ * Id equality is the intended test — `graph/search` and `graph/related` return
+ * nodes from the same graph. Name equality is a deliberate backstop: if those
+ * id spaces ever diverge, an id-only test would report `unlinked` for EVERY
+ * pair, which is a systematic false negative that would quietly unpin every
+ * handle the tool emits. Two cheap tests beat one silent failure mode.
+ */
+const sameNode = (a: GraphNode, b: GraphNode): boolean =>
```
robertabbott: `sameNode`'s name arm is documented as a backstop against id-space divergence, but it also fires on the near-duplicate metric twins this file cites as KE-812: identical name, different id. When it does, `verified: "pair"` is returned even though the id on the entity's list is not the id `next_call` pins, and the summary upgrades to "the strongest free confirmation available … Run the next_call below verbatim."

The PR notes 5 `pair` verdicts still retrieve nothing. Is a name-only match strong enough to carry the `pair` wording, or should id-match and name-only-match be separable so the twin case does not read as confirmation?

workers/src/tools/tako_available_data.ts
```diff
@@ -538,24 +598,62 @@ const tako_available_data = {
...
+      const entityRef = entities[0] ?? null;
+
+      // Probed only when there is a rank-0 metric to CHECK. With the re-pin
+      // branch gone the probe cannot supply a metric, so running it when the
+      // global search resolved none would buy nothing and cost a round trip —
+      // that case falls straight through to the coverage drill below.
+      let verified: PairVerdict = "resolution";
+      let entityMetricMatches: GraphNode[] = [];
+      if (entityRef !== null && rank0 !== null) {
+        const scoped = await pairProbe(
```
robertabbott, line 643: The probe is awaited serially after the four-way `Promise.all`, so it adds a full round trip to every lookup where an entity and a rank-0 metric resolved — the common case.

The comment at lines 518–528 rejects the entity coverage-promotion probe on precisely this ground: "a `graph/related` probe is ~0.7s and adds a second round trip … a 6.4x regression on every lookup. Not worth it here." A reader hitting that comment and then this code sees two opposite conclusions about the same cost. Should that comment be updated to say why the tradeoff comes out differently for correctness than for entity promotion?

workers/src/tools/tako_available_data.ts
```diff
@@ -377,6 +390,53 @@ const tako_available_data = {
       }
     };
 
+    // The PAIR-CONFIRMATION probe: does this entity actually hold this metric?
+    //
+    // `graph/related` scoped to the entity with a substring filter, one call
```
robertabbott: "one call per filter variant, in parallel" — the final commit reduced this to a single call and `metricFilter` returns exactly one filter. Worth rewording so the next reader does not go looking for the fan-out?

workers/src/tools/_pair_confirm.ts
25eliu: Fixed — that path now returns `resolution`.

You are right that nothing about linkage is established there: there is no pinned node to check, and `metricFilter`'s diagnose mode picks the longest token for near-miss *recall*, not to confirm anything. Claiming "holds nothing matching" while `entityMetricMatches` is populated was a straight contradiction — live, Pfizer returns 20 entries for `q="expense"` and the summary still said the list held nothing.

`unlinked` now means only what the enum advertises. The near-misses still ride along on `resolution`, since they are the whole payload when no handle is emitted (see the reply on `_available_data.ts`).

workers/src/tools/tako_available_data.ts
25eliu: Real bug, fixed — but the suggested test does not work, and I want to flag why.

**`relation.total` ignores the `q` filter.** Measured on prod:

```
node_id=<lockheed>&relation=metrics&limit=20            -> items 20, total 250, capped true
node_id=<lockheed>&relation=metrics&q=Backlog&limit=20  -> items 20, total 250, capped true
```

Identical. So `total > items.length` would fire on essentially every entity with more than 20 metrics and degrade nearly every verdict to `resolution`.

The sound signal is **`next_cursor`** — null on a complete filtered list (Apple/`Gross Margin (%)` returns 1 item, no cursor), non-null when cut off (Lockheed/`Backlog` returns 20 and cursor `'20'`).

Now:
- found -> `pair`, **even if truncated** (truncation cannot un-prove a positive)
- not found + complete -> `unlinked`
- not found + truncated -> `resolution`

Also raised `PAIR_PROBE_LIMIT` 20 -> 100 (endpoint max) to make truncation rare.

This is worse than you knew: **2 of the 4 `unlinked` verdicts in the PR were on truncated lists**, including Lockheed/backlog, the example the PR is argued on. It saw 20 of a list it never finished. The verdict happens to be right — I confirmed exact `Backlog` is absent from the first 100 — but the reasoning was unfounded, and it now correctly degrades to `resolution`, which costs that rescue. PR numbers are stale as a result; noted below.

workers/src/tools/_available_data.ts
25eliu: Fixed, and thank you — it was worse than "not appended". Gating on `verified === "unlinked"` made the clause unreachable on exactly the path where the list is the only payload, so those names were fetched, sliced to 3, and dropped every time.

Now keyed on *having* `entityMetricMatches` rather than on the verdict, so it survives the change to `resolution` from the other thread.

One thing the first attempt got wrong: appending them was not enough. In the backend's own order the slice of 3 was the wrong three. Live, before ordering:

```
nearest to "R&D expense": Operating costs and expenses, Advertising Expense (Normalized), Change in Accrued Expenses
```

`R&D Expenses (Normalized)` was still being dropped. They are now ordered by query overlap, prefix-aware because exact token equality ties `expense` against `expenses` — the same morphology issue the metric half is ungated for. That looseness is confined to display order and touches no verdict and no pin. Live, after:

```
nearest to "R&D expense": R&D Expenses (Normalized), Operating costs and expenses, Advertising Expense (Normalized)
```

workers/src/tools/_pair_confirm.ts
25eliu: Removed — `sameNode` is now id-only, and the measurement says the arm was never earning anything.

Across the 24 prod pairs: **18 `pair` verdicts matched by id, 0 by name alone.** So the id-space-divergence backstop was empirically dead while still able to fire on exactly the KE-812 twin you describe — same name, different id, only one carrying cards — and report `pair` for a node whose id is not the one `next_call` pins.

The module comment now cites that 18/0 measurement instead of the hypothetical, and notes what to re-measure if verdicts ever collapse to `resolution` en masse (which is how divergence would now surface, rather than as a false `unlinked`).

To be clear about what this does *not* fix: the 5 `pair` verdicts that retrieve nothing are all genuine id matches. An edge is not a card, and that is backend coverage.

workers/src/tools/tako_available_data.ts
25eliu: Fair — two opposite conclusions about the same cost, with nothing explaining the difference. Comment updated to say why it comes out differently.

The promotion probe bought a nicer **entity name** in the echo while the pin, and therefore what the priced call retrieves, was unchanged — cosmetic, so any latency beats it. The pair probe decides whether the emitted handle is **pinned at all**, and pinning a metric the entity does not hold is what returns 0 cards.

The cost is real and I have not hidden it: one serial round trip on the common path.

workers/src/tools/tako_available_data.ts
25eliu: Reworded — the final commit did reduce it to a single call and the comment was left behind. It now reads "ONE `graph/related` call ... narrowed by ONE substring filter" and notes that the second variant was measured to change no verdicts before being removed, so a reader does not go hunting for a fan-out that is not there.

robertabbott (approved):

PR #223, opened 2026-08-07
fix(native card): every tab but the first errored, and the card was cut off

25eliu:
Two bugs reported today against the native card on claude.ai, with screenshots. They are independent and get a commit each.

## 1. "There was an error loading the data." on every tab but the first

`embed_proxy.ts`'s header claimed the embed page carries its own data — *"Zero `fetch`/XHR/axios in the page. So there is no API to call and no credential to hold."* That is true of the first paint and false of every interaction after it.

Reading the real production `Card.js`: a multi-tab card inlines `component_data` for the **active tab only** and fetches each other tab at click time.

```js
const I = window.location.href;
const H = new URL("/knowledge/get_data/", I);
await fetch(H, { method: "POST", body: JSON.stringify({ ...vizConfig, pub_id, dark_mode }) })
```

Inside the widget, `window.location.href` is the host's sandbox document. So the POST went to claude.ai's origin — not in `connectDomains`, answered by nothing — and the card rendered its own `CardError` on eleven of Nvidia's twelve tabs.

**Why ChatGPT was fine, and why that is not luck.** `_chart_widget.ts` branches on `window.openai`. ChatGPT commits to the nested `<iframe src="https://tako.com/embed/…">`, where that fetch is same-origin and ordinary. claude.ai blocks `frame-src`, so it takes the native path — the only path that can hit this.

**Fix:** a third route plus a shim to reach it.

- `POST /embed-data/` forwards to `${PUBLIC_BASE_URL}/knowledge/get_data/`. Origin and path are constants; the caller controls only the body, which is the same thing anyone pointed at the public embed page already controls. No cookie, no `Authorization`, no invented `X-CSRFToken` — the endpoint is public (verified against production: bare anonymous POST → 200). No redirect followed. Pinned `application/json` out rather than echoing upstream's, because this is the OAuth origin.
- `dataProxyShim` wraps `window.fetch` in the proxied document and repoints exactly one resolved pathname. A `<base>` tag cannot do this: the page passes `location.href` explicitly rather than relying on `document.baseURI`. Every other fetch the card makes (the `staticPrefix`-relative `geo/*.json` map features) passes through, and every step falls through to the real `fetch` on a throw — a broken chart is worse than a broken tab.

Two things measured rather than assumed, both of which changed the code:

| | |
|---|---|
| **The preflight must NAME `content-type`** | `application/json` is not CORS-safelisted, so the browser preflights and will not send the POST without it. The handler can be perfectly correct and never reached. |
| **The body GROWS** | `Card.js` merges each response's `viz_config` into what it holds and re-posts the whole thing: 118 KB on the first call → **1.29 MB after all twelve tabs**, ~+100 KB/tab. A first cut at a 1 MB cap 413'd on the sixth tab, showing the clicked tab the very error this removes. Cap is 8 MB, and read incrementally with a cancel at the ceiling rather than `arrayBuffer()`-then-check: Cloudflare accepts 100 MB bodies into a 128 MB isolate, and exceeding that kills the isolate instead of throwing something catchable — on the origin that also serves `/authorize`. |

## 2. The card cut off after "Show more"

`NATIVE_HEIGHT_REPORTER` reported on `resize`, `load`, and timers at 400/1200/3000 ms. That covers Card.js mounting. It does not cover the card being interactive. Measured on the production card at a 762 px frame:

| state | natural height |
|---|---|
| initial (Stock tab) | 527 |
| after a tab switch | 619 |
| after `Show more` | 641 |
| Financials tab | 796 |
| Salaries tab | 1027 |

Neither interaction fires `resize` or `load`, the ladder is done by 3 s, and the click happens whenever the user clicks. The embed page's own CSS is `:root { overflow-y: hidden }`, so growth past the last reported height is **not scrolled and not scaled — it is gone**, mid-row. "Show me" in the report was the card's own **Show more** button.

A `ResizeObserver` closes it, with two details that keep it from becoming a layout loop:

- It observes `document.body`, not `documentElement`. The fit is applied to `documentElement`, so observing that would mean every correction observing itself.
- Applying the fit still changes body's *width* (`width:(100/s)%`), which the observer does see. So `last` records body's `scrollHeight` **after** the fit, and a pass whose reading matches returns before touching a single style. Correction → one confirming callback → quiet.

## Also in here

**The proxy was forwarding a live CSRF token into the sandbox.** The body carries `<input type="hidden" name="csrfmiddlewaretoken">`. The island strip never touched it and the fail-closed guard never noticed — it greps the camelCase `csrfToken`, and this is Django's other spelling. So every render forwarded it while this file's header said it did not. **Not exploitable**: our fetch is credential-free, so the token pairs with a `Set-Cookie` we drop, and Django's double-submit check needs the matching cookie. But the header should be true. Stripped, and the guard now covers both spellings.

**Smoke gets the third leg.** Without it this failure is silent post-deploy — the chart draws and the tabs error. It POSTs the real `config-json` island and asserts `component_data` comes back, so a moved upstream contract or an undersized cap goes red in CI instead of in a screenshot.

## Known gap, not fixed here

The card's thumbs up/down (`POST /api/v1/feedback` from `RateThisCard.js`) needs `X-CSRFToken` + `credentials: "include"`, which the proxy strips by design. Those buttons render and silently fail in the native card. Fixing it needs a decision about whether card feedback should work from a third-party sandbox at all, which is a bigger question than this PR.

## Test plan

- [x] `npm run typecheck` — all four tsconfig projects clean
- [x] `npm test` — 1104 + 88 + 68 pass
- [x] `npm run schemas:check` / `registry:check` clean
- [x] Widget test confirmed **red** without the reporter fix (`expected [] to deeply equal [ 'body' ]`)
- [x] End-to-end against a local worker with production bindings, driving the real card from a `file://` document (a harsher base than claude.ai's) through the same fetch → `document.write` the widget does:
  - all **twelve tabs** load, no error text
  - the fit tracks growth: `scale(0.949)` → `scale(0.780)` on Show more → `scale(0.628)` on Financials, each exactly `CAP/natural`
  - **zero** `documentElement` style mutations over four idle seconds — no resize loop
  - the proxied page carries no `csrfToken` and no `csrfmiddlewaretoken`
- [x] New smoke assertions run green against that worker (shim present, island parsed, 200 JSON with `component_data`, `ACAO: *`)
- [ ] Post-merge: `npm run smoke` against staging, then production


robertabbott (commented):
Nine inline. Two of them (the `ACAO: *` write path on `/knowledge/get_data/`, and `readBoundedBody`'s 2x peak) I'd want answered before this reaches production, since both land on the OAuth origin ahead of the OAuth router.

Checked and clean, so you know it was looked at: the shim's escaping and `to=T+u.search`, the injection anchor's reachability from upstream content, the upstream URL being fully constant under `redirect: "manual"`, the pinned response headers, the `last`/`force` state machine's termination, `__FALLBACK_HEIGHT__`, and the blast radius of all six new exports.

workers/src/embed_proxy.ts
```diff
@@ -578,6 +788,197 @@ export async function handleEmbedProxy(
...
+}
+
+/**
+ * Handle `POST /embed-data/` — forward the card's tab-data request to Tako and
+ * hand the JSON back with CORS.
+ *
+ * The upstream URL is `${PUBLIC_BASE_URL}${UPSTREAM_DATA_PATH}` and nothing
+ * about it comes from the request: not the origin, not the path, not the query.
+ * The caller controls only the body, which is the same thing any browser
+ * pointed at the public embed page already controls.
```
robertabbott: This route is the first thing that lets a browser on an arbitrary origin reach `/knowledge/get_data/`, and that endpoint **writes**. `GetData.post` (tako `app/backend/knowledge/views.py:62`) is `csrf_exempt` and unauthenticated, and for any `pub_id` in the body it does `ChartConfig.objects.get(pub_id=pub_id)` then `chart_config.viz_config = ...; chart_config.save()` — no ownership check, and `frozen_after_sharing` is `null=True` so the default is writable. Today Django's `CORS_ALLOWED_ORIGINS` is a closed allowlist (`CORS_URLS_REGEX` commented out), so a drive-by page's preflight for a JSON POST fails and the browser never sends it. `access-control-allow-origin: *` here removes that barrier, and since only `content-type`/`accept` are forwarded, Tako sees one Cloudflare egress IP for all of it.

So is "the caller controls only the body, which is the same thing anyone pointed at the public embed page already controls" still true, given the caller also picks the row that gets overwritten? The shim never sends a real `pub_id` (`window.frameElement?.id ?? ""` is `""` cross-origin), so rejecting a non-empty `pub_id` here would cost the feature nothing. Related: `wrangler.jsonc`'s "What it exposes, reviewed before promoting" block still names two routes.

workers/src/embed_proxy.ts
```diff
@@ -578,6 +788,197 @@ export async function handleEmbedProxy(
...
+    }
+  } catch (err) {
+    console.warn("[mcp] embed data proxy could not read the request body:", err);
+    return "unreadable";
+  }
+  // An `ArrayBuffer` rather than the `Uint8Array` view over it, so the value is
+  // a `BodyInit` under every one of the four tsconfig projects that compile this
+  // module — two of them load DOM lib types, where a generic
+  // `Uint8Array<ArrayBufferLike>` no longer satisfies `BufferSource`.
+  const out = new ArrayBuffer(total);
```
robertabbott, line 917: `const out = new ArrayBuffer(total)` allocates a second full copy while `chunks` is still reachable, so peak is ~2x the cap (16 MB) rather than "the largest thing we ever hold is the cap itself" as the docstring above states — and `body` then stays live across the whole 20 s `DATA_UPSTREAM_TIMEOUT_MS` fetch. Nothing bounds concurrency, so roughly 8 simultaneous 8 MB POSTs from a single IP — well inside 200/60 s — reach the 128 MB isolate limit. That is the exact OOM this function was written to prevent, on the isolate that also serves `/authorize` and `/token`. Could the chunks be released as they are copied, or the body streamed straight into the upstream fetch?

workers/src/embed_proxy.ts
```diff
@@ -540,11 +740,21 @@ export async function handleEmbedProxy(
...
+  //
+  // Both of Django's spellings are covered. The guard used to name only
+  // `csrfToken`, so the `csrfmiddlewaretoken` hidden input in the body sailed
+  // through every render — the strip is new above, and this is what keeps it
+  // honest if the tag's shape changes.
+  for (const { token, stripped } of [
+    { token: "csrfToken", stripped: sanitized.removedCsrf },
+    { token: "csrfmiddlewaretoken", stripped: sanitized.removedCsrfInput },
+  ]) {
+    if (!stripped && html.includes(token)) {
```
robertabbott: `if (!stripped && html.includes(token))` skips the scan entirely once the strip flag is set, so a token can still cross into the sandbox. `sanitizeEmbedHtml`'s island replacer returns `whole` unchanged for any `csrfToken` script whose body is not parseable JSON — so an upstream page carrying the JSON island **plus** a second inline script that merely mentions `csrfToken` sets `removedCsrf = true` and forwards the second one. Neither replacement re-emits the token string (`data-tako-stripped="csrf"` and the HTML comment), so `html.includes(token)` on its own is strictly stronger and would not false-positive on a correctly stripped page. Is the `!stripped &&` conjunct load-bearing? This is the same shape as the miss the PR is fixing.

workers/src/embed_proxy.ts
```diff
@@ -578,6 +788,197 @@ export async function handleEmbedProxy(
...
+    console.warn(
+      `[mcp] embed data proxy got non-JSON content-type "${upstreamType}"`,
+    );
+    return textResponse("upstream returned non-JSON", 502);
+  }
+
+  // Streamed, not buffered. Nothing here needs rewriting — the payload is card
+  // data, not markup with our origin baked into it — and a tab of a large card
+  // measures ~350 KB, which is not worth holding.
+  return new Response(response.body, {
```
robertabbott: `new Response(response.body, ...)` commits `status: 200` and the pinned JSON content-type before a single byte is read, and `clearTimeout(timeout)` has already run in `finally` — so the 20 s bound covered time-to-headers only. Two consequences: a stalled upstream body is an unbounded spinner in the tab the user just clicked (`DATA_UPSTREAM_TIMEOUT_MS`'s own comment notes `Card.js` imposes no deadline, "so this is the only one"), and a stream that errors mid-flight arrives as truncated JSON, `.json()` throws, and the card shows the exact "There was an error loading the data." this route exists to remove — with nothing logged, unlike every other failure path here. Should the signal stay live for the stream?

workers/src/embed_proxy.ts
```diff
@@ -92,13 +118,70 @@ export const EMBED_PROXY_PREFIX = "/embed-html/";
...
+ * roughly +100 KB per tab. A first cut at 1 MB therefore worked for five tabs
+ * and then 413'd — the tab the user clicked showing exactly the error this route
+ * exists to remove, which is why the number is measured rather than guessed.
+ *
+ * 8 MB is ~80 tabs of headroom on the largest card Tako ships, and still bounds
+ * what an anonymous caller can hand Django to parse (or make this worker
+ * buffer). Subtabs are free: one response carries all of a tab's subtabs, so
+ * clicking through EPS/Revenue/EBITDA issues no further requests at all.
+ */
+const MAX_DATA_REQUEST_BYTES = 8 * 1024 * 1024;
```
robertabbott: The ceiling that actually binds is Django's, not this one: `DATA_UPLOAD_MAX_MEMORY_SIZE = 3670016` (3.5 MB) in tako `app/config/settings/base.py:366`. Anything past that is refused upstream and mapped here to a 502 — the tab error again, under a log line reading `upstream HTTP 400` rather than anything about size. That puts real headroom at ~35 tabs rather than the "~80 tabs" this comment claims, and the number lives in a different repo where a change to it would not show up in this PR's tests. Not reachable on a twelve-tab card today. Worth pinning the cap below 3.5 MB, or at least naming the upstream limit here?

workers/scripts/smoke.ts
```diff
@@ -527,6 +541,82 @@ try {
...
+    // config, which the page carries in its `config-json` island. A shim
+    // pointing at a route that 502s upstream is the same user-visible failure as
+    // no shim at all.
+    const island =
+      /<script id="config-json" type="application\/json">([\s\S]*?)<\/script>/.exec(
+        nativeBody,
+      );
+    if (island === null) {
+      console.warn(
+        `[warn] could not find the config-json island in the proxied page, so ` +
```
robertabbott: This `console.warn` turns the whole `/embed-data/` leg into a skip that exits 0. If the island regex stops matching — an id change, an attribute reorder, an extra attribute, or the proxy altering the tag — every assertion below it (200, `ACAO`, JSON content-type, `component_data`) is skipped and the job stays green. That is the failure mode this file already argues against a few lines up for the 404 case: "warnings exit 0, so it would have been reported by a green job whose log nobody opens." Should a non-matching island be a `fail()` when `isDeployedTarget`, the same way the `/embed-html/` 404 is?

workers/src/embed_proxy.ts
```diff
@@ -578,6 +788,197 @@ export async function handleEmbedProxy(
...
+    });
+  } catch (err) {
+    console.warn("[mcp] embed data proxy upstream failed:", err);
+    return textResponse("upstream unavailable", 502);
+  } finally {
+    clearTimeout(timeout);
+  }
+
+  if (!response.ok) {
+    console.warn(`[mcp] embed data proxy upstream HTTP ${response.status}`);
```
robertabbott: None of this route's three 502 log lines carries a correlator, so a 502 cannot be tied back to a card. This line and its siblings for `upstream failed` and the non-JSON content-type print no pub_id, no CF-Ray, and no body excerpt, while `/embed-html/` logs `... for ${pubId}` throughout — and the request body here is opaque JSON, so nothing else in the log identifies which card was being fetched. `handleEmbedProxy` already knows the pub_id at the point it injects the shim; could it append one to the shim target so these lines can name it?

workers/test/widget/widget-dom.test.ts
```diff
@@ -1133,6 +1133,124 @@ describe("native card upgrade (executed)", () => {
...
+
+  it("re-reports its height when the card relayouts, not just on a timer", async () => {
+    const m = mountWidget(staticWidgetHtml());
+    stubFetch(m, () => htmlResponse(NATIVE_HTML));
+    let natural = 400;
+    const layout = stubLayout(m, {
+      offsetHeight: () => natural,
+      // Body's own content height. Kept distinct from `offsetHeight` so the two
+      // roles stay visible: one is reported, one only decides whether to look.
+      scrollHeight: () => natural,
```
robertabbott: `scrollHeight: () => natural` is independent of any style the reporter applies, so this test cannot exercise the property the new comment says stops the loop: that `last=raw()`, read after `d.style.width=(100/s)+'%'` widens the root and reflows the body, matches what the next guard reads. And after the scale is applied further down, `fireObserver()` is never called again — so "correction, one confirming callback, quiet" goes untested in both halves. Could the stub make `scrollHeight` a function of the applied width, then fire once more and assert no further report?

workers/test/widget/widget-dom.test.ts
```diff
@@ -1133,6 +1133,124 @@ describe("native card upgrade (executed)", () => {
...
+      `scale(${500 / 640})`,
+    );
+  });
+
+  it("survives a host with no ResizeObserver", async () => {
+    // The observer is feature-detected: losing it has to leave the ladder's
+    // behaviour intact, not throw and take that down too.
+    const m = mountWidget(staticWidgetHtml());
+    stubFetch(m, () => htmlResponse(NATIVE_HTML));
+    delete (m.widgetWin as unknown as { ResizeObserver?: unknown })
```
robertabbott: `delete (m.widgetWin as ...).ResizeObserver` is a no-op, so this test asserts nothing the pre-PR code did not already satisfy: jsdom does not implement `ResizeObserver`, so the property never existed on the widget window — the test above only works because it assigns its own stub. It also cannot separate the `typeof ResizeObserver!=='undefined'` guard from the surrounding `try{}catch(e){}`, since either alone would pass it. Should it instead define a `ResizeObserver` whose constructor throws, so the catch is what gets proven?

workers/src/embed_proxy.ts
25eliu: Confirmed, and fixed — this was the right thing to block on. I checked `views.py` rather than reasoning about it: `if pub_id:` → `ChartConfig.objects.get(pub_id=pub_id)` → `viz_config`/`data` overwritten and saved, no ownership check, `frozen_after_sharing = BooleanField(null=True)` so the default lands in the writable branch, `csrf_exempt` on the dispatch. And `CORS_ALLOWED_ORIGINS` is the closed allowlist you describe with `CORS_URLS_REGEX` commented out. So `ACAO: *` here was the whole barrier.

No, that sentence was not still true, and I've replaced it rather than softened it — the docstring now says so explicitly, because "the caller controls only the body" reads as reassurance right up until you know the body picks the row.

Took your suggestion: `findUpstreamWrite` refuses a non-empty `pub_id` with a 400, which leaves `if pub_id:` unreachable and makes the route read-only upstream. Two things I added beyond it:

- It refuses non-string truthy values too (`1`, `["a"]`). Django's check is a truthiness test, so a number reaches the write just as well as a string.
- The body is parsed to check it and then **discarded** — the original bytes are forwarded, so nothing Django receives depends on a JSON round-trip here. The parse also buys a JSON-validity check the route was previously forwarding blind, so a non-object body is now a 400 as well.

Confirmed free before relying on it: I logged the real `pub_id` across a full twelve-tab walk in a browser, and it was `""` on all eleven data requests.

`ACAO: *` can't be narrowed instead, which is worth recording next to the gate: the widget document's origin inside the host sandbox is opaque, so there's no value to allowlist. The docstring says that too.

There's now a post-deploy assertion as well — the smoke POSTs a body carrying a non-empty pub_id and requires a 400, because `findUpstreamWrite` passing its unit tests says nothing about whether the build serving traffic calls it.

And the `wrangler.jsonc` block is updated: three routes, plus a paragraph singling out `/embed-data/` as the one to re-read, with the write branch and the gate named.

workers/src/embed_proxy.ts
25eliu: Right on both counts, and the docstring was overclaiming. Fixed two ways, because the peak and the cap were separate problems:

- The chunks are released as they're copied (`chunks.shift()` in the copy loop), so the accumulated chunks and the result never both exist in full. Peak is now ~the cap rather than 2x it.
- The cap is 3 MB rather than 8 MB, for the reason in your other comment — Django refuses past 3.5 MB anyway, so the extra headroom was unusable.

Together that takes the peak per request from 16 MB to ~3 MB, which moves the count of concurrent max-size POSTs needed to reach 128 MB from single digits to past what the meter allows.

I did not stream the body straight into the upstream fetch. With a streamed request body the cap can only be enforced by erroring the stream mid-send, which turns a clean 413 into an aborted upstream request that has to be mapped back to 413 from a fetch rejection — more moving parts than the buffered version, for a body that's measured at 1.29 MB and now capped at 3.

workers/src/embed_proxy.ts
25eliu: No, it wasn't load-bearing — dropped. You're right that it's the same shape as the miss this PR is fixing, which is the part that decided it.

Verified your reasoning about false positives before removing it: neither replacement re-emits the token string (the island becomes `data-tako-stripped="csrf"` over a token-free dict, the input becomes an HTML comment), so an unconditional `includes` can't fire on a correctly stripped page — and every other test in the file that expects a 200 is the standing proof of that.

The flags stay, but only to describe the case in the log: "a strip that reported success — a second occurrence the replacer did not match" vs "no strip — the upstream shape has likely changed". Those are different operational problems and the line now says which one it is.

There's a test for exactly your scenario — the real island plus an unparseable script mentioning `csrfToken`. It asserts `removedCsrf === true` and that the second copy survives the sanitizer, so it's pinning that the route is the thing that has to refuse, not the sanitizer.

workers/src/embed_proxy.ts
25eliu: Yes — the signal now stays live. `clearTimeout` is out of the `finally`; a `settle()` closure clears it when the body actually finishes, so the 20 s bound covers the whole exchange instead of time-to-headers.

On the truncation half, I went with your framing rather than my first instinct. I had planned to buffer the response so a mid-flight failure could be a clean 502, and then noticed buffering doesn't actually buy the user anything: `Card.js` calls `.json()`, so a 502 and a truncated 200 render the *identical* "There was an error loading the data." card. What it would cost is real — I measured responses at up to 1.57 MB, larger than the requests, since the accumulated config comes back with the new tab's data — so a second buffer would have undone the request cap two comments up, on the same isolate.

So streaming stays, and what was actually missing gets fixed: the log. `countingPassThrough` wraps the body and writes `response stream failed after ${seen}B for ${pubId}` when the source errors, which was the one failure on this route that reached a user with nothing written down. It also handles client-side cancel (clears the timer, cancels the upstream read, doesn't log — a closed tab isn't a failure).

Implementation note: this is an explicit reader loop rather than a `TransformStream`, because `Transformer.cancel` isn't in the lib types all four tsconfig projects see.

Tested by erroring a response body mid-stream: the consumer's `.text()` rejects rather than resolving to a silently short document, and the log line carries the pub_id and the byte count.

workers/src/embed_proxy.ts
25eliu: Confirmed — `DATA_UPLOAD_MAX_MEMORY_SIZE = 3670016` at `app/config/settings/base.py:363`. Cap is now 3 MB, under it, so everything this route accepts is something Django will also accept and an over-cap body gets a 413 that names the real problem instead of a 502 logged `upstream HTTP 400`.

The "~80 tabs" claim is gone; the comment now says ~28 against the measured 1.29 MB, names Django's constant and the file it lives in, and says explicitly that the number lives in another repo where a change to it wouldn't show up in this PR's tests — so the cap is deliberately not tuned closer to 3.5 MB.

Also took the smaller half of this: the `!response.ok` log line now appends a note for a 400 specifically, since that's the status a too-large body arrives as and the bare status told you nothing.

workers/scripts/smoke.ts
25eliu: Yes — it's a `fail()` when `isDeployedTarget` now, using the same shape as the `/embed-html/` 404 a few lines up. The comment cites the reason you did: a warning exits 0, so this would have been reported by a green job whose log nobody opens, and it's the leg that proves tabs load at all.

Also added a second assertion in that block while I was here: after the 200, it POSTs a body carrying a non-empty `pub_id` and requires a 400. That's the write gate from your first comment, asserted against the build actually serving traffic rather than against source.

workers/src/embed_proxy.ts
25eliu: Took your suggestion. The route is `/embed-data/{pub_id}` now, `handleEmbedProxy` bakes the id into the shim target, and all three 502 lines plus the two 400 refusals name the card.

Path segment rather than a query, for a reason your suggestion surfaced: `?pub_id=` would have broken the shim, whose `to = T + u.search` would concatenate a second `?` when the target already carried one. As a segment it reuses `PUB_ID_RE` and matches `/embed-html/`'s shape.

Guarded against the correlator becoming anything more than a correlator: it never touches the upstream URL or the body, a malformed one makes the route decline (so it stays invisible, like the other two), and there's a test asserting the upstream URL is the bare constant and does **not** contain the pub_id.

workers/test/widget/widget-dom.test.ts
25eliu: Correct, and the test was worthless as written. Rewrote it.

`scrollHeight` is now derived from the width actually applied to the root, so the reflow the guard depends on is real. And the assertions are on the **read count** rather than the report count, which turned out to be the only signal here that's both synchronous and exact: an early return reads body's scrollHeight once, in the guard; a pass that does work reads it a second time to record `last`. Counting reports can't tell "returned early" from "ran and re-reported the same number", and counting style mutations can't either, since a full pass clears and re-applies to the same values.

So it now asserts `+1` read on a no-op callback, `+2` on the pass that does the work, then fires twice more after the scale and asserts `+1`, no new report, and `transform`/`width` untouched each time. Confirmed by mutation: deleting the guard line reddens it.

One thing worth passing back, since it cost me a wrong first attempt. I initially asserted "no further reports" over a 10 ms window and it failed — not because of the guard, but because the `load` listener fires a **forced** pass in there, and a forced pass is supposed to report regardless. That's why the honest version of this assertion has to be synchronous. It's in the commit message so the next person doesn't re-derive it.

workers/test/widget/widget-dom.test.ts
25eliu: Right, it was a no-op. Replaced with a constructor that throws, as you suggested — and doing that exposed something I'd rather state than paper over.

A throwing constructor with the test asserting "the ladder survives" **still passes with the `try/catch` removed**. I checked by mutation. The reason is placement: the observer is wired last, after the first report, the listeners and the timer ladder, so an escaping throw costs none of them. So the catch isn't load-bearing for the ladder, and any test claiming it is would be theatre.

What the catch is actually worth is not letting an uncaught error loose inside a widget document, which hosts report to the user as a broken app. That's what the test asserts now — a window `error` listener, expected empty — and removing the catch reddens it (`expected [ Array(1) ] to deeply equal []`). It also asserts the constructor really was called, so the catch is on the live path rather than skipped by the `typeof` guard.

The `typeof` half keeps its own separate case, which jsdom satisfies for free by never defining the global — which, as you say, is exactly why the throwing constructor was needed to make the pair mean anything. Both cases now name in-comment what the other one can't reach.

I considered moving the observer earlier so the catch would be load-bearing, and didn't: you want the first height report out as early as possible, and reordering production code to make a test meaningful is the wrong direction.

workers/src/embed_proxy.ts
```diff
@@ -578,6 +840,396 @@ export async function handleEmbedProxy(
...
+  // a `BodyInit` under every one of the four tsconfig projects that compile this
+  // module — two of them load DOM lib types, where a generic
+  // `Uint8Array<ArrayBufferLike>` no longer satisfies `BufferSource`.
+  const out = new ArrayBuffer(total);
+  const view = new Uint8Array(out);
+  let offset = 0;
+  // `shift()` rather than `for..of`: each chunk becomes unreachable as soon as it
+  // is copied, so the accumulated chunks and the result never both exist in full.
+  // See the docstring — this is the half that actually bounds the peak.
+  for (let chunk = chunks.shift(); chunk !== undefined; chunk = chunks.shift()) {
```
robertabbott: `new ArrayBuffer(total)` on line 901 runs before the first `shift()`, so at that instant `chunks` still holds every byte of the body — and `shift()` only drops a reference. Nothing reclaims it until a GC, which has no reason to run inside a synchronous copy loop. So peak is still ~2x the accepted body (6 MB at the new cap), not "the accumulated chunks and the result never both exist in full" as this comment and the docstring both now state.

The 8 MB → 3 MB change is real and does most of the work: peak per request went 16 MB → ~6 MB. The `shift()` did not add to it. Is the claim the one you want to keep, or is the honest bound 2x the cap with the cap doing the bounding?

robertabbott (commented):
Re-review of the delta since 1bba9ad. All nine prior comments are addressed, and each fix carries a test that can actually fail — I checked the write gate against `views.py` (`if pub_id:` is now unreachable), the two real callers in the tako repo (`useGenericScoreData.ts:220` and `useTransactionDrawer.ts:90`, both `pub_id: ""`), and the fail-closed behaviour of `findUpstreamWrite` against every `json.loads`/`JSON.parse` divergence I could construct — a UTF-16 body, `NaN`, duplicate keys, `__proto__`, and truthy non-strings all land on 400 rather than through. The merge from main applied the reporter changes cleanly on top of main's own rewrite of `_chart_widget.ts`.

Three new things, all on 1eb503b. The first two are the same subject as the old `readBoundedBody` comment, so they are worth a number before this reaches production.

workers/src/embed_proxy.ts
```diff
@@ -578,6 +840,396 @@ export async function handleEmbedProxy(
...
+ * route was otherwise forwarding blind.
+ *
+ * Exported for tests.
+ */
+export function findUpstreamWrite(
+  body: ArrayBuffer,
+): "not-json" | "not-an-object" | "writes" | undefined {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(new TextDecoder().decode(body));
```
robertabbott, line 986: This decodes and `JSON.parse`s the whole body, on every request, before the fetch. So a 3 MB body is 3 MB of chunks plus a 3 MB result plus a ~3 MB decoded string plus a parsed object graph, all live at once — larger than anything `readBoundedBody`'s docstring accounts for, and new in this commit rather than pre-existing.

That puts ~12–15 MB per request against the 128 MB isolate, so the concurrent count that reaches the limit is back in single digits — which is the thing the 8 MB → 3 MB change was meant to move past — and `NATIVE_CARD_RATE_LIMITER` is per-IP at 200/60 s, so one caller can hold that many in flight. Same isolate as `/authorize` and `/token`.

Is there a measured per-request peak for this route? Real bodies were measured at 1.29 MB after all twelve tabs; the cap is sized off Django's 3.5 MB instead, which buys ~28 tabs of headroom nothing has asked for. Would sizing it off the measurement be the cheaper lever than making the gate avoid materializing the document?

workers/src/embed_proxy.ts
```diff
@@ -578,6 +840,396 @@ export async function handleEmbedProxy(
...
+    console.warn(`[mcp] embed data proxy upstream failed for ${pubId}:`, err);
+    return textResponse("upstream unavailable", 502);
+  }
+
+  if (!response.ok) {
+    settle();
+    console.warn(
+      `[mcp] embed data proxy upstream HTTP ${response.status} for ${pubId}` +
+        (response.status === 400
+          ? ` — a 400 here is usually the request body exceeding Django's DATA_UPLOAD_MAX_MEMORY_SIZE (3.5 MB)`
```
robertabbott: `MAX_DATA_REQUEST_BYTES` is now 3 MB and `DATA_UPLOAD_MAX_MEMORY_SIZE` is 3670016 (3.5 MB, `app/config/settings/base.py:393`), so a body large enough to trip `RequestDataTooBig` gets a 413 from the cap above and never reaches Django. Lowering the cap made size the one cause this 400 cannot have.

What can arrive as a 400 is the `jsonschema.exceptions.ValidationError` branch of `GetData.post`, which returns `HttpResponseBadRequest` with an empty message outside `DEBUG` — so the log line is the only place that could name it. Should the note point at the schema instead, or is the size hint here for a future where the cap moves above 3.5 MB?

workers/src/embed_proxy.ts
25eliu: You're right, and I've retracted the claim rather than tried to earn it.

`new ArrayBuffer(total)` runs with every chunk still live, so the peak is ~2x the accepted body regardless of how the copy loop is written — dropping references cannot lower a peak that has already been reached. The docstring now says exactly that: the honest bound is 2x the cap, and the **cap** is what does the bounding. Which is what moved the cap; see the reply on `findUpstreamWrite`.

The `shift()` loop is gone too, back to a plain `for..of`. Its only justification was the sentence you just falsified, and it bought nothing on its own terms either — `chunks` goes unreachable one line later when the function returns. A loop written oddly for a reason that no longer holds is how the next person re-derives this.

I did consider actually getting the peak to ~1x and decided against it, since it's the obvious follow-on question. Pre-allocating the cap makes every 118 KB first request pay 2 MB. A doubling buffer needs a trailing `slice`, which is *3x*, not 1x — unless it returns a `Uint8Array` subarray, which reopens the `BufferSource` typing the docstring above it already documents across the four tsconfig projects. Neither is worth it against a cap that's now sized off measurement.

One piece of your framing I'd push back on, and I've written it into the docstring rather than just asserting it here. That 2x doesn't **stack**. A Workers isolate is single-threaded and the copy loop has no `await` in it, so at most one request can be inside it at a time — same for the decode and parse in `findUpstreamWrite`. What concurrent requests genuinely hold at once is one body each: accumulating chunks here, then the returned buffer for the duration of the upstream fetch. So N in-flight requests cost ~N × body plus one transient peak, which puts 128 MB at ~60 simultaneous max-size POSTs rather than single digits. The cap is still the lever — this is unauthenticated traffic ahead of `/authorize` — but the number it has to be sized against is not the one either of us was using.

workers/src/embed_proxy.ts
25eliu: Took the lever. `MAX_DATA_REQUEST_BYTES` is 2 MB, and the change that matters more than the number is what it's sized off: the **measurement** (1.55x the 1.29 MB a real twelve-tab walk reaches, ~19 tabs against the 12 that card has), not Django's ceiling. Staying under 3.5 MB is kept as a property — everything this route accepts is something Django will also accept — but it's no longer the reason for the value. You're right that the old number bought ~29 tabs nothing had asked for, and this is not headroom that's free.

`findUpstreamWrite` now says what it costs, too. The gate is worth its memory, but the docstring described only what it checks, which is how a per-request peak lands in review rather than in the file.

There is no measured per-request peak for this route — I have request and response sizes, not heap. So I'd rather correct the arithmetic than quote a number I don't have. Two things change it:

- The decode, the parse and the check are one **synchronous** block — no `await` between `TextDecoder().decode` and the return — and an isolate is single-threaded. So at most one request's decoded string and object graph can be live at any instant. They can't stack across concurrent requests, and neither can the 2x copy peak in `readBoundedBody`, for the same reason.
- What *does* stack is one body per in-flight request: chunks while reading, then the buffer for the 20 s the upstream fetch may run. The string and the graph are already garbage by then — `findUpstreamWrite` returns before the fetch starts.

So the concurrent bound is ~N × cap plus one transient, which at 2 MB is ~60 simultaneous max-size POSTs to reach 128 MB, not single digits. That's the accounting now written into both docstrings.

I did not make the gate avoid materializing the document. Every cheap version of that is a scan for `"pub_id"` in the raw bytes, and it has to agree with `json.loads` on `"pub_id"`, duplicate keys and UTF-16 — the exact parity you probed last round and that a full parse gets for free. Trading a fail-closed check for a hand-rolled one, on the gate that's the reason this route is safe to expose, is the wrong side of that trade. The cap was the cheaper lever, as you said.

workers/src/embed_proxy.ts
25eliu: Confirmed and repointed. I read `GetData.post` rather than reasoning from the status: the reachable 400 is the `jsonschema.exceptions.ValidationError` branch, and outside `DEBUG` it's `HttpResponseBadRequest("")` — an empty body — so this log line really is the only record that will ever exist of it.

The note now names that branch and says the response body is empty by design, so nobody goes looking for detail that was never sent. The size hint is gone rather than kept for a hypothetical future: the cap moved to 2 MB in the sibling comment, so it's further under `DATA_UPLOAD_MAX_MEMORY_SIZE` than before, and a comment describing a cause the code makes impossible is worse than no comment. If the cap ever goes above 3.5 MB, that's the change that should bring its own note.

The `MAX_DATA_REQUEST_BYTES` docstring still names Django's constant, since that's a real ceiling worth knowing about when the number moves — it's just no longer what sizes it.

robertabbott (approved):

PR #224, opened 2026-08-07
fix(widget): a chart-less call left an unexplained grey box

25eliu:
## The bug

Reported on ChatGPT, 2026-08-07: a tool call whose result carried no chart rendered an empty grey card above the answer.

It reproduces 100%. It is not a glitch — it is what a mounted widget with nothing to draw looks like on a host with a minimum card height, and it fires on every chart-less `tako_search` / `tako_answer` / `tako_visualize` call, plus every failed one.

## Why the mount can't just be skipped

Hosts read the widget URI from `tools/list` registration `_meta` (`openai/outputTemplate`, `ui.resourceUri`). That metadata is written once per tool, before any call, so it cannot know whether the next result will have a card. Confirmed on the wire against production — a zero-card search still ships `_meta: {"ui": {"resourceUri": "ui://tako/embed/chart"}}`.

The bundle's half was already right: it hides everything, sets the document to zero height, and notifies `size-changed {height: 0}`. I verified that against the **live production bundle bytes** pulled via `resources/read` and executed under jsdom. ChatGPT keeps its card regardless.

## The fix, one change per side

**The bundle labels the space it cannot escape.** `collapse()` shows a centered "No chart for this result".

`position: fixed` with explicit offsets is what makes this safe instead of a second bug:

| host behaviour | outcome |
|---|---|
| honours the shrink | viewport is 0 px tall, nothing renders — unchanged |
| keeps a minimum-height card (ChatGPT) | the line centers in the space already reserved |
| sizes the frame from content | measured in Chromium: `documentElement.scrollHeight` reads identically with the label shown or hidden, `body.scrollHeight` stays 0 |

No height above zero is notified after the collapse, so the label can never reserve a box a host already gave up.

The colour comes from the resolved host theme (`#6b7280` light, `#b4b8bd` dark). The bundle's placeholder grey measured ~2.2:1 against a light host — a label only its author can read. Caught by rendering it, not by a test.

**The server stops asking.** A chart-less result now names no ui resource at all: the per-call resolver returns `undefined` rather than falling back to the static bundle, and `mcp.ts` omits `ui` / `ui/resourceUri`. Production ships `_meta: {ui: {…}}` on a zero-card search today; after this it ships no `_meta`. This is the only lever that *removes* the box rather than dressing it, and it reaches hosts that decide per result. ChatGPT and claude.ai are not among them — hence both halves.

Also corrected a comment in `index.test.ts` claiming `tako_search` throws on empty for ChatGPT to dodge this exact gap. It does not, and has not since the zero-result `guidance` path replaced the throw.

## Verified

Rendered in a local MCP Apps host harness at a ChatGPT-like 145 px card, light and dark, with the real zero-card payload. Host log confirms `size-changed {"width":760,"height":0}` for both:

Both panes show the label centered in the reserved card, tinted for that theme, with the widget canvas painted in the host's own surface colour so no inner rectangle appears. (Screenshot available locally — GitHub attachments need a browser upload.)

- 1083 worker tests, 91 widget tests, 68 script tests pass; all four typecheck projects clean; `registry:check` ok.
- +5 tests: the empty state on both the MCP Apps and ChatGPT delivery paths, the theme-correct colour on both themes, the never-notify-above-zero invariant (`notifyIntrinsicHeight` called exactly `[1, 0]`), no empty state once a chart paints, and both sides of the resolver branch asserted on the wire.

## Cross-client verification (asked for explicitly)

Every client, both result shapes. The widget-reference matrix is now a test (`mcp.test.ts`), so a change that helps one host at another's expense fails in CI:

| client | chart | no chart |
|---|---|---|
| claude | per-`pub_id` reference | **absent** |
| chatgpt | per-`pub_id` reference | **absent** |
| unknown (Cursor, Windsurf, Gemini CLI, Claude Code, Agent SDK) | absent, PNG content block | absent, nothing attempted |

The `unknown` row is what says nothing else regressed: `widgetSuppressed` leaves `ui` undefined in `registerTool`, so those clients never reach the resolver at all — the branch cannot touch the long tail of MCP hosts even in principle, and the portable image block is asserted still present.

**Executed in Chromium against real production payloads**, driving each host's actual protocol (Claude: MCP Apps handshake with `hostContext` theme + surface, then `ui/notifications/tool-result`; ChatGPT: `window.openai.toolOutput` defined before the bundle runs):

| scenario | result |
|---|---|
| Claude, zero-card | label shown, fills the 760×145 box, light `rgb(107,114,128)` / dark `rgb(180,184,189)`, `size-changed height: 0`, `body.scrollHeight: 0`, `documentElement.scrollHeight` unchanged at 145 |
| Claude, real chart (baked `image_data_url`, 176 KB) | PNG paints at 349 px, then the native card upgrades in place, frame 409 px, label never shown |
| ChatGPT, real chart (32 KB structuredContent) | embed iframe committed to the right URL, heights `[1, 720, 460, 449]`, label never shown |

Nothing else reads what was removed: `scripts/smoke.ts` asserts the **registration** `_meta.ui.resourceUri` from `tools/list` (untouched) and hits `/embed-html/` directly; the baked per-chart resource and its `buildFallbackWidgetHtml` path always carry their own content. Omitting `_meta` entirely is an already-exercised path — every non-widget tool ships results that way today.

**Not executed on WebKit or Gecko.** `playwright install webkit` produces a broken 15 MB directory with no launcher in this environment, twice. Rather than claim coverage I do not have, the CSS now removes the dependency: `height: 100vh` alongside the fixed offsets, so the label fills the frame's viewport whether an engine resolves fixed-in-iframe against the viewport (all current engines) or against the document (WebKit historically, where `top/bottom: 0` on our deliberately zero-height document would have collapsed it to nothing). It is 0 either way when the host honoured the shrink, so a collapsed widget still cannot become visible or measurable. Worst case on an untested engine is the label not centering — never worse than today's blank box.

## Test plan

- [ ] Deploy to staging, run a `tako_search` that returns zero cards from ChatGPT dev mode, confirm the card reads "No chart for this result" instead of blank.
- [ ] Same call from claude.ai, confirm the label and the theme colour.
- [ ] Confirm a chart-bearing call is untouched: chart renders, and the result `_meta` still carries the per-`pub_id` `ui.resourceUri`.

### DIFF
--- a/workers/src/mcp.test.ts
+++ b/workers/src/mcp.test.ts
@@ -618,6 +618,42 @@ describe("chart render gates per client", () => {
       (result.structuredContent as { guidance?: string }).guidance,
     ).toMatch(/tako_available_data/);
   });
+
+  it("a chart-less result references no ui resource at all", async () => {
+    // The result-side half of the empty-widget fix. A zero-card call has
+    // nothing to render, so it must not point at a widget: a host that decides
+    // per RESULT then mounts nothing, and no empty card appears.
+    //
+    // This does NOT reach ChatGPT or claude.ai — both read the widget URI from
+    // `tools/list` registration `_meta`, which is static per tool and stays
+    // declared (asserted in index.test.ts). Their empty card is handled inside
+    // the bundle, by the labelled empty state. This is for spec-compliant hosts
+    // that honour the per-call reference, and it is the only lever that removes
+    // the box rather than dressing it.
+    mockFetchSequence([
+      jsonResponse(200, { cards: [], web_results: [], request_id: "req-3" }),
+    ]);
+
+    const result = await callSearch("claude");
+
+    const meta = result._meta as Record<string, unknown> | undefined;
+    expect(meta?.ui).toBeUndefined();
+    expect(meta?.["ui/resourceUri"]).toBeUndefined();
+  });
+
+  it("a chart-bearing result still points at its baked per-chart widget", async () => {
+    // The other side of the same branch: declining to reference a resource on
+    // an empty result must not cost the populated one its per-pub_id URI.
+    mockFetchSequence([searchResponse(), realPngResponse()]);
+
+    const result = await callSearch("claude");
+
+    const meta = result._meta as
+      | { ui?: { resourceUri?: string }; "ui/resourceUri"?: string }
+      | undefined;
+    expect(meta?.ui?.resourceUri).toBe("ui://tako/embed/chart/c1");
+    expect(meta?.["ui/resourceUri"]).toBe("ui://tako/embed/chart/c1");
+  });
 });
 
 describe("detectMcpClient", () => {
--- a/workers/src/mcp.test.ts
+++ b/workers/src/mcp.test.ts
@@ -654,6 +654,61 @@ describe("chart render gates per client", () => {
     expect(meta?.ui?.resourceUri).toBe("ui://tako/embed/chart/c1");
     expect(meta?.["ui/resourceUri"]).toBe("ui://tako/embed/chart/c1");
   });
+
+  /**
+   * The full client × has-chart matrix for the resource reference, because the
+   * fix is per-RESULT and the surfaces it touches are per-CLIENT. Every cell is
+   * asserted, so a change that helps one client at another's expense fails
+   * here rather than in someone's chat window.
+   *
+   *                 chart          no chart
+   *   claude        per-pub_id     absent
+   *   chatgpt       per-pub_id     absent
+   *   unknown       absent         absent   ← never had a widget; PNG block instead
+   *
+   * The `unknown` row is the load-bearing one for "nothing else regressed":
+   * those clients never reach the resolver at all (`widgetSuppressed` leaves
+   * `ui` undefined in `registerTool`), so the branch cannot touch the long tail
+   * of MCP hosts even in principle.
+   */
+  it("references a widget per client and per result, and never for unknown clients", async () => {
+    const uiOf = (result: { _meta?: Record<string, unknown> }) =>
+      (result._meta as { ui?: { resourceUri?: string } } | undefined)?.ui
+        ?.resourceUri;
+
+    // chatgpt, chart present: keeps its reference. (One fetch only — ChatGPT
+    // skips the PNG bake, and a second would throw.)
+    mockFetchSequence([searchResponse()]);
+    expect(uiOf(await callSearch("chatgpt"))).toBe("ui://tako/embed/chart/c1");
+
+    // chatgpt, no chart: no reference. The registration `_meta` still carries
+    // `openai/outputTemplate`, so ChatGPT mounts anyway and the bundle's empty
+    // state is what covers it — see chatgpt-path.test.ts.
+    mockFetchSequence([
+      jsonResponse(200, { cards: [], web_results: [], request_id: "req-4" }),
+    ]);
+    expect(uiOf(await callSearch("chatgpt"))).toBeUndefined();
+
+    // unknown, chart present: no widget reference ever, and the PNG content
+    // block instead — the portable path, untouched by this branch.
+    mockFetchSequence([searchResponse(), pngResponse()]);
+    const unknownWithChart = await callSearch("unknown");
+    expect(uiOf(unknownWithChart)).toBeUndefined();
+    expect(
+      unknownWithChart.content.filter((b) => b.type === "image"),
+    ).toHaveLength(1);
+
+    // unknown, no chart: no reference, no image block, and no second fetch
+    // (no image_url to fetch) — nothing to render and nothing attempted.
+    mockFetchSequence([
+      jsonResponse(200, { cards: [], web_results: [], request_id: "req-5" }),
+    ]);
+    const unknownEmpty = await callSearch("unknown");
+    expect(uiOf(unknownEmpty)).toBeUndefined();
+    expect(unknownEmpty.content.filter((b) => b.type === "image")).toHaveLength(
+      0,
+    );
+  });
 });
 
 describe("detectMcpClient", () => {
--- a/workers/src/mcp.ts
+++ b/workers/src/mcp.ts
@@ -1228,14 +1228,30 @@ function registerTool(
           // `AnyToolModule` erases handler types at the registry
           // boundary; resolvers narrow it themselves.
           const resolvedUri = ui.dynamic.resolveUriFromInput(input, output);
-          resultMeta = {
-            ...(resultMeta ?? {}),
-            ui: {
-              ...((resultMeta?.ui as Record<string, unknown> | undefined) ?? {}),
-              resourceUri: resolvedUri,
-            },
-            "ui/resourceUri": resolvedUri,
-          };
+          // `undefined` = this call produced nothing to render. Then the
+          // result names NO ui resource: pointing at the static bundle
+          // regardless (what this used to do) is an instruction to mount a
+          // widget for a result that has no chart in it, and on a host with a
+          // minimum widget-card height that instruction is exactly the empty
+          // grey box users report. A host that decides per RESULT now mounts
+          // nothing at all.
+          //
+          // Hosts that read the URI from `tools/list` registration `_meta`
+          // are unaffected — that metadata is static per tool and still
+          // declares the widget, because a tool descriptor cannot know
+          // whether a future call will have a card. ChatGPT and claude.ai are
+          // both in that bucket; their chart-less mount is handled inside the
+          // bundle by the labelled empty state in `collapse()`.
+          if (resolvedUri !== undefined) {
+            resultMeta = {
+              ...(resultMeta ?? {}),
+              ui: {
+                ...((resultMeta?.ui as Record<string, unknown> | undefined) ?? {}),
+                resourceUri: resolvedUri,
+              },
+              "ui/resourceUri": resolvedUri,
+            };
+          }
         } catch (err) {
           console.error(
             `dynamic.resolveUriFromInput failed for ${tool.name}:`,
--- a/workers/src/tools/_chart_widget.test.ts
+++ b/workers/src/tools/_chart_widget.test.ts
@@ -87,19 +87,34 @@ describe("chart widget HTML", () => {
   });
 
   it("positions the empty state out of flow so it cannot grow the box", () => {
-    // Three properties carry the whole design and none is cosmetic:
-    // `position: fixed` fills the host's reserved viewport without
-    // contributing to `scrollHeight` (a host that sizes from content must not
-    // see the label as a reason to make room); `height: 100vh` makes that hold
-    // even on an engine that resolves fixed-in-iframe against the document
-    // rather than the viewport, where `top/bottom: 0` alone would collapse the
-    // label on a zero-height document; and the individual offsets rather than
-    // the `inset` shorthand keep it parseable by older engines.
+    // Two properties carry the design and neither is cosmetic: `position:
+    // fixed` fills the host's reserved viewport without contributing to
+    // `scrollHeight` (a host that sizes from content must not see the label as
+    // a reason to make room), and `height: 100vh` makes that hold even on an
+    // engine that resolves fixed-in-iframe against the document rather than
+    // the viewport, where `top/bottom: 0` alone would collapse the label on a
+    // zero-height document.
+    //
+    // Long-hand offsets vs the `inset` shorthand is NOT asserted: both have
+    // shipped everywhere since 2021, so pinning the spelling would only fail a
+    // future cleanup.
     const ui = buildChartAppUiResourceFromOutputPubId(ENV);
     expect(ui.html).toContain("#tako-empty");
     expect(ui.html).toMatch(/#tako-empty\s*\{[^}]*position:\s*fixed/);
     expect(ui.html).toMatch(/#tako-empty\s*\{[^}]*height:\s*100vh/);
-    expect(ui.html).not.toMatch(/#tako-empty\s*\{[^}]*inset:/);
+  });
+
+  it("ships a readable empty-label colour for a host that declares no theme", () => {
+    // The silent-host fallback lives in CSS, not in `collapse()`: base is the
+    // light value (an unstyled frame composites to an opaque white base) with a
+    // `prefers-color-scheme` dark override. A single compromise grey was worse
+    // in BOTH directions — 3.25:1 against 4.83:1 on white — and 13px normal
+    // text needs 4.5:1, not the 3:1 large-text threshold.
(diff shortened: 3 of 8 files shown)


workers/src/tools/_chart_widget.ts
```diff
@@ -1483,10 +1517,41 @@ const WIDGET_HTML = `<!doctype html>
...
+      // (\`applyHostSurface\` paints the canvas to match), so the body's
+      // placeholder grey lands around 2:1 against a light host — legible to
+      // whoever wrote it and to nobody else. Pick per resolved theme instead:
+      // ~4.7:1 on light, ~5:1 on dark. \`hostTheme()\` is known by now on every
+      // path that has a theme at all (the no-chart guard runs after
+      // \`applyHostSurface\`); the fallback is the one grey that clears 3:1 in
+      // both directions, for the watchdog case where no host ever answered.
+      var t = hostTheme();
+      empty.style.color =
+        t === "light" ? "#6b7280" : t === "dark" ? "#b4b8bd" : "#8b8f95";
```
noahjax: **[Suggestion]** The `#8b8f95` fallback is tuned against a backdrop this file documents as not occurring.

When `hostTheme()` is null, `applyHostSurface` is a no-op in both tiers (tier 1 needs `mcpHostSurface`, tier 2 is gated on `mcpHostTheme !== null`), and the measured note at ~line 866 says an unstyled cross-origin / out-of-process frame composites to an **opaque white base**. So the silent-host case is a known-white backdrop, not an unknown one.

Numbers on 13px text:

| colour | on white | on `#191817` | worst case |
|---|---|---|---|
| `#8b8f95` (this fallback) | 3.25:1 | 5.40:1 | **3.25** |
| `#6b7280` (the light value) | 4.83:1 | 3.66:1 | **3.66** |

The compromise grey loses on both the worst case and the backdrop the file says actually shows up. Note also that 3:1 is the large-text threshold; this is 13px normal text, where AA is 4.5:1.

Cleaner still: make the light value the CSS base and add `@media (prefers-color-scheme: dark) { #tako-empty { color: #b4b8bd } }`, letting JS override only when the host declared a theme. That clears 4.5:1 in both directions for the null case, using a signal the iframe does receive.

workers/src/tools/_chart_widget.ts
```diff
@@ -1483,10 +1517,41 @@ const WIDGET_HTML = `<!doctype html>
...
+      // Contrast, not decoration. The label sits on the host's own surface
+      // (\`applyHostSurface\` paints the canvas to match), so the body's
+      // placeholder grey lands around 2:1 against a light host — legible to
+      // whoever wrote it and to nobody else. Pick per resolved theme instead:
+      // ~4.7:1 on light, ~5:1 on dark. \`hostTheme()\` is known by now on every
+      // path that has a theme at all (the no-chart guard runs after
+      // \`applyHostSurface\`); the fallback is the one grey that clears 3:1 in
+      // both directions, for the watchdog case where no host ever answered.
+      var t = hostTheme();
+      empty.style.color =
```
noahjax: **[Suggestion]** `collapse()` is shared with the 10 s watchdog, which fires when *nothing* arrived — a failed call, but also a merely slow delivery. It sets `rendered = true`, so a tool-result landing at 10.1 s is discarded.

Before this PR that produced an invisible zero-height widget. Now it produces "No chart for this result" on a min-height host — an affirmative claim that is wrong precisely when the chart was late rather than absent.

`widget-dom.test.ts:430` asserts the same label on the failed-call path, so this reads as a decision rather than an oversight. Worth either saying why one string covers all three cases, or passing a flag so the watchdog path stays blank and the label is reserved for "structured content arrived and had no chart".

workers/test/widget/chatgpt-path.test.ts
```diff
@@ -181,6 +181,32 @@ describe("ChatGPT Apps SDK delivery paths", () => {
...
+    // `openai/outputTemplate` registration metadata, so a zero-card search
+    // still gets a widget card, and it holds that card at its minimum height
+    // through the shrink to zero — an unexplained grey void beside a working
+    // answer. The bundle cannot un-mount itself; it can only make the space
+    // read as intentional. Delivered on PATH 1, the route ChatGPT actually
+    // uses, because a guard that only works over postMessage would not have
+    // covered this host at all.
+    const heights: number[] = [];
+    const m = mountWidget(html(), {
+      toolOutput: { cards: [], web_results: [], usage: {} },
```
noahjax, line 195: **[Suggestion]** The host this was reported on is the one whose label colour is not asserted.

This mount sets no `theme` on `openai`, so `hostTheme()` falls through to the null fallback and the test exercises `#8b8f95` — not ChatGPT's themed value. `widget-dom.test.ts` covers light and dark, but only over the MCP Apps handshake, and `window.openai.theme` is the *only* theme source on ChatGPT.

Adding `theme: "light"` / `"dark"` here and asserting the resulting colour would make a regression in that read fail in CI rather than in a chat window.

workers/src/tools/types.ts
```diff
@@ -263,8 +263,16 @@ export interface AppUiResource {
      * of a tool result (e.g. during pre-registration validation in
      * tests); resolvers should fall back to a sensible default URI in
```
noahjax: **[Nit]** This paragraph and the new one below it now give contradictory contracts for the same return value: "resolvers should fall back to a sensible default URI" when `output` is `undefined`, versus "return `undefined` to say this call produced nothing to render."

The shipped resolver does the latter for both conditions — `pubId === ""` catches a missing `output` too. Worth rewording the older paragraph so the next reader does not implement the retired rule.

workers/src/tools/_chart_widget.test.ts
```diff
@@ -86,6 +86,22 @@ describe("chart widget HTML", () => {
...
+    // see the label as a reason to make room); `height: 100vh` makes that hold
+    // even on an engine that resolves fixed-in-iframe against the document
+    // rather than the viewport, where `top/bottom: 0` alone would collapse the
+    // label on a zero-height document; and the individual offsets rather than
+    // the `inset` shorthand keep it parseable by older engines.
+    const ui = buildChartAppUiResourceFromOutputPubId(ENV);
+    expect(ui.html).toContain("#tako-empty");
+    expect(ui.html).toMatch(/#tako-empty\s*\{[^}]*position:\s*fixed/);
+    expect(ui.html).toMatch(/#tako-empty\s*\{[^}]*height:\s*100vh/);
+    expect(ui.html).not.toMatch(/#tako-empty\s*\{[^}]*inset:/);
```
noahjax: **[Nit]** This assertion locks in a stylistic choice on a justification that does not hold. `inset` shipped in Safari 14.1, Chrome 87 and Firefox 66 (2021), so "the offsets parse in every engine that has ever run this bundle" is equally true of `inset`.

Keep the four offsets if you prefer reading them — but a test that fails a future cleanup should either go, or be re-anchored on something checkable. The `height: 100vh` assertion above it is well-founded and worth keeping.

noahjax (approved):
Approving. I verified the two things that would make this unsafe and both hold by construction: `topCardChartFields` emits `pub_id`/`embed_url`/`image_url` as a unit and `tako_visualize` throws on an empty `card_id`, so the server's `pub_id` gate and the bundle's `embed_url`/`image_url` gate cannot disagree and no per-result host can lose a chart that would have painted; and `embedOrigin` is only armed in the render path, so no stray `tako-embed-height` can notify a height after the collapse. The "ships no `_meta` at all" claim also checks out — `nativeCardUrl` bails on an empty `pubId`, so `extraMeta` returns `undefined` on a zero-card result and `resultMeta` stays unset.

Suggestions below, none blocking. The one I'd most like an answer on is the null-theme fallback colour: the grey chosen to hedge both directions has a *worse* worst case than the light value already in the file, and the case it hedges against is one this same file measures as not happening.

workers/src/tools/_chart_widget.ts
robertabbott: The label's colour is set once inside `collapse()` and never revised, so a theme flip after the collapse leaves exactly the contrast this PR exists to fix — should the `host-context-changed` handler re-apply it? That handler (line 2036) calls `mergeHostContext` + `applyHostSurface`, which repaints the canvas to the new theme, but nothing touches `empty.style.color`: a dark→light flip leaves `#b4b8bd` on a now-light surface, under 2:1. Unlike the chart itself — a baked PNG with no light twin, which is why that handler's comment says it cannot re-theme — this label is live DOM and re-colouring is one line. `widget-dom.test.ts:1481` already has the harness for delivering that notification.

workers/src/mcp.ts
```diff
@@ -1228,14 +1228,30 @@ function registerTool(
...
+          // regardless (what this used to do) is an instruction to mount a
+          // widget for a result that has no chart in it, and on a host with a
+          // minimum widget-card height that instruction is exactly the empty
+          // grey box users report. A host that decides per RESULT now mounts
+          // nothing at all.
+          //
+          // Hosts that read the URI from `tools/list` registration `_meta`
+          // are unaffected — that metadata is static per tool and still
+          // declares the widget, because a tool descriptor cannot know
+          // whether a future call will have a card. ChatGPT and claude.ai are
```
robertabbott, line 1242: This new comment says claude.ai reads the URI from `tools/list` registration `_meta`; the comment 27 lines above it, at 1213-1215, still says "claude.ai reads these from the result's `_meta`, so per-call routing works". The registration-side conclusion at 908-925 ("Verified delivered correctly (curl), but claude.ai loads the widget URI from `tools/list` registration metadata and ignores per-call overrides") backs the new one, so 1213-1215 is the stale claim — and it is the claim a reader would use to decide whether this branch removes claude.ai's empty box or only labels it. Can it be corrected here? `_chart_widget.ts:145` carries the same stale line.

workers/src/tools/types.ts
```diff
@@ -263,8 +263,16 @@ export interface AppUiResource {
...
+     * Return `undefined` to say "this call produced nothing to render".
+     * `mcp.ts` then omits the result-level `ui` keys entirely rather
+     * than falling back to the static URI, so a host that decides per
+     * RESULT mounts no widget at all. Hosts that read the URI from
+     * `tools/list` registration `_meta` (ChatGPT, claude.ai) mount
+     * regardless — for them the bundle's own empty state is what keeps
+     * a chart-less call from looking broken (see `collapse()`).
      */
-    resolveUriFromInput: (input: unknown, output?: unknown) => string;
+    resolveUriFromInput: (input: unknown, output?: unknown) => string | undefined;
```
robertabbott, line 276: The signature now permits `undefined`, but the doc two lines above still tells resolvers to "fall back to a sensible default URI" when `output` is absent — which is precisely what `buildChartAppUiResourceFromOutputPubId` stopped doing. A future resolver author following that sentence reintroduces the empty card. Should it be replaced by the new rule rather than left sitting above it?

workers/src/mcp.test.ts
```diff
@@ -618,6 +618,97 @@ describe("chart render gates per client", () => {
...
+   *   claude        per-pub_id     absent
+   *   chatgpt       per-pub_id     absent
+   *   unknown       absent         absent   ← never had a widget; PNG block instead
+   *
+   * The `unknown` row is the load-bearing one for "nothing else regressed":
+   * those clients never reach the resolver at all (`widgetSuppressed` leaves
+   * `ui` undefined in `registerTool`), so the branch cannot touch the long tail
+   * of MCP hosts even in principle.
+   */
+  it("references a widget per client and per result, and never for unknown clients", async () => {
```
robertabbott, line 674: The client × has-chart matrix only exercises `tako_search` (`callSearch` pins the tool name), but `tako_answer.ts:401` and `tako_visualize.ts:475` share the changed resolver, and `tako_answer` is the tool most likely to hit the new branch — `topCardChartFields` yields `undefined` whenever nothing is citable, so an ordinary text-only answer now ships no `ui` reference where it previously named the static bundle. Can the matrix be parameterised over the three chart tools?

workers/test/widget/widget-dom.test.ts
```diff
@@ -326,6 +330,83 @@ describe("static widget bundle (executed)", () => {
...
+    deliver(m, toolResult({ cards: [] }), m.wrapperWin);
+    const heights = (
+      m.toParent.filter(
+        (msg) =>
+          (msg as { method?: string }).method ===
+          "ui/notifications/size-changed",
+      ) as Array<{ params: { height: number } }>
+    ).map((msg) => msg.params.height);
+    // 1 px is the mount-time floor sent before any data arrives.
+    expect(Math.max(...heights.slice(1))).toBe(0);
```
robertabbott, line 393: This test never exercises anything *after* the collapse, so it does not pin the invariant its name claims. `notifyHeight(1)` at mount and `notifyHeight(0)` in `collapse()` are the only two size-changed messages on this path, so `Math.max(...heights.slice(1))` is `Math.max(0)` — the same assertion the test above already makes via `sizeChanges.at(-1)`. Should it deliver a post-collapse event (a second tool-result, a theme flip, the watchdog under fake timers) and then assert no further notification? The ChatGPT-path test's `toEqual([1, 0])` is the stronger form.

workers/src/tools/_chart_widget.test.ts
```diff
@@ -86,6 +86,22 @@ describe("chart widget HTML", () => {
...
+    // contributing to `scrollHeight` (a host that sizes from content must not
+    // see the label as a reason to make room); `height: 100vh` makes that hold
+    // even on an engine that resolves fixed-in-iframe against the document
+    // rather than the viewport, where `top/bottom: 0` alone would collapse the
+    // label on a zero-height document; and the individual offsets rather than
+    // the `inset` shorthand keep it parseable by older engines.
+    const ui = buildChartAppUiResourceFromOutputPubId(ENV);
+    expect(ui.html).toContain("#tako-empty");
+    expect(ui.html).toMatch(/#tako-empty\s*\{[^}]*position:\s*fixed/);
+    expect(ui.html).toMatch(/#tako-empty\s*\{[^}]*height:\s*100vh/);
```
robertabbott, line 104: `/#tako-empty\s*\{[^}]*height:\s*100vh/` also matches `min-height: 100vh`, which would not pin the height at all — should the pattern anchor `height:` so the weaker declaration cannot satisfy it? The offsets get the same treatment in reverse: the test asserts `inset:` is absent but never asserts `top/right/bottom/left: 0` are present, and the source comment calls those the first load-bearing declaration.

robertabbott (commented):
One issue inline (the empty label's colour does not survive a theme flip), the rest suggestions. The core change checks out: `pub_id` and `embed_url`/`image_url` are lifted as a set in all three chart tools, so the new `undefined` branch is unreachable on a chart-bearing result, and nothing outside `mcp.ts` reads the result-level `_meta.ui` keys.

PR #225, opened 2026-08-07
fix(chatgpt): the reloaded card kept its dimensions and lost its chart

25eliu:
Two independent ChatGPT problems, both about Tako losing a decision it should have won.

## 1. The card goes blank on reload

Reported: a `tako_answer` card renders correctly, then comes back as an empty box after reloading the conversation. The `structuredContent` degrades to `{"width":900,"height":720}`.

**The server was never at fault.** Verified against prod: a repeat `tako_answer` returns all ten declared keys, and `topCardChartFields` emits its six widget fields all-or-nothing, so no server path produces that pair alone. Those two values are `DEFAULT_WIDTH`/`DEFAULT_HEIGHT`.

**The actual bug** is that ChatGPT rehydrates with a stripped `toolOutput`, and a stripped object is still *truthy*. So `pickFromOpenAi`'s `a || b` chain selected it, handed it to `render()`, and the no-chart guard correctly concluded "no card" and collapsed. Truthiness was the selector where *carrying a chart* should have been.

- Both pick functions are now one `pickPayload(globals)` over one ordered candidate list — globals channels first, then `window.openai`, then the mirror. Selection is by four tiers: **first whole result in list order** (chart or not), then a chart-carrying fragment, then the mirror, then the first truthy candidate. That last tier matters — it keeps a genuinely chart-less result reaching `render()` for the labelled empty state rather than hanging the poll.
- Completeness outranks the chart preference, and that ordering is load-bearing rather than cosmetic. Preferring any chart-bearing candidate across channels lets a stale card in `widget.structuredContent` beat a fresh zero-card `toolOutput` — a chart for a question that returned nothing, which is the same harm the mirror is gated against.
- Selecting correctly cannot conjure back a payload ChatGPT dropped, so the widget mirrors the rendered card into widget state (`setWidgetState`, restored as `window.openai.widgetState`) and consults it **after** every live channel.
- The restore is gated on `looksComplete`, so it cannot overpaint a genuine "no results": a zero-card search still ships `cards: []`, the stripped payload ships no result-shaped key at all.

Claude never had this failure mode — it gets a per-call dynamic `ui/resourceUri` carrying the pub_id. ChatGPT reads a static template URI, so `toolOutput` was its only identity.

## 2. Web search fires before Tako, even when the user says "Tako"

Server instructions are the only surface that can influence that choice, since a tool description is read once the model is already looking at Tako's tools.

They said *"reach for Tako BEFORE a generic web search"*, which concedes the generic search is still a step to take and invites doing both. Now: *"instead of a separate web search, not alongside one"*, led by the capability that earns it — Tako searches the live web and the data graph in the same call.

Trimmed 1226 to 817 chars. Kept: the domain list (the only discovery trigger on a host with many servers connected) and "pick one, don't chain them" verbatim, the one line with a measured effect. Cut: two paragraphs restating what `tako_available_data` and `tako_contents` already say themselves.

`tako_available_data` also moves after the retrieval pair, reversing a past fix, and its "NOT a required first step" hedge is gone. The contradiction that motivated both is removed at the source. **Round 1 extended this:** the same permission lived in `skills/tako-financial-research/SKILL.md` and `skills/tako-macroeconomics/SKILL.md` in their own words, neither written against the server instructions — so "the hedge only existed to push back on this paragraph" was incomplete reasoning, and the skills lose it too. **Round 2 found it still standing** in `README.md`, which embeds all three skills whole for manual claude.ai upload. The parity test that should have caught that diffed frontmatter only; it now diffs the body too.

The ordering test is replaced by one asserting the invariant that actually caused the incident — not the order, but the **contradiction** between the two surfaces — as a pair rather than as two absences, so it permits the rollback the `mcp.ts` comment documents.

## Test plan

- [x] **1255 tests pass** (1085 + 102 widget + 68 scripts), 4 typecheck projects, `registry:check`, `schemas:check`
- [x] 10 widget tests drive the real bundle: state is mirrored; a stripped `toolOutput` repaints from the mirror on **both** the synchronous and the `set_globals` event path; a live result beats a stale mirror on both; a chart-less mirror does not satisfy the search; a whole zero-card result outranks a chart in a lesser channel; no duplicate state writes across two deliveries; a host without `setWidgetState` renders normally
- [x] **Mutation-checked**, since round 1 found a test that could not fail. Each of these turns exactly one test red and no others: delete `lastPersistedKey`; swap the completeness and chart tiers; restore the `||` short-circuit in the event handler. The instructions pair-assertion was checked across all four oblige/deny states — only "both" fails.
- [x] **Skill/README parity, mutation-checked too** (round 2): the hedge back in the README alone → 1 red; a SKILL.md body line reworded with the README left stale → 1 red; the one declared normalization removed → all 3 red. `tako-web-traffic` matching byte-for-byte under that substitution today is the evidence it is the only sanctioned delta.
- [ ] **Needs a human:** reload a live ChatGPT conversation containing a Tako card. jsdom cannot prove ChatGPT restores `widgetState`. Round 1 widened the coverage to the event path, which does not remove this need. If it still collapses, check whether `window.openai.widgetState` is populated at all in that frame — that separates "we never persisted" from "ChatGPT dropped the state too", which need different fixes.

Independent of #223; branched from `main`, no overlap with the native-card work.


workers/src/tools/_chart_widget.ts
```diff
@@ -1774,28 +1779,139 @@ const WIDGET_HTML = `<!doctype html>
...
+    for (i = 0; i < live.length; i++) {
+      if (looksComplete(live[i])) return picked;
+    }
+    return hasChart(w.openai.widgetState) ? w.openai.widgetState : picked;
   }
+
+  // Same best-first selection as \`pickFromOpenAi\`, and for the same reason:
+  // a stripped-but-truthy \`toolOutput\` must not mask a sibling channel that
+  // still carries the chart. No widget-state fallback here — this path fires
+  // on a live \`openai:set_globals\` event, never on rehydration.
```
robertabbott: Does the rehydration fix skip the path ChatGPT may actually deliver on? `pickFromGlobals` has no widget-state fallback, and the handler is `render(pickFromGlobals(globals) || pickFromOpenAi())` — so a stripped `toolOutput` arriving via `openai:set_globals` is truthy, wins `firstUsable`'s second pass, short-circuits the `||`, and collapses with `rendered = true` before the mirror is ever read. The comment justifies this as "never on rehydration", but the 250 ms poll below exists precisely because "ChatGPT populates the global at unpredictable times" — if that timing holds on reload too, the fix silently no-ops in the case it targets. What rules out the event path here?

workers/src/tools/_chart_widget.ts
```diff
@@ -1774,28 +1779,139 @@ const WIDGET_HTML = `<!doctype html>
...
+  // truthy object, so a plain \`a || b\` chain picks it, hands it to render(),
+  // and collapses — which IS the reported bug, with a healthier channel
+  // sitting unread further down. Prefer a candidate carrying a chart; fall
+  // back to the first truthy one so a genuinely chart-less result still
+  // reaches render() and gets the labelled empty state, rather than being
+  // mistaken for "nothing has arrived yet" and hanging the poll.
+  function firstUsable(candidates) {
+    var i;
+    for (i = 0; i < candidates.length; i++) {
+      if (hasChart(candidates[i])) return candidates[i];
```
robertabbott, line 1910: Does the `looksComplete` veto guard only half the selection? The chart-preference loop scans every candidate before list order applies, so a lower-priority channel carrying a chart now outranks a fresh chart-less `toolOutput` — the reverse of the `||` chain it replaces. A genuine zero-card result next to a stale `widget.structuredContent` selects the stale one, `hasChart(picked)` returns at 1890, and `looksComplete` never runs. That is the same "chart for a question that returned nothing" the veto exists to stop, reached by the other route. The zero-card test only populates `toolOutput`, so it cannot see this. Should a `looksComplete` candidate win before the chart preference applies?

workers/src/mcp.test.ts
```diff
@@ -1349,13 +1349,24 @@ describe("SERVER_INSTRUCTIONS", () => {
...
+  it("does not oblige a free-tool call the tool's own description denies", () => {
     const free = SERVER_INSTRUCTIONS.indexOf("`tako_available_data`");
-    const answer = SERVER_INSTRUCTIONS.indexOf("`tako_answer`");
-    const search = SERVER_INSTRUCTIONS.indexOf("`tako_search`");
     expect(free).toBeGreaterThan(-1);
-    expect(free).toBeLessThan(answer);
-    expect(free).toBeLessThan(search);
+    // No sequencing verb at server level: "first", "before", "start with".
+    expect(SERVER_INSTRUCTIONS).not.toMatch(
+      /\b(first|before|start with|begin with)\b/i,
```
robertabbott: Does this test pin the resolution rather than the invariant it names? The regex forbids "first"/"before" anywhere in `SERVER_INSTRUCTIONS` — including "ask it for a measure's exact name **before** spending a priced call", the wording this file shipped one commit ago. The comment added in `mcp.ts` names promoting that paragraph as the first thing to retry if search-first routing returns, and that rollback wants the hedge restored alongside it; both assertions block it. Would asserting the pair — an obligation implies no contradicting hedge — survive the rollback the comment documents?

workers/src/tools/tako_available_data.ts
```diff
@@ -88,7 +88,7 @@ const NER_LABELS = [
 const DESCRIPTION = [
   "Find what proprietary, continuously-updated structured data exists on something — summarized in one call. Free and fast.",
   "",
-  "Ask it when the question IS coverage: what does Tako have on X, is this measure tracked at all, what is it called. Then build the real question around what comes back. NOT a required first step — for a straightforward data question, tako_search or tako_answer directly is usually right.",
+  "Ask it when the question IS coverage: what does Tako have on X, is this measure tracked at all, what is it called. Then build the real question around what comes back.",
```
robertabbott, line 91: Was this hedge only there to contradict `SERVER_INSTRUCTIONS`? Two bundled skills still carry the same guidance in their own words: `skills/tako-financial-research/SKILL.md:25` ("Not a warm-up before every lookup. Go straight to `tako_answer` when you just want the number") and `skills/tako-macroeconomics/SKILL.md:23`. Neither was written against the server instructions, so the permission to skip the free tool looks load-bearing on its own. With it gone here, "Worth one call first when you need a measure's EXACT name" is the only routing signal left in the description every client sees, and the two surfaces now disagree. Should the skills lose it in this PR, or should this line stay?

workers/test/widget/chatgpt-path.test.ts
```diff
@@ -243,3 +243,105 @@ describe("ChatGPT Apps SDK delivery paths", () => {
...
+    const m = mountWidget(html(), {
+      toolOutput: STRIPPED_TOOL_OUTPUT,
+      widgetState: STRIPPED_TOOL_OUTPUT,
+      notifyIntrinsicHeight: (h: number) => heights.push(h),
+    });
+    expect(renderedIframe(m)).toBe(false);
+    expect(heights).toEqual([1, 0]);
+  });
+
+  it("does not rewrite state for the same card", () => {
```
robertabbott: Can this test fail if `lastPersistedKey` is deleted? Both `fireSetGlobals` calls carry a chart, so the first render latches `rendered = true` at `_chart_widget.ts:1770` and the second returns at the `if (rendered)` guard — the write-once key is never consulted. The guard does have a reachable path (`persistWidgetState` runs at 1622, above the `return false` at 1762, so a payload that passes the no-chart guard but has no valid URL re-persists on every poll tick), but this test does not take it. Would a payload that reaches that branch, driven across two poll ticks, make the assertion load-bearing?

workers/src/tools/_chart_widget.ts
25eliu: Nothing rules it out, and you're right that the comment was asserting something we can't check. Fixed in f5c1509.

The `||` is gone. `pickFromGlobals` and `pickFromOpenAi` are now one `pickPayload(globals)` over one ordered candidate list — globals channels first (an event just delivered them), then the `window.openai` channels, then the mirror. The event handler is `render(pickPayload(globals))` and the poll is `render(pickPayload(null))`, so the same precedence applies no matter which channel the payload came through and the question of whether `set_globals` fires on rehydration doesn't have to be answered.

New test `restores the mirror when the stripped payload arrives on the EVENT path`: mount with an empty `window.openai` (so the synchronous probe can't be what renders), populate `widgetState`, then fire `set_globals` carrying the stripped `{width, height}`. Red before, green after. Its converse, `a live chart on the event path still beats the mirror`, guards the regression unification could plausibly introduce.

workers/src/tools/_chart_widget.ts
25eliu: Confirmed, and the fix had to go further than "check `looksComplete` first" — thanks for the precise framing. f5c1509.

Trying your suggestion literally (a `looksComplete && hasChart` tier ahead of a `looksComplete` tier) still failed the test I wrote for your scenario. The stale `widget.structuredContent` is a whole prod payload, so it carries its own `cards` key and is every bit as complete as the fresh zero-card `toolOutput` ahead of it — the combined tier just reached across and picked it again.

So completeness decides alone, in list order, and the chart preference never crosses a whole result:

1. first `looksComplete`, chart or not — `toolOutput` is canonical, the rest are compatibility fallbacks, so order is the only freshness proxy available
2. first `hasChart` — only when nothing whole arrived at all; this is the tier that steps over the stripped `{width, height}`
3. the mirror
4. first truthy

New test `a whole zero-card result outranks a chart in a lesser channel` populates both channels, which as you note the zero-card test could not see. Mutation-checked: swapping tiers 1 and 2 turns it red and nothing else.

workers/src/mcp.test.ts
25eliu: Right on both counts — it pinned the resolution, and it blocked the rollback its own neighbour documents. Rewritten as the pair in 9e5f0af.

One correction to the framing, which the mutation runs settled: the rollback does *not* want the hedge restored alongside it. Promoting the paragraph restores the obligation ("ask it for a measure's exact name BEFORE spending a priced call"); restoring the hedge too would recreate exactly the oppositional state that caused the incident. Obliging a call that nothing contradicts is coherent — it is the position this file held one commit ago. So the pair form permits the rollback and still catches the incident.

The sequencing check is now scoped to the line naming the tool (unscoped, it fires on any "before" anywhere), and the denial side matches the family of wordings rather than the single phrase this PR removed.

Four states, each checked by mutation:

| instructions oblige | description denies | result |
|---|---|---|
| no | no | pass — today |
| **yes** | no | pass — the rollback. The old assertion failed here |
| no | **yes** | pass |
| **yes** | **yes** | **fail** — the incident |

workers/src/tools/tako_available_data.ts
25eliu: No — you're right that the reasoning was incomplete, and the two skills prove it. Resolved the other way, in 9e5f0af: the skills lose it.

The call came down to reach. The tool description is read by every client; the skills only reach clients that bundle them. Keeping the permission in the narrow surface and dropping it from the broad one is the disagreement you identified, just inverted, so the fix was to stop stating it in the places that were never the surface of record for it.

- `skills/tako-financial-research/SKILL.md:25` — dropped "Not a warm-up before every lookup. Go straight to `tako_answer` when you just want the number."
- `skills/tako-macroeconomics/SKILL.md:23` — dropped "Not a warm-up before every lookup."

Both keep everything that is skill-specific and measured: entity ambiguity surfacing on Costco, PCE requiring an exact indicator name. What's left across all three surfaces is one consistent statement — the free tool answers a coverage question, and it is worth a call when you need an exact name.

The premise you falsified is corrected in the `mcp.ts` comment rather than left standing, since "it only existed to push back on this paragraph" is what a future reader would otherwise use to justify the next removal.

workers/test/widget/chatgpt-path.test.ts
25eliu: It could not — confirmed by deleting `lastPersistedKey` and watching it stay green. Fixed in f5c1509, and I took exactly the branch you identified.

The payload is a whole prod result with its two URLs swapped to `ftp://`. Strings, so the no-chart guard lets it through and `persistWidgetState` runs at 1622; then they fail `^https?://` on every render branch, so render() falls to `return false` at 1762 with `rendered` still unset and the second delivery re-enters. Two `set_globals` events, one write asserted.

Renamed to `does not rewrite state for the same card, across two deliveries`, with a comment recording why a healthy payload cannot test this — otherwise the obvious "simplification" is to swap the fixture back.

Mutation-checked alongside the other two guards: deleting the write-once key turns this test red and only this test.

skills/tako-financial-research/SKILL.md
```diff
@@ -22,7 +22,7 @@ Reliably NOT in the data graph, and where web carries the answer instead: forwar
...
 - `tako_search`: reach for it when you want **breadth or a chart**, not a number: fanning out across several companies or metrics to see what exists, or pulling the card when the chart itself is the deliverable. Its cards are captions — S&P Global and Visible Alpha rows are license-gated, so a search that returns the right card still owes you a `tako_answer` call for the value. One card renders inline (see Pick the right card).
-- `tako_available_data`: FREE, and the right tool when the question is **what Tako covers**: does this metric exist, under what exact name, for which entity. It also surfaces entity ambiguity early (`"Costco"` matches both Costco Wholesale Corporation and Costco Wholesale Australia). Not a warm-up before every lookup. Go straight to `tako_answer` when you just want the number.
+- `tako_available_data`: FREE, and the right tool when the question is **what Tako covers**: does this metric exist, under what exact name, for which entity. It also surfaces entity ambiguity early (`"Costco"` matches both Costco Wholesale Corporation and Costco Wholesale Australia).
```
robertabbott, line 25: Does the hedge survive in a fourth surface this PR did not check? `README.md:397` and `README.md:573` carry these two skill bullets verbatim, both still ending "Not a warm-up before every lookup" — 397 also keeps "Go straight to `tako_answer` when you just want the number." So the disagreement moved rather than closed, and the `mcp.ts` comment's "Three surfaces now say the same thing about the free tool" is not true as written. CI cannot catch it: `workers/scripts/skills.test.ts:105` matches the README block on `^---\nname: X\n…\n---$`, so it diffs frontmatter only and never the body — and its own docblock calls SKILL.md↔README drift the proven-to-recur failure it exists to stop. Should the README bullets go in this PR, and should the parity check extend past the frontmatter?

robertabbott (commented):
Re-reviewed the delta (`1762d9ea..9e5f0af`). All four logic and test findings are resolved, and two of them are resolved better than I proposed — the pick unification removes the unverifiable `set_globals` premise rather than answering it, and you were right that a `looksComplete && hasChart` tier still reaches across a whole zero-card result. Re-traced `selectPayload` against the reload, zero-card, stale-sibling and stripped-mirror cases; each lands where the comments claim.

One new finding below, from the fix to the hedge thread.

skills/tako-financial-research/SKILL.md
25eliu: Yes to both, and the `mcp.ts` "three surfaces" line was wrong the moment I wrote it — the README copies were never in that count. 85de1e0.

**The bullets.** Both gone. A repo-wide sweep confirms `README.md:397` and `:573` were the last of it; nothing in `llms-full.txt`, `docs/`, or the submission JSON carries it.

**The parity check.** Extended to the whole file, which turned out to be cleanly possible: I diffed each README block against its SKILL.md with correct boundaries, and there is exactly **one** sanctioned delta, uniform across all three skills —

| | SKILL.md | README |
|---|---|---|
| | ``…live on the Tako MCP server (server name `tako`).`` | `…live on the Tako MCP server installed in Step 1.` |

— because the README block is step 2 of a numbered install flow and the server name is introduced two steps later. That is declared as a named constant rather than fuzzed over, so a second intentional delta has to be written down where the next reader sees it. The evidence that it is the *only* one: `tako-web-traffic` matches byte-for-byte under that single substitution today, with no other change.

Mutation-checked, since round 1 was about a test that could not fail:

- hedge reintroduced in README alone → 1 red (`tako-macroeconomics`)
- a SKILL.md body heading reworded, README left stale → 1 red
- the normalization removed → all 3 red

**The comment.** It no longer counts surfaces. A count is not the invariant and prose cannot hold it, so it now records that an earlier version claimed "three" and was wrong, and points at the test instead.

Also: thanks for re-tracing `selectPayload` against the four cases rather than taking the comments' word for it.

robertabbott (approved):

PR #238, opened 2026-08-13
feat(widget): opt every displayed card into the embed page's share control

25eliu:
## What this does

Tako's embed page grew an opt-in share control in TakoData/tako#28735 — a trigger in the card's top-right action cluster opening a dialog with copyable embed/card links. It shipped inert: nothing renders unless the embedding URL carries `?showShare=true`, and that PR's own words were "Enabling it is one line in TakoData/tako-mcp." This is that line, plus the clipboard delegation the dialog needs.

- `buildChartUrls` now appends `&showShare=true` to `embed_url`. It is the choke point for every URL the widget renders: search/answer top-card widget fields, `tako_visualize`, the baked per-pub_id widget resource, and the probe/commit iframe. Per-card `embed_url`s inside the `cards` array (markdown listings, structured output) are backend passthrough and unchanged — only the rendered card opts in. `withHostTheme` rewrites only the `dark_mode=` segment, so the param survives theme rewrites.
- The widget's embed iframe gains `allow="fullscreen; clipboard-write"`. The share dialog copies via `navigator.clipboard.writeText`, which needs every ancestor frame to delegate `clipboard-write`; this is the one hop we own.

## What stays PNG / share-less, by design

The embed page hard-suppresses the control even when opted in — auth-gated cards (both links would 401 for a recipient), glanceable tiles, white-label embeds, `?screenshot=true` renders, and data-less postMessage shells. The native-card `document.write` path (claude.ai's interactive card) cannot show it *yet*: every gate (`isShareShown`, `getCardPubId`) and the dialog's URL builder (`canonicalOrigin`) reads `window.location`, which in that path is the widget sandbox URL, not `/embed/<id>/`. This PR already carries the opt-in on the proxy's upstream fetch — inert until a small tako-repo follow-up bakes the param, the pub_id, and the canonical origin into `json_script` islands the gates can fall back on. Once that tako PR lands, the share CONTROL lights up on claude.ai with no further change here; the copy action there depends on claude.ai's own sandbox delegating `clipboard-write` (the widget's iframe `allow` doesn't apply to the document.write path — there is no nested frame), with the dialog's select-and-copy fallback otherwise.

## Verified live

- Prod already serves the share machinery (the `card-require-auth` json_script is in the live embed HTML), so this is not gated on a backend deploy.
- Headless check on a real prod card: with `?showShare=true` a "Share this card" button renders and the dialog offers the embed and card links; without the param, zero share elements.
- End-to-end on chatgpt.com through a dev tunnel: interactive card renders with the share control; both links present and correct.
- Clipboard: ChatGPT's sandbox does not delegate `clipboard-write`, so the dialog lands on its designed fallback — the URL pre-selects and the button shows the ⌘C hint. That rejection happens at OpenAI's hop; ours no longer blocks anything.

## Tests

`buildChartUrls` gets direct unit coverage (exact `embed_url`/`image_url` shapes, param kept off the PNG url), the widget HTML is asserted to carry the `clipboard-write` delegation, and the search-tool integration expectation pins the full URL. All three vitest configs pass (37 + 3 + 4 files), typecheck clean across all four tsconfig projects.

## Test plan

- [ ] CI green
- [ ] After staging deploy: ask claude.ai (staging connector) for a chart → interactive card shows "Share this card"; dialog links open the right card
- [ ] Confirm a glanceable single-answer card still renders without the control
- [ ] Spot-check an `embed_url` from tool text output opens with the share control in a plain browser tab

### DIFF
--- a/docs/chatgpt-app-review.md
+++ b/docs/chatgpt-app-review.md
@@ -115,13 +115,18 @@ the declaration and the URL cannot drift. No wildcards, no third-party frames, o
 (`/embed/{pub_id}/`).
 
 **What the framed page contains** (measured on production, `pub_id`
-`VKd7qE8K9Ba16kMFENNQ`, 2026-08-04): the chart, its title, and source attribution. Three
-external origins are referenced — Tako's own asset CDN (`d12w4pyrrczi5e.cloudfront.net`,
-16 refs: `Card.js`, its chunks, fonts, card images), `www.spglobal.com` (4 refs, all inside
-the card's own JSON island as the `source_details.url` for an S&P-sourced series), and —
-before this change — `www.googletagmanager.com`. There are **no advertisements, no checkout,
-no signup, no sign-in, and no upgrade UI**: the served markup contains zero `<a>` elements,
-and no occurrence of "sign up", "log in", "upgrade", "pricing", "subscribe" or "checkout".
+`VKd7qE8K9Ba16kMFENNQ`, 2026-08-04; re-measured 2026-08-13 with `showShare=true`): the
+chart, its title, source attribution, and — since the share opt-in — a share control in the
+card's action cluster, opening a dialog of copyable Tako URLs for this same card (copy
+fields, not navigation; the control mounts client-side and is Tako PR #28735's opt-in
+surface). Three external origins are referenced — Tako's own asset CDN
+(`d12w4pyrrczi5e.cloudfront.net`, 16 refs: `Card.js`, its chunks, fonts, card images),
+`www.spglobal.com` (4 refs, all inside the card's own JSON island as the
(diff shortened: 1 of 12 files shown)


robertabbott (commented):
Reviewed the share opt-in, the clipboard delegation, and the origin pin. Findings inline. The PR body's load-bearing claims all check out: `withHostTheme`'s `dark_mode=[^&]*` stops at the `&` so the param survives, the pin is substituted on the single path that serves the bundle, `WIDGET_HTML` has exactly one placeholder, both widget harnesses run the substituted bundle so the PIN test passes for the right reason, and nothing string-compares `embed_url`.

robertabbott (approved):

PR #239, opened 2026-08-13
feat(mcp): recognize the ChatGPT desktop app (codex client) as ChatGPT family

25eliu:
## Summary

The merged ChatGPT desktop app connects through its Codex runtime as `codex-mcp-client/<version>` and fell to the `unknown` client surface: 3 free-tier tools, no `securitySchemes`, no reauth flow. This PR makes the desktop app a first-class **ChatGPT-family** client AND (since #240 merged into this branch) flips it to the same interactive chart widget as chatgpt.com:

- New `"codex"` client kind (`detectMcpClient`) + `isChatGptFamilyClient` predicate
- Codex gets the ChatGPT product surface: split agent pair, anonymous discoverable listing, `securitySchemes` injection (with the real client threaded through to the rollout log), the Apps-SDK reauth challenge, and family-routed `descriptionByClient`
- **Widget flip (#240)**: codex joins `isWidgetClient` — widget `_meta` on the three chart tools, default-on `tako_visualize`, dimensions-only PNG read (`bakeImage` gated on the family). Its hard dependency, TakoData/tako#29218's `frame-ancestors codex-sandbox:`, is verified live on production tako.com (2026-08-14).

## History: the flip was staged, then landed

A full pre-landing review measured the flip build live: a widget-classified codex tool call returns `[text]` only (widget `_meta` and image blocks are mutually exclusive), so the flip originally shipped as a separate draft (#240) gated on the production CSP deploy. That deploy is live and #240 merged into this branch on 2026-08-14 after local end-to-end verification (see the verification comment below).

**Known, accepted costs** (tracked as follow-ups): a flag-off (`enable_mcp_apps = false`) desktop app and the headless Codex CLI share this UA and lose the inline PNG for widget-classified calls. The CLI's `notifications/progress` handling (which decides whether the agent split is right for it) has not been measured — follow-up filed from the round-2 review.

## Review

**Round 1** (pre-landing, 3 specialists + adversarial passes): security clean — UA spoofing widens only what a connection *sees*, never what it can *execute* (tier-keyed dispatch gate verified); 5 revert-resistance gaps closed; 12 stale comments fixed.

**Round 2** (robertabbott, all 8 comments addressed in `ac133b1`): `client` param on the schemes adapter now required; `descriptionByClient`/`annotationsByClient` key types narrowed to the two annotation families (dead `codex:`/`unknown:` keys are now compile errors, docstring fixed); `assertChatgptSubmissionParity` asserts the codex surface equals chatgpt's at both tiers; codex `bakeImage` pins added for answer + visualize; anonymous codex `tools/list` pinned against unknown's free-tier trio; widget/family membership is a per-kind expected table.

## Safety

Mergeable and deployable now — the CSP dependency is live on production. Non-codex clients: no behavior change today (`descriptionByClient` overrides would also reach `unknown` clients by design — documented submission-consistency asymmetry — but no tool defines one yet). Codex: gains the full family surface + interactive widget; the flag-off/CLI PNG trade-off above is the one regression, accepted deliberately.

## Test plan

- [x] `npm run typecheck` (all four passes)
- [x] `npm test` — 1144 + 120 + 68, all green (post-rebase onto main at 0.20.0, post round-2 review)
- [x] `registry:check` / `schemas:check` green (now includes the codex↔chatgpt surface-parity assertion)
- [x] Live validation against the desktop app (0.147 + 0.148): list/call/schemes/reauth surfaces verified
- [x] Local worker verified as `codex-mcp-client`: initialize + `tools/list` serve the 5-tool family surface with widget `_meta` on the three chart tools
- [x] Production CSP verified: `curl -sI https://tako.com/embed/<pub_id>/` carries `frame-ancestors … codex-sandbox:`

### DIFF
--- a/workers/scripts/gen-registry.ts
+++ b/workers/scripts/gen-registry.ts
@@ -290,6 +290,35 @@ export function assertChatgptSubmissionParity(
     }
   }
 
+  // The submission covers the ChatGPT PRODUCT, and the product has two MCP
+  // transports: chatgpt.com's connector AND the desktop app (detected as
+  // "codex"). Every surface equality above is asserted against "chatgpt"
+  // only, so without this block a revert of codex's family/widget
+  // membership (e.g. dropping it from `isChatGptFamilyClient`) would leave
+  // `registry:check` green while the desktop app lists a DIFFERENT tool set
+  // than the submission declares — review finding on PR #239. Equality of
+  // the two members' surfaces at both tiers is exactly "the family shares
+  // one surface", which is the premise the submission rests on.
+  for (const tier of ["authenticated", "free"] as const) {
+    const chatgptSurface = tools
+      .filter((t) => isToolOnSurface(t.name, "chatgpt", noOptIns, tier))
+      .map((t) => t.name)
+      .sort();
+    const codexSurface = tools
+      .filter((t) => isToolOnSurface(t.name, "codex", noOptIns, tier))
+      .map((t) => t.name)
+      .sort();
+    if (JSON.stringify(codexSurface) !== JSON.stringify(chatgptSurface)) {
+      problems.push(
+        `codex (ChatGPT desktop app) ${tier} surface [${codexSurface.join(
+          ", ",
+        )}] diverges from chatgpt's [${chatgptSurface.join(
+          ", ",
+        )}] — the submission describes the ChatGPT product, which includes the desktop app; restore family membership in workers/src/tools/_surface.ts`,
+      );
+    }
+  }
+
   const HINT_KEYS = ["readOnlyHint", "openWorldHint", "destructiveHint"] as const;
   for (const [name, resolved] of expected) {
     const declared = declaredTools[name];
--- a/workers/src/index.test.ts
+++ b/workers/src/index.test.ts
@@ -468,16 +468,13 @@ describe("worker routing", () => {
     }
   });
 
-  it("POST /mcp tools/list serves the codex desktop app the family surface, without the widget", async () => {
-    // End-to-end pin for the ChatGPT desktop app (codex runtime): it gets
-    // the ChatGPT-family treatment — securitySchemes injection included —
-    // but NOT the widget `_meta` and NOT default-on tako_visualize, both
-    // of which are gated on `isWidgetClient` until tako.com's
-    // `frame-ancestors` allows `codex-sandbox:` AND the flip is validated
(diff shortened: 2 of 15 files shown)


robertabbott (commented):
Two claims in the description do not hold as stated. "Every client's behavior is unchanged except codex" — routing `descriptionByClient` through `annotationClientFamily` also retargets every `unknown` client (Cursor, Windsurf, Claude Code) the first time a tool adds a `chatgpt:` override. "Codex strictly gains" — an opted-in headless Codex CLI loses the single-call `tako_agent` for the 40s-ceiling split pair, and that same-UA problem is the one the chart PNG got a staged rollout and a dedicated test for.

The security surface is clean: the anonymous widening is listing-only (the dispatch gate at `mcp.ts:1098` is client-independent and runs before the handler, tier derives from the `Authorization` header before `client` is computed, and `noauth` only reaches the three genuinely-anonymous tools).

Also: main is 7 commits ahead of the merge base, so this needs a rebase before merge.

robertabbott (approved):

robertabbott (commented):
Round-2 follow-up, post-merge — all 8 round-1 comments verified as addressed. The 7 code fixes hold under independent checking: the schemes `client` param is genuinely required with all 5 test sites updated, the `"claude" | "chatgpt"` key narrowing is type-safe with no casts introduced, the new codex parity block is non-vacuous at both tiers, the `bakeImage` pins fail on revert via `seen[0].range`, the direct-dispatch free-tier test bypasses no middleware (`index.ts:33` routes straight to `handleMcpRequest`), and the `EXPECTED_MEMBERSHIP` table forces both memberships while being order-independent. The CLI agent-split answer is a reasonable product call and the follow-up is the right shape.

The flip's hard dependency also checks out: tako#29218 is merged and `codex-sandbox:` is present on `origin/production`. I did not relitigate the flag-off/CLI PNG loss — it is disclosed as an accepted cost.

Four new findings inline, from the reapplied flip. #1 is the one I would fix soonest: it affects incident response for code already live in production.

Separately, and not caused by this PR: `workers-smoke` failed on main right after this deploy, but it has been red since at least 2026-08-11 with the identical `tako_visualize returned no card_id` failure and the identical `tako_search._meta.ui.resourceUri missing` warn — including on 0.19.5, before this branch existed. So a widget-contract change shipped to production behind a post-deploy gate that can no longer signal anything. Worth a fix independent of this PR.

PR #244, opened 2026-08-17
feat(mcp): actionable failure guidance for out-of-credits and auth-gated tools

25eliu:
## What this changes

Three failure modes that used to dead-end the calling model now answer with guidance it can act on. Authenticated instructions stay **byte-identical** — existing integrations see zero change until something actually fails.

### 1. Authenticated out-of-credits (Django 402) → curated `payment_required` result

**Scenario — before:** A Cursor user with their own Tako API key runs out of credits mid-research. Their agent calls `tako_search("NVDA revenue vs AMD")` and gets back:

> `Django returned 402 for POST /api/v1/knowledge_search/: {"error_type": "PAYMENT_REQUIRED"}`

The model can't tell "add credits" from "server bug" — it typically retries a few times (burning turns), then tells the user "Tako appears to be down."

**After:** the same call returns an `isError` tool result the model can relay:

> This Tako account is out of credits, so the call could not run. Retrying without adding credits will fail again. `tako_available_data` is free and still works. *(+ on positively-identified Claude clients only: "Add credits to the account on tako.com to continue.")*

with `_meta["tako/error"] = { kind: "payment_required", status: 402, ... }` for clients that branch on structured detail. ChatGPT-family **and unknown** clients get the same message *without* the tako.com pointer — OpenAI commerce policy forbids promoting purchases through an app, and `unknown` fails closed for the same reason `annotationClientFamily` buckets it with chatgpt (an OpenAI crawler, or a ChatGPT UA `detectMcpClient` hasn't learned yet — the desktop app's `codex-mcp-client` UA was exactly that once).

The free tier's 402 masking (`"temporarily out of shared capacity"`) is untouched — an anonymous caller has no account to top up, so naming credits there would only send them hunting a fault in their own setup.

### 2. Anonymous connections get honest `initialize` instructions

**Scenario — before:** An anonymous Claude Desktop connection receives server instructions saying "`tako_contents` reads one source in full…" — but the anonymous surface *hides* that tool entirely. The system prompt advertises a tool the model cannot see; when it calls it anyway, the SDK answers "tool not found," which reads as a broken server.

**After:** anonymous connections get `FREE_TIER_SERVER_INSTRUCTIONS`, identical except the last paragraph:

> This connection is anonymous: `tako_available_data`, `tako_search`, and `tako_answer` are the tools that run here. Other tools — like `tako_contents`, which reads one source in full (…) — become available on a connection authenticated with a Tako account.

("the tools that *run*", not "the full toolset" — on anonymous ChatGPT two gated tools stay listed for the link-account UI, so an executable-set claim is what's true on every client.) The shared paragraphs are spread from one array so the tiers can't drift; a test pins authenticated `getInstructions()` `===` the pre-PR string.

### 3. Anonymous calls to known-but-gated tools answer sign-in guidance, not "tool not found"

**Scenario — before:** An anonymous Claude Code user asks "export the rows behind that CPI chart." The agent calls `tako_contents` (it knows the name from the instructions or from an earlier authenticated session) and gets the SDK's bare *"tool not found"* — a dead end claiming the tool doesn't exist.

**After:** a pre-SDK gate in `handleMcpRequest` answers with the same result ChatGPT's dispatch gate already used:

> This tool requires a linked Tako account. Sign in with Tako to continue.

carrying the `_meta["mcp/www_authenticate"]` challenge (what ChatGPT's link UI keys on; other hosts ignore unknown `_meta`). The model relays a true, actionable answer, and on ChatGPT the sign-in flow can trigger.

**The promise is true by construction.** The gate answers only when the caller's *authenticated* surface (same client, same `?tools=` opt-ins) would actually carry the tool:

| Anonymous call | Answer | Why |
|---|---|---|
| `tako_contents` | sign-in guidance | auth alone unlocks it |
| `tako_agent` (no `?tools=agent`) | SDK "tool not found" | sign-in would NOT unlock it — promising auth just moves the dead end one sign-in later |
| `tako_agent` via `…/mcp?tools=agent` | sign-in guidance | auth on this URL unlocks it |
| `tako_typo_tool` | SDK "tool not found" | claiming a nonexistent tool needs auth sends the caller on a pointless sign-in |

Security-invariant notes: nothing widens what anonymous connections can **execute** (`registerTool`'s dispatch gate remains the barrier, defense-in-depth); gated calls still burn no per-IP quota and never reach Django; known-vs-unknown name probing discloses nothing (`registry/server.json` and anonymous `tools/list` already publish the list).

## Review

Ran a full independent code review before opening this PR (byte-comparison of instructions vs main, full-file reads, independent test/typecheck runs). Verdict: no Critical issues; both Important findings are fixed in the second commit — (a) the gate originally keyed on raw registry membership, which made the sign-in promise false for 7 of 9 gated names (opt-in/client-scoped tools); (b) the top-up pointer originally failed *open* for `unknown` clients, repeating the exact misclassification that once hid the ChatGPT desktop app. Minor findings (jsonrpc envelope check, client in the gate's log line, instructions phrasing on ChatGPT, e2e initialize coverage) are folded in.

## Test plan

- [x] `npm test` — 1154 + 120 + 68 pass, including 11 new/updated pins:
  - authenticated 402 → curated message; pointer on claude e2e, absent on chatgpt **and** unknown e2e; `kind: payment_required` + status in `_meta`
  - hidden-tool gate e2e: sign-in guidance (no per-IP burn, Django never called), typo fall-through, opt-in-without-flag fall-through, opt-in-with-`?tools=` guidance
  - anonymous `initialize` returns the free variant over real HTTP; authenticated `getInstructions()` `===` pre-PR string; tier variants share all but the last paragraph
  - free-tier 402 masking + "PAYMENT_REQUIRED text in a 403 body is not credit exhaustion" unchanged
- [x] 4-pass `npm run typecheck`
- [x] `npm run registry:check` (deploy gate) — ok, 12 tools
- [ ] Post-deploy: `wrangler tail` grep for `layer=pre-dispatch` to watch how often anonymous callers hit gated tools in the wild


workers/src/mcp.ts
```diff
@@ -1443,6 +1483,89 @@ export function djangoErrorToToolResult(err: DjangoError): {
...
+ * no way to tell "add credits" from "server bug". Unlike the free tier's
+ * `FREE_TIER_CREDITS_MESSAGE` (which deliberately masks the cause, because
+ * an anonymous caller has no account to top up), an authenticated caller
+ * OWNS the exhausted balance, so naming the cause is the actionable answer.
+ *
+ * The base message states the cause, the retry semantics (retrying without
+ * adding credits fails again — stops the agent burning turns), and the
+ * surviving free move (`tako_available_data`).
+ */
+export const PAYMENT_REQUIRED_MESSAGE =
```
robertabbott: Does "add credits" match what a FREE or PRO subscriber can actually do? The 402 on `/api/v1/knowledge_search/` and `/api/v1/answer/` comes from `SubscriptionCreditThrottle`, whose body is `{"error":"insufficient_credits","message":"…Upgrade your plan for more credits.","upgrade_url":"/pricing"}` — it drains `Subscription.monthly_credits_remaining` (FREE 1200/mo, PRO 7500/mo), which regrants each cycle. The purchasable `ApiCreditAccount` pool is enterprise-scoped and is not what `check_and_deduct_credits` draws from. So for most callers there is no credit to add, and "Retrying without adding credits will fail again" is unconditional across a monthly reset. Should the text mirror Django's own remedy? (The tests fixture `{"error_type":"PAYMENT_REQUIRED"}`, which neither of these two endpoints emits.)

workers/src/mcp.ts
```diff
@@ -1443,6 +1483,89 @@ export function djangoErrorToToolResult(err: DjangoError): {
...
+ * `detectMcpClient` has not learned yet lands on `unknown` (the desktop
+ * app's `codex-mcp-client` UA did exactly that before its rule was
+ * added), and serving it commerce copy is an app-rejection risk. A
+ * generic third-party client merely missing the pointer is cosmetic — the
+ * base message already names the cause.
+ *
+ * Bare domain, no path: deep links rot (the `/account/` path the old
+ * free-tier upsell used was already stale when it was removed).
+ */
+export const PAYMENT_REQUIRED_TOPUP_SUFFIX =
```
robertabbott: Claude Code lands in `unknown` and so never gets this pointer — is that intended? `detectMcpClient` maps `claude-code/<version>` to `"unknown"` at mcp.ts:308, deliberately, so it keeps the inline-PNG path rather than widget `_meta`. That makes the flagship authenticated raw-Bearer client bucketed here with OpenAI crawlers. The gating reuses a widget-capability signal as a commerce-policy proxy, and the two populations differ. Would keying the suffix on an explicit non-OpenAI allowlist (`claude` + `claude-code`) rather than `client === "claude"` hold the policy line while reaching the client that most needs it?

workers/src/mcp.ts
```diff
@@ -188,14 +188,51 @@ export const SERVER_VERSION = "0.20.0"; // x-release-please-version
...
+ */
+export const FREE_TIER_SERVER_INSTRUCTIONS = [
+  ...SHARED_INSTRUCTION_PARAGRAPHS,
+  "",
+  // "the tools that run", not "the full toolset": on ChatGPT-family
+  // anonymous connections two auth-required tools stay LISTED for the
+  // link-account UI (see `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES`), so
+  // a toolset-count claim would be false there; what's true on every
+  // client is which tools EXECUTE.
+  "`tako_available_data` is free, and answers what data Tako has on an entity or a metric, including a measure's exact name. This connection is anonymous: `tako_available_data`, `tako_search`, and `tako_answer` are the tools that run here. Other tools — like `tako_contents`, which reads one source in full (an exportable card's rows, or a web page's text by url) — become available on a connection authenticated with a Tako account.",
```
robertabbott: Nothing pins this hardcoded three-tool sentence to `FREE_TIER_TOOL_NAMES` — adding a fourth free tool, or dropping `tako_answer`, leaves every test green while `initialize` ships a false statement of what runs, injected above the tool descriptions in the host system prompt. The new guard at mcp.test.ts:1491 only asserts `toContain` for four names. The repo pins this class of drift elsewhere (`freetier.test.ts:88`, `assertChatgptSubmissionParity`). Worth asserting the sentence names each `FREE_TIER_TOOL_NAMES` member and no other registry tool besides the `tako_contents` teaser?

workers/src/mcp.ts
```diff
@@ -1535,8 +1756,23 @@ export async function handleMcpRequest(
...
       case "limited":
         return freeTierLimitResponse(meterResult.requestId);
-      case "allowed":
+      case "allowed": {
+        // Known-but-auth-required tool called anonymously: answer with
+        // sign-in guidance instead of letting the SDK say "tool not
+        // found" (see `freeTierHiddenToolResponse`). Client detection and
+        // `?tools=` parsing are re-run later on the SDK path with logging;
+        // both are pure, so the double call cannot disagree.
+        const gated = freeTierHiddenToolResponse(
```
robertabbott: With this gate in front, is `registerTool`'s free-tier dispatch gate (mcp.ts:1137) still reachable over HTTP? Both `CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES` members pass every guard here — in the registry, outside `FREE_TIER_TOOL_NAMES`, and on the authenticated ChatGPT surface — so the one test that exercised it (freetier.test.ts:900) now returns from this line instead, while its comments still read "Blocked at dispatch". The gate is worth keeping as defence in depth, but should it get a direct `createMcpServer`-level test so it is not silently unexercised?

workers/src/mcp.ts
```diff
@@ -1473,6 +1596,104 @@ function djangoErrorKind(err: DjangoError): string {
...
+ * Probing note: this gate lets an anonymous caller distinguish known
+ * from unknown tool names, which discloses nothing — the full tool list
+ * is published in `registry/server.json` and anonymous `tools/list`
+ * already works.
+ *
+ * Metering note: these calls were never per-IP metered (see
+ * `isMeteredJsonRpcBody`) and still aren't — the global ceiling already
+ * counted the request before admission returned.
+ */
+export function freeTierHiddenToolResponse(
```
robertabbott: On a client with no sign-in flow, is "Sign in with Tako to continue" actionable? The text and its `mcp/www_authenticate` challenge were built for ChatGPT's link UI; a Claude Code or Cursor user's actual remedy is putting a Tako API key in their MCP config, so the model relays an instruction the user cannot follow. `_surface.ts:157` still states the old policy for those clients ("a listed tool that errors on call wastes agent turns on hosts with no linking UI to show"), and README:296 / llms.txt:14 still describe the sign-in prompt as ChatGPT-only. Should the guidance branch on client, and the three doc sites follow?

robertabbott (commented):
One item that spans files outside the diff, so it has no line to sit on: `docs/chatgpt-app-review.md` §1 is the commerce-policy record for OpenAI's app review, and it states enforcement as `freetier.test.ts`'s blanket ban across "every message the module can emit — including any added later". This PR adds two model-visible strings outside that module — `FREE_TIER_SERVER_INSTRUCTIONS` and `PAYMENT_REQUIRED_TOPUP_SUFFIX` — reachable only by the new hand-written per-client cases. Worth recording both and their gating in §1 so the record still matches what the app can emit?

Inline findings below; the `paymentRequiredToolResult` one is the load-bearing one.

workers/src/mcp.ts
25eliu: Confirmed against the monorepo and fixed in ec1fb83 — this was the load-bearing catch. Both real emitters are now handled: `SubscriptionCreditThrottle` (monthly plan credits, regrant each cycle) and `meters_api_credits` (prepaid PAYG, where "add credits" is actually true). Rather than hand-maintaining a remedy that's wrong for one account type, `paymentRequiredToolResult` now splices Django's own `message`/`error_message` from the 402 body — so subscribers see "Upgrade your plan for more credits." and PAYG orgs see "Add credits to continue." — with a neutral fallback for unrecognized shapes. The retry line is reset-safe ("until the account has credits again"). The remedy splice is gated to commerce-allowed clients only (Django's remedies are promotional copy under OpenAI's policy); ChatGPT-family/unrecognized keep the cause-only message. Fixtures now use both real bodies instead of the invented one.

workers/src/mcp.ts
25eliu: Fixed in ec1fb83 with your suggested shape: a new `commerceCopyAllowedForUa` predicate replaces `client === "claude"` — an explicit Anthropic allowlist (the claude family plus `claude-code`, which stays "unknown" in `McpClientKind` for widget routing). The kind taxonomy is untouched (it gates five surface behaviors), and the flag threads `handleMcpRequest` → `createMcpServer` → `registerTool` as a purpose-built boolean. Still fails closed for null/unrecognized UAs. Unit-tested for both populations, plus an e2e pin that a `claude-code/2.1.220` UA gets the remedy while `python-httpx` does not.

workers/src/mcp.ts
25eliu: Added in ec1fb83: "the free-tier toolset sentence tracks FREE_TIER_TOOL_NAMES exactly" (mcp.test.ts) asserts the last paragraph names every `FREE_TIER_TOOL_NAMES` member as backticked code and no other registry tool except the `tako_contents` teaser — so adding or dropping a free tool fails the suite until the sentence follows.

workers/src/mcp.ts
25eliu: The dispatch gate does already have direct createMcpServer-level coverage that bypasses the pre-SDK gate — the in-memory `callTool` tests in mcp.test.ts ("anonymous ChatGPT call to %s returns the www_authenticate challenge WITHOUT executing", both discoverable tools, plus the ctx-tier-only fail-closed regression) — so it isn't silently unexercised. What was stale was the HTTP test's comment: fixed in ec1fb83 to say the pre-dispatch gate answers first over HTTP and to point at the in-memory tests as the dispatch gate's coverage.

workers/src/mcp.ts
25eliu: Fixed in ec1fb83. The message now names both remedies — "This tool requires a Tako account. Sign in with Tako, or connect with a Tako API key, to continue." — one string for every client rather than a per-client branch: OAuth-capable hosts (ChatGPT, claude.ai, Claude Code's /mcp flow) have a sign-in, config-file clients have the API key, and the `mcp/www_authenticate` challenge still rides `_meta` for hosts that key on it. All three doc sites updated: the `_surface.ts` policy comment now distinguishes listing (ChatGPT-only) from call-time guidance (universal), and README/llms.txt describe the account-guidance answer instead of the old ChatGPT-only sign-in prompt.

25eliu:
Addressed all 5 line comments + the review-summary item (docs/chatgpt-app-review.md §1 now records `FREE_TIER_SERVER_INSTRUCTIONS` and the payment-required strings with their per-client gating) in ec1fb83. Full suite (1159+120+68), 4-pass typecheck, and registry:check green. @claude-address

robertabbott (approved):

PR #245, opened 2026-08-18
fix(widget): reveal the ChatGPT committed iframe only when the embed has painted

25eliu:
## Problem

On the ChatGPT desktop app (and chatgpt.com), a `tako_search` chart had an ugly loading phase: the widget committed its cross-origin `tako.com/embed/...` iframe the moment the tool result arrived, immediately unhid it, and notified the host of the full chart height. The card expanded to full size seconds before the embed had painted anything — a full-height blank box in the chat.

A second route to the same symptom: a **failed** tool call. An `isError` result carries no `structuredContent`, but ChatGPT mounts the widget from static `openai/outputTemplate` registration anyway, holds it at its default mount height, and ignores shrink notifications — so the no-data watchdog's deliberately unlabelled collapse left a tall, pure-blank card next to the model's apology text.

## Fix 1 — reveal-on-load, with a veil instead of `display:none` (`212f736`)

The committed frame is now **veiled** at commit time: a `.veiled` class pins it to height 0 while keeping its layout box alive, so the embed page boots at the **real column width**. (Plain `display:none` would give the embed a 0-width viewport — mobile media queries, degenerate layout — and a `tako-embed-height` measured there would be notified and made permanent by ChatGPT's pin-the-max-height quirk.)

The frame is revealed only when one of two render-proof signals fires:

- **`load` on the frame** — via the existing committed-iframe watchdog, which was already listening. The reveal re-fits the aspect height at the settled column width, then notifies the host. Expansion and pixels arrive together.
- **an accepted `tako-embed-height` message** — the embed is demonstrably running, so the handler unveils, and now also **settles the watchdog**: previously a stalled subresource holding `load` past the 8s timeout would downgrade a chart the user was already looking at.

Failure paths are unchanged: a `frame-src` CSP violation or the watchdog timeout still downgrade to the PNG (which reveals itself on `img.load`), and a search returning zero cards still collapses. The veil itself is confined to the `window.openai` committed-iframe branch (ChatGPT + codex desktop only).

## Fix 2 — label the box a failed call leaves behind (`daf3289`)

Every server-built failure carries `_meta["tako/error"]` (Django errors, free-tier credits/capacity, and — since the review round — both free-tier rate limiters), and SDK-wrapped throws carry the envelope's `isError: true`. The widget now reads that signal from `window.openai.toolResponseMetadata` / `openai.widget` on the ChatGPT path and from `params` on the MCP-Apps `ui/notifications/tool-result` path, and collapses with a labelled empty state ("No chart: the call failed.") instead of an unlabelled blank. Hosts that honour shrink notifications still get the zero-height collapse; hosts that keep the box (ChatGPT) show a space that reads as intentional. A host exposing no failure signal degrades to exactly the old watchdog path.

**This commit does touch the MCP Apps path** (`isErrorSignal(params)` on `ui/notifications/tool-result`): on claude.ai-class hosts a failed call now labels the collapse instead of collapsing blank. Chart-bearing and zero-card deliveries there are byte-identical to before.

## Hardening from review

A 4-way pre-landing review pass caught the veil's `display:none` viewport bug, the watchdog/handshake race, a stale-height notify, and the `tako-embed-height` sender-pinning gap (`event.source === frame.contentWindow` on top of the origin gate) — all fixed in `212f736`.

robertabbott's review round, addressed in `472ecb0`:

- **Stray `about:blank` load (blocking)**: the srcless `#tako-embed` queues a `load` for its initial empty document, delivered right after the watchdog attaches on the synchronous `toolOutput` path — one event that settled the watchdog (killing the PNG downgrade for the widget's life, latent pre-veil) and revealed the veiled frame around nothing. `onLoad` now ignores a load whose `contentDocument` is readable and still `about:blank`; the real embed is cross-origin and reads null.
- **Terminal error collapse (blocking)**: an error sighting on the sniffed ChatGPT channels is now only recorded (`errorSeen`); the poll runs its full window so a late payload always wins, and the watchdog turns the sighting into the labelled collapse at exit. The MCP-Apps path keeps the immediate label — there the host has delivered THE result.
- Free-tier limiters now carry the `tako/error` discriminant; `pickErrorFromOpenAi` probes `openai.widget`; the veiled→timeout→PNG path and the mid-veil resize stand-down are now under test; three pre-veil comments swept; the unreachable `embedReportedHeight` guard kept but honestly labelled as a defensive invariant.

## Test plan

- [x] Committed frame stays veiled with the 1px mount floor notified, reveals at the fitted height on `load`; re-fits when layout settles between commit and `load` (`widget-dom.test.ts`)
- [x] `tako-embed-height` while veiled reveals AND settles the watchdog (violation-after-handshake does not downgrade); wrong-sender reports ignored
- [x] Stray `about:blank` load ignored — veil kept, watchdog still armed (a later CSP violation downgrades); genuine load after the stray still reveals
- [x] Veiled→timeout→PNG downgrade end to end, PNG self-revealing on `img.load`; resize re-fit stands down mid-veil
- [x] Failed calls: `tako/error` via `toolResponseMetadata` labels at the watchdog exit; envelope `isError` spellings covered; a payload landing after an error sighting still renders; MCP-Apps `isError` params label immediately; a renderable payload always beats a stale error signal
- [x] Free-tier limiter results pin `_meta["tako/error"]`
- [x] Full suites green: 1,159 main + 134 widget + 68 scripts; all four `tsc --noEmit` projects clean
- [ ] Live check on the ChatGPT desktop app (`tako_probe` + `enable_mcp_apps` in `~/.codex/config.toml`) before flipping prod traffic — including one forced failure to confirm `toolResponseMetadata` actually carries the error envelope

## Risk

The veil is scoped to the `window.openai` committed-iframe branch (ChatGPT + codex desktop only); if `load` never fires and no handshake arrives, behavior is identical to today's 8s timeout path. The failed-call labelling additionally touches the MCP-Apps `tool-result` delivery, where it activates only on `isError` results that previously collapsed blank — success deliveries are unchanged. The widget only ever notifies larger-after-smaller heights, working *with* ChatGPT's height pinning rather than against it. If a host exposes no failure signal under `isError`, the labelling degrades to the pre-existing silent watchdog collapse.

### DIFF
--- a/workers/src/tools/_chart_widget.ts
+++ b/workers/src/tools/_chart_widget.ts
@@ -2254,6 +2254,63 @@ const WIDGET_HTML = `<!doctype html>
     return "cards" in o || "answer" in o || "web_results" in o;
   }
 
+  // Does this object say the CALL ITSELF failed? Two spellings, both
+  // positive claims rather than inferences: the MCP envelope's
+  // \`isError: true\`, and the server's own \`_meta["tako/error"]\`
+  // discriminant (attached to every mapped upstream failure — Django
+  // errors, free-tier capacity; see \`djangoErrorToToolResult\` in mcp.ts).
+  // Either is the affirmative knowledge the no-data watchdog lacks: a
+  // failed call and a late delivery stop being indistinguishable the
+  // moment one of these is visible.
+  function isErrorSignal(o) {
+    if (!o || typeof o !== "object") return false;
+    if (o.isError === true) return true;
+    var direct = o["tako/error"];
+    if (direct && typeof direct === "object") return true;
+    var meta = o._meta;
+    if (meta && typeof meta === "object") {
+      var nested = meta["tako/error"];
+      if (nested && typeof nested === "object") return true;
+    }
+    return false;
+  }
+
+  // The failed-call signal across ChatGPT's envelope spellings — same
+  // best-first probing (and the same reason) as \`pickMetaFromOpenAi\`:
+  // \`toolResponseMetadata\` "preserves the full MCP result envelope,
+  // including hidden \`_meta\`", but the envelope's internal shape is
+  // ChatGPT's, not the spec's.
+  function trmHasError(trm) {
+    if (!trm || typeof trm !== "object") return false;
+    return (
+      isErrorSignal(trm) ||
+      isErrorSignal(trm.mcp_tool_result) ||
+      isErrorSignal(trm.call_tool_result)
+    );
+  }
+  function pickErrorFromOpenAi() {
+    var w = window;
+    if (!w || !w.openai || typeof w.openai !== "object") return false;
+    return trmHasError(w.openai.toolResponseMetadata);
+  }
+
+  // Label-and-collapse for a KNOWN-failed call. The same collapse the
+  // no-chart guard uses, with failure copy: hosts that keep the box anyway
+  // (ChatGPT pins its mount height and ignores shrinks) then show an
+  // intentional labelled state instead of a blank card that reads as a
+  // display failure. The label makes no claim about WHY — the host renders
+  // the error text itself, right next to this box.
+  //
+  // Every call site checks the payload channels FIRST: a result that can
+  // paint must never be pre-empted by leftover error metadata in a lesser
+  // channel, and within one call a chart and \`isError\` never coexist.
+  function collapseFailed() {
+    if (rendered) return;
+    empty.textContent = "No chart: the call failed.";
+    collapse(true);
+    log("failed call, collapsed widget with label");
+  }
+
   // Best-first over ordered candidates, plus the widget-state mirror as the
   // last thing tried. Truthiness is NOT the selector: ChatGPT's stripped
   // reload payload (\`{width, height}\`) is a perfectly truthy object, so a
@@ -2369,16 +2426,30 @@ const WIDGET_HTML = `<!doctype html>
   // ChatGPT populates the global at unpredictable times. Cost: one
   // property read every 250 ms.
   if (!render(pickPayload(null), pickMetaFromOpenAi())) {
-    var attempts = 0;
-    var handle = setInterval(function () {
-      attempts += 1;
-      // Re-read the meta each tick, not once before the loop: the payload and
-      // its \`_meta\` can land on different ticks, and pinning a stale
-      // \`undefined\` here would reintroduce the exact bug this fixes.
-      if (render(pickPayload(null), pickMetaFromOpenAi()) || attempts >= 40) {
-        clearInterval(handle);
-      }
-    }, 250);
+    // Payload first, error second — see \`collapseFailed\`. An \`isError\`
+    // result never grows a payload later, so acting on it synchronously
+    // spares the user the 10 s of blank box the watchdog would allow.
+    if (pickErrorFromOpenAi()) {
+      collapseFailed();
+    } else {
+      var attempts = 0;
+      var handle = setInterval(function () {
+        attempts += 1;
+        // Re-read the meta each tick, not once before the loop: the payload and
+        // its \`_meta\` can land on different ticks, and pinning a stale
+        // \`undefined\` here would reintroduce the exact bug this fixes.
+        if (render(pickPayload(null), pickMetaFromOpenAi())) {
+          clearInterval(handle);
+          return;
+        }
+        if (pickErrorFromOpenAi()) {
+          collapseFailed();
+          clearInterval(handle);
+          return;
+        }
+        if (attempts >= 40) clearInterval(handle);
+      }, 250);
+    }
   }
 
   // No-data watchdog. If nothing renderable has arrived by the time the
@@ -2433,7 +2504,17 @@ const WIDGET_HTML = `<!doctype html>
     // covers both hosts' timing.
     var eventMeta =
       globals && hasWidgetMeta(globals._meta) ? globals._meta : undefined;
-    render(pickPayload(globals), eventMeta || pickMetaFromOpenAi());
+    if (render(pickPayload(globals), eventMeta || pickMetaFromOpenAi())) return;
+    // Nothing renderable on any channel — but the event may be the delivery
+    // of a FAILED call (the host populated \`toolResponseMetadata\` and fired
+    // the event with no useful detail, the same timing PATH 3 has for
+    // success). Check the event's own globals and the live snapshot.
+    if (
+      (globals && trmHasError(globals.toolResponseMetadata)) ||
+      pickErrorFromOpenAi()
+    ) {
+      collapseFailed();
+    }
   };
   EVENT_NAMES.forEach(function (name) {
     window.addEventListener(name, handler);
@@ -2511,7 +2592,11 @@ const WIDGET_HTML = `<!doctype html>
     // but a second release must not be able to replay the same payload.
     heldToolResult = null;
     log("released held tool-result", { why: why, theme: mcpHostTheme });
-    render(held.structuredContent, held._meta);
+    if (!render(held.structuredContent, held._meta) && isErrorSignal(held)) {
+      // The held first result can be a FAILED call too — same treatment as
+      // the direct-delivery branch above.
+      collapseFailed();
+    }
   }
 
   function sendInitRequest() {
@@ -2668,7 +2753,12 @@ const WIDGET_HTML = `<!doctype html>
       // Per the MCP Apps spec §"Wire protocol — Host → View
       // notification", \`params._meta\` is part of the tool-result
       // notification.
-      render(params.structuredContent, params._meta);
+      if (!render(params.structuredContent, params._meta) && isErrorSignal(params)) {
+        // A failed call: \`isError\` results carry no \`structuredContent\`, so
+        // render() had nothing to paint — but the params themselves say why.
+        // Label now instead of leaving the box to the silent watchdog.
+        collapseFailed();
+      }
       return;
     }
     if (msg.type === "tako-embed-height" && embedOrigin && event.origin === embedOrigin) {
--- a/workers/src/tools/_chart_widget.ts
+++ b/workers/src/tools/_chart_widget.ts
@@ -683,9 +683,11 @@ const WIDGET_HTML = `<!doctype html>
      between widget mount and chart data arrival on hosts with slow
      postMessage delivery (~200 ms on claude.ai); ChatGPT's
      window.openai.toolOutput is synchronous so no visible flash.
-     The chart-rendering paths (iframe / img.load / iframe-fallback)
-     un-hide the relevant element AND notifyHeight(actual height) in
-     lockstep, so the widget grows naturally on success. -->
+     The img.load and iframe-fallback paths un-hide the relevant element
+     AND notifyHeight(actual height) in lockstep; the committed-iframe
+     path defers both to its reveal (\`revealCommittedFrame\` /
+     tako-embed-height — the frame stays veiled until the embed proves it
+     painted), so the widget still only grows when there are pixels. -->
 <div id="tako-placeholder" class="hidden">Loading chart…</div>
 <!-- Shown only by \`collapse()\`, i.e. only once we know there is no chart to
      draw. Never a loading state: on a host that honours the shrink this sits
@@ -755,9 +757,10 @@ const WIDGET_HTML = `<!doctype html>
 
   // Height for the embed iframe, derived from the CARD's real aspect ratio.
   //
-  // A cross-origin iframe never sizes itself to its content, and the Tako embed
-  // page does not report its height, so this height is the only thing deciding
-  // the box. It used to be \`structuredContent.height\` — a flat 720 — against a
+  // A cross-origin iframe never sizes itself to its content, so until the
+  // embed's own \`tako-embed-height\` report arrives (it DOES emit one now —
+  // see \`embedReportedHeight\`, which outranks this everywhere) this height
+  // is the only thing deciding the box. It used to be \`structuredContent.height\` — a flat 720 — against a
   // chat column around 700 px wide. That is a near-square frame holding a card
   // whose real aspect is 2.18:1 for a plain chart, so the card painted ~320 px
   // tall and left ~400 px of empty background beneath it. Ranked/list cards run
@@ -1473,6 +1476,28 @@ const WIDGET_HTML = `<!doctype html>
     }
     function onLoad() {
       if (settled) return;
+      // Stray-load guard. \`#tako-embed\` is parser-inserted with no \`src\`,
+      // so the browser queues a \`load\` for its initial about:blank
+      // document. On the synchronous \`toolOutput\` path — ChatGPT's primary
+      // one — render() commits the src and this listener attaches in the
+      // same parse task, so that stray event is delivered NEXT tick, long
+      // before the embed responds. Unguarded it would settle the watchdog
+      // (no CSP downgrade, no timeout, no PNG fallback for the rest of the
+      // widget's life) and reveal the veiled frame at full height around
+      // nothing — both halves of the bug this file exists to prevent.
+      //
+      // about:blank is the ONE document this frame can host that is
+      // same-origin-readable; the real embed is cross-origin, where
+      // \`contentDocument\` reads null (or throws on stricter engines). So a
+      // readable document still sitting on about:blank identifies the
+      // stray exactly, with no timing heuristics.
+      try {
+        var straydoc = frame.contentDocument;
+        if (straydoc && straydoc.location && straydoc.location.href === "about:blank") {
+          log("ignored stray about:blank load");
+          return;
+        }
+      } catch (e) { /* cross-origin read — the genuine embed load */ }
       cleanup();
       log("committed iframe loaded", { src: url });
       // Reveal-on-load: the embed has rendered, so it is finally safe to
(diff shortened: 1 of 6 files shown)


25eliu:
Added `daf3289` — the other half of the blank-widget symptom.

This PR's veiled-commit fix covers the **successful call whose embed hasn't painted yet**. The new commit covers the **failed call**: an `isError` result carries no `structuredContent`, ChatGPT mounts the widget from static `openai/outputTemplate` anyway at its default height and ignores shrink notifications, and the no-data watchdog's deliberately unlabelled 10 s collapse left a tall pure-blank card next to the model's apology text.

The widget now reads the positive failure signal — `_meta["tako/error"]` (attached to every mapped upstream failure) and the envelope's `isError: true` (SDK-wrapped throws) — via `window.openai.toolResponseMetadata` on the ChatGPT path and `params` on the MCP-Apps path, and collapses immediately with the labelled empty state ("No chart: the call failed.") instead of blank.

Checked on the combined branch: 1159 + 129 + 68 tests pass, all four typecheck passes clean. One live caveat: whether ChatGPT populates `toolResponseMetadata` for `isError` results isn't provable from the jsdom harness — if it doesn't, behaviour degrades to exactly the old watchdog path, so there's no regression either way. Worth watching one real failure after deploy.

workers/src/tools/_chart_widget.ts
```diff
@@ -1448,6 +1475,10 @@ const WIDGET_HTML = `<!doctype html>
       if (settled) return;
       cleanup();
       log("committed iframe loaded", { src: url });
+      // Reveal-on-load: the embed has rendered, so it is finally safe to
+      // give the card its height. Wired through the watchdog because it is
+      // the one place already listening for the event.
+      if (typeof opts.onLoaded === "function") opts.onLoaded();
```
robertabbott: Does the srcless `#tako-embed` fire an `about:blank` `load` into this handler before the embed ever navigates? The iframe (line 695) is parser-inserted with **no `src`**, so the browser queues a `load` for its initial empty document. On PATH 1 — ChatGPT's synchronous `toolOutput`, the primary path — `render()` runs in that same parse task: it sets `frame.src`, then `watchCommittedIframe` attaches `onLoad`, and the already-queued stray `load` fires next tick. That one event `cleanup()`s the watchdog (no CSP downgrade, no 8s timeout, so no PNG fallback for the rest of the widget's life) and calls `revealCommittedFrame()`, unveiling at full height around a blank frame — the exact bug this PR removes. `probeInteractiveIframe` guards precisely this with `probeNavigated` (lines 1305, 1350). Should this path carry the same flag? The tests dispatch `load` by hand (`widget-dom.test.ts:199`), so the real ordering is unmodelled either way.

workers/src/tools/_chart_widget.ts
```diff
@@ -2292,16 +2426,30 @@ const WIDGET_HTML = `<!doctype html>
...
-      // its \`_meta\` can land on different ticks, and pinning a stale
-      // \`undefined\` here would reintroduce the exact bug this fixes.
-      if (render(pickPayload(null), pickMetaFromOpenAi()) || attempts >= 40) {
-        clearInterval(handle);
-      }
-    }, 250);
+    // Payload first, error second — see \`collapseFailed\`. An \`isError\`
+    // result never grows a payload later, so acting on it synchronously
+    // spares the user the 10 s of blank box the watchdog would allow.
+    if (pickErrorFromOpenAi()) {
```
robertabbott: Is `collapseFailed()` at boot safe to make terminal? It sets `rendered = true`, and this branch arms no 250ms poll interval at all — every later `openai:*` event calls `render()`, which returns `true` immediately on `rendered`. So one early error read permanently loses a payload that lands afterwards. The invariant holding it up is "a chart and `isError` never coexist", but the loop in the `else` right below exists because "the payload and its `_meta` can land on different ticks", and your own note on this PR says ChatGPT's `toolResponseMetadata` behaviour under `isError` is not provable from the jsdom harness. Should the interval still run so a late payload can win, with the collapse only as the loop's exit?

workers/src/tools/_chart_widget.ts
```diff
@@ -2177,6 +2254,63 @@ const WIDGET_HTML = `<!doctype html>
     return "cards" in o || "answer" in o || "web_results" in o;
   }
 
+  // Does this object say the CALL ITSELF failed? Two spellings, both
+  // positive claims rather than inferences: the MCP envelope's
+  // \`isError: true\`, and the server's own \`_meta["tako/error"]\`
+  // discriminant (attached to every mapped upstream failure — Django
+  // errors, free-tier capacity; see \`djangoErrorToToolResult\` in mcp.ts).
```
robertabbott: This says `tako/error` is "attached to every mapped upstream failure", but the two free-tier limiter results are not — should they be? `freeTierLimitResponse` (`workers/src/freetier.ts:397`) and `freeTierGlobalLimitResponse` (`:465`) are built at the HTTP layer, bypass the MCP SDK, and return `{content, isError: true}` with no `_meta` at all — unlike `djangoErrorToToolResult`, `paymentRequiredToolResult`, `freeTierCreditsToolResult` and `authRequiredToolResult`, which all carry it. Detection for those then rests entirely on bare `isError` surviving into `toolResponseMetadata`, and two of the four envelope spellings pinned in `chatgpt-diagnostic.test.ts:211-216` are the bare `_meta` object, where `isError` has nowhere to live. Rate limiting is the most common anonymous-ChatGPT failure, so that is the one that stays blank.

workers/src/tools/_chart_widget.ts
```diff
@@ -2177,6 +2254,63 @@ const WIDGET_HTML = `<!doctype html>
...
+  // best-first probing (and the same reason) as \`pickMetaFromOpenAi\`:
+  // \`toolResponseMetadata\` "preserves the full MCP result envelope,
+  // including hidden \`_meta\`", but the envelope's internal shape is
+  // ChatGPT's, not the spec's.
+  function trmHasError(trm) {
+    if (!trm || typeof trm !== "object") return false;
+    return (
+      isErrorSignal(trm) ||
+      isErrorSignal(trm.mcp_tool_result) ||
+      isErrorSignal(trm.call_tool_result)
```
robertabbott: The comment above says "same best-first probing as `pickMetaFromOpenAi`", but this drops that function's fifth candidate — should `w.openai.widget` be probed here too? `pickMetaFromOpenAi` (2221-2227) reads `trm`, `trm._meta`, `trm.mcp_tool_result._meta`, `trm.call_tool_result._meta`, **and** `w.openai.widget._meta`. On a host that delivers the envelope under `openai.widget` — the spelling the comment at 1620 says dev-mode custom connectors actually use — a `tako/error` the meta path would have found is never looked at, and the failed call falls back to the silent 10s watchdog collapse.

workers/src/tools/_chart_widget.ts
```diff
@@ -2037,12 +2108,18 @@ const WIDGET_HTML = `<!doctype html>
     placeholder.classList.add("hidden");
     rendered = true;
     var finalHeight = committedHeight !== null ? committedHeight : h;
-    notifyHeight(finalHeight);
+    // The committed-iframe branch defers the host height notification to
+    // \`revealCommittedFrame\` — announcing the full chart height here is
+    // what used to expand the card around a still-loading embed. The
+    // no-image \`validEmbed\` fallback keeps the immediate notify: it has
+    // no watchdog wired to reveal it later.
+    if (!useIframe) notifyHeight(finalHeight);
```
robertabbott: With the tail notify gated on `!useIframe`, what does the user see for the whole veiled window? `render()` hides `#tako-placeholder` and returns, and the last height the host heard is the 1px mount floor — so on a slow embed there is no chart, no spinner, and effectively no box for up to `IFRAME_SETTLE_MS` (8s) before the PNG downgrade. Pre-PR the card was sized and blank; now it is absent. Would leaving the placeholder visible until the reveal read better, or is the empty sliver the deliberate choice?

workers/src/tools/_chart_widget.ts
```diff
@@ -1939,7 +1970,42 @@ const WIDGET_HTML = `<!doctype html>
...
+        // host a stale number.
+        if (embedReportedHeight) return;
+        // Re-fit at reveal time: the resize listener stands down while
+        // the frame is veiled, and the column may have laid out (or
+        // changed) between commit and load.
+        var next = aspectHeight(imgNaturalW, imgNaturalH);
+        if (next !== null) committedHeight = next;
+        setFrameHeight(committedHeight);
+        notifyHeight(committedHeight);
+        log("committed iframe revealed", { height: committedHeight });
```
robertabbott: Is there any way to turn the veil off, or to see it fail, without a worker redeploy? The probe next door is gated three ways (`INTERACTIVE_IFRAME_PROBE_ENABLED` at 204, `_meta.interactive_probe` shipped per response, the client re-check at 1917); the veil is unconditional. And `log()` is `console.log` only (1612), inside a cross-origin widget iframe with no collector, so `committed iframe revealed` and `settled via embed handshake` exist only in someone's devtools. With the live ChatGPT desktop check still unticked in the test plan, how would "the chart never appeared" reach you other than as a user report?

workers/src/tools/_chart_widget.ts
```diff
@@ -1939,7 +1970,42 @@ const WIDGET_HTML = `<!doctype html>
...
+        // Idempotent — the embed-height handler may have revealed first —
+        // and a downgraded (hidden) frame must never be resurrected.
+        if (frame.classList.contains("hidden")) return;
+        if (!frame.classList.contains("veiled")) return;
+        frame.classList.remove("veiled");
+        // The embed's own report is AUTHORITATIVE (see embedReportedHeight):
+        // if it arrived first, the handler already sized the frame and told
+        // the host — re-notifying \`committedHeight\` here would hand the
+        // host a stale number.
+        if (embedReportedHeight) return;
```
robertabbott: Can `embedReportedHeight` ever be true at this line? Reaching it requires `onLoad` to have run, but the only writer of that flag (2778) also calls `settleCommittedWatch()`, whose `cleanup()` removes the `load` listener — so a handshake-then-load ordering never reaches `revealCommittedFrame` at all, and load-then-handshake sets the flag strictly after this returns. If that reading is right the guard is unreachable, which would also explain why no test covers it. Is there an ordering I am missing?

workers/src/tools/_chart_widget.ts
```diff
@@ -1394,24 +1415,30 @@ const WIDGET_HTML = `<!doctype html>
...
+      log("committed iframe settled via embed handshake", { src: url });
+    };
     function downgrade(reason) {
       if (settled) return;
       cleanup();
       // Unload and hide the frame before painting, so a late-arriving embed
       // cannot appear on top of the PNG, and stop honoring its height
       // messages.
       embedOrigin = null;
+      frame.classList.remove("veiled");
```
robertabbott: Is the veiled-to-timeout-to-PNG path exercised anywhere? No test advances `IFRAME_SETTLE_MS` (8000) — every downgrade test (`widget-dom.test.ts:971,986,999,1024`) and the settle test at `:797` use `fireFrameSrcViolation`, and the two tests that do advance timers (`:434`, `:500`) are not on the committed-iframe path. So this new `remove("veiled")` before `add("hidden")`, and the PNG revealing itself on `img.load` out of a veiled frame, never run. Same shape at 2028: both resize tests fire `load` first, so deleting the `veiled` guard breaks nothing. Two tests, one commit?

workers/src/tools/_chart_widget.ts
```diff
@@ -661,6 +661,15 @@ const WIDGET_HTML = `<!doctype html>
     #tako-empty { color: #b4b8bd; }
   }
   .hidden { display: none !important; }
+  /* The committed iframe's awaiting-load state. Height 0 — NOT
```
robertabbott: Three comments still describe the pre-veil world after this PR updated three others — worth the same sweep? Line 759: "the Tako embed page does not report its height", contradicted by your new 1359/1400 "DOES emit now (measured)". Line 686: "The chart-rendering paths (iframe / img.load / iframe-fallback) un-hide the relevant element AND notifyHeight(actual height) in lockstep" — no longer true of the committed-iframe path, which is the one this PR changed. Line 2639: "The Tako web app does not emit it yet", sitting directly above the handler that now reveals the frame and disarms the watchdog off that message.

robertabbott (commented):
Two of these are worth resolving before merge (the first two inline); the rest are non-blocking. Also: the PR body describes only `212f736`. `daf3289` is described in a comment, and it changes the MCP Apps path too (`isErrorSignal(params)` on `ui/notifications/tool-result`), so "claude.ai paths are untouched" and "Risk: scoped to the `window.openai` committed-iframe branch" no longer hold for the branch as it stands. Worth folding into the body before merge.

workers/src/tools/_chart_widget.ts
25eliu: Confirmed, and it's worse than stated: the stray load also `cleanup()`s the watchdog, so the CSP/timeout→PNG downgrade was dead for the widget's life (that half was latent pre-veil). Fixed in 472ecb0 — `onLoad` now ignores a load whose `contentDocument` is readable and still `about:blank`; the genuine embed is cross-origin and reads null, so no timing heuristics and no `probeNavigated`-style flag needed (a flag can't discriminate here — the stray arrives *after* the src commit). The real ordering is now modelled: two new tests pin stray-ignored (with a later CSP violation still downgrading) and genuine-load-after-stray revealing.

workers/src/tools/_chart_widget.ts
25eliu: Agreed — took your design in 472ecb0. The sniffed-channel sighting is now only recorded (`errorSeen`); the poll always runs its full window so a late payload wins, and the no-data watchdog turns the sighting into the labelled collapse at exit. The MCP-Apps `tool-result` path keeps the immediate label deliberately: there the host has delivered THE result for the call, so nothing is left to wait for. New test pins the exact scenario you described — error read at boot, payload landing two ticks later still renders, watchdog no-ops.

workers/src/tools/_chart_widget.ts
25eliu: Good catch — fixed in 472ecb0. Both limiter results now carry `_meta: {"tako/error": {kind: "rate_limited" | "global_rate_limited"}}`, mirroring their 429 branches' `data.kind` spellings (kept distinct per the existing topology-visibility comment). The widget comment's "every mapped upstream failure" claim is now actually true, and reworded to name the limiters. Tests updated to pin the `_meta`.

workers/src/tools/_chart_widget.ts
25eliu: Fixed — `pickErrorFromOpenAi` now probes `openai.widget` through the same `trmHasError` (which covers its `isError`, `_meta["tako/error"]`, and nested envelope spellings), so the comment's "same probing as pickMetaFromOpenAi" holds. In 472ecb0.

workers/src/tools/_chart_widget.ts
25eliu: Deliberate, keeping the sliver — Eric's call after weighing it. Absent-then-appear reads as progressive rendering while the model's text streams next to it, and a "Loading chart…" box has a real cost on this host: it needs a height (~240px) notified up front, and ChatGPT's pin-the-max-height quirk means a fitted chart shorter than the spinner box inherits a permanent empty band — the artifact this whole file fights. Worst case the sliver lasts 8s and lands on the PNG, i.e. strictly no worse than pre-PR minus the blank box.

workers/src/tools/_chart_widget.ts
25eliu: Accepting the risk for this PR, with the rationale on record: the veil's failure surface is bounded by the watchdog. "Chart never appeared" requires load AND the embed-height handshake AND the 8s timeout to all fail — and the timeout path ends in the PNG downgrade, i.e. pre-PR behavior arriving 8s late, now exercised under test (472ecb0 adds the veiled→timeout→PNG test that was missing). The unticked live desktop check stays the gate before this takes prod traffic. If we want a lever later, the probe's `_meta` flag pattern extends naturally to a server-shipped veil opt-out — deliberately not adding that surface in an already review-heavy PR. On observability: agreed `log()` is devtools-only; a widget beacon is a real follow-up, not a one-liner here.

workers/src/tools/_chart_widget.ts
25eliu: Your reading is right — no ordering is missing. Handshake-first removes the load listener via `settleCommittedWatch`'s cleanup, so `revealCommittedFrame` is never reached; load-first sets the flag strictly after the reveal returns. Kept the guard anyway, relabelled honestly in 472ecb0 as a defensive, currently-unreachable invariant (the embed's report is authoritative; never notify a stale aspect estimate over it) so a future rewiring of the settle path can't silently regress the height contract. Judgment call was keep-vs-delete; keep won for the one-line price.

workers/src/tools/_chart_widget.ts
25eliu: Both added in 472ecb0 (plus two more that fell out of the stray-load fix): veiled→`IFRAME_SETTLE_MS` timeout → `veiled` removed, `hidden` added, frame unloaded, PNG painted and self-revealing on `img.load`; and the resize re-fit standing down mid-veil (column change before `load` is picked up at reveal, not fought mid-veil). Deleting the `veiled` guard or the `remove("veiled")` now fails tests.

workers/src/tools/_chart_widget.ts
25eliu: All three swept in 472ecb0: the aspect-height header now says the embed DOES report (and that the report outranks the estimate everywhere), the placeholder comment describes the committed path's deferred reveal instead of the lockstep claim, and the message-handler header describes `tako-embed-height` as the live channel it now is (resize + reveal + settle, origin- and sender-gated).

25eliu:
Addressed all 9 comments from this review round in 472ecb0 (both blocking items fixed; 3 answered as judgment calls on-thread) and folded daf3289 into the PR body with the corrected risk scope. @claude-address

robertabbott (approved):

PR #247, opened 2026-08-19
feat(mcp): free-tier upsell + recovery paths and retry semantics for model-facing errors

25eliu:
## Summary

Improvements to Tako's model-facing messages, from an audit of every string the MCP surface puts in front of a host's model. All changes are additive and ride inside the existing HTTP-200 tool-result envelopes — no HTTP status changes, no `_meta` removals, ChatGPT-anonymous surface byte-identical (test-enforced, since an unknown UA could be OpenAI's review crawler).

### 1. Commerce-gated account upsell on free-tier limit messages
The anonymous tier's rate-limit and capacity messages carry no account copy because they reach ChatGPT's model (OpenAI commerce policy). On connections positively identified as Anthropic clients (`commerceCopyAllowedForUa` — the same allowlist gating the authenticated 402 remedy), the limit error now appends: *"Connecting a Tako account (tako.com) lifts these anonymous-access limits."* — the conversion moment Exa-style keyless tiers use. Defaults false everywhere; ChatGPT and unrecognized UAs see byte-for-byte today's messages. Compliance records (chatgpt-app-review.md §1, workers/README.md, the guard-test comment) state the gate.

### 2. Timeouts carry the transient-retry sentence
`DjangoTimeoutError` has no HTTP status, so the status-gated retry sentence never fired for the likeliest transient failure (130s search/answer ceilings). A named `retryable` predicate treats timeout-or-transient-status as one rule; a tool's own `modelGuidance` still wins. (A companion change adding 500 to `RETRYABLE_STATUS` was reverted in review: that exclusion was deliberate — a Django 500 is usually a handler crash the same input reproduces — and no observed transient-500 recovery backed the reversal.)

### 3. Authenticated 401s carry a recovery path on every client
Previously only ChatGPT-family clients got the `mcp/www_authenticate` re-link challenge; a claude.ai or Claude Code user with a rotated/revoked key got a dead-end `Django returned 401…` at HTTP 200 — jargon, no next step, no host flow. Now every authenticated client gets curated text (`AUTH_INVALID_MESSAGE`: reconnect the connector / update the API key) plus the challenge, whose `error_description` names the same remedy pair. Free-tier 401s untouched: there the rejected credential is the shared free-tier key, and "sign in again" would mask an outage as a per-user failure (guard test now pins the text channel too, not just `_meta`).

### 4. Anonymous gated-tool guidance names the concrete step on Claude clients
`anonymousSignInHint` appends *"In Claude, open Settings → Connectors and connect Tako"* on claude.ai and *"run /mcp and authenticate with Tako (or re-add the server with a Tako API key)"* on Claude Code — OAuth first, since this Worker is a full authorization server. Fails closed: ChatGPT, codex, and unrecognized UAs get byte-identical text (pinned as a hardcoded literal across both gate layers, plus a table-driven unit test over every UA family including `codex-mcp-client`). The option is computed only on free-tier connections — its sole reader is the free-tier dispatch gate.

## Test plan

- [x] Upsell: absent by default from all three producers; appended in both response shapes when allowed; hygiene rules (no scheme URL, no stale host, no rate figure, never "free"); e2e claude-code UA gets it, ChatGPT/unknown byte-identical, shared-account 402 keeps the masked cause
- [x] Direct `DjangoTimeoutError` result carries "timed out" + retry sentence; 500 pinned as the non-retryable 5xx example
- [x] Claude-client authenticated 401 → `AUTH_INVALID_MESSAGE`, no "Django", `invalid_token` challenge, `tako/error.kind` preserved; ChatGPT/codex 401 tests assert the same text; free-tier 401 guard asserts `_meta` AND text
- [x] Sign-in hints: e2e claude.ai UA, Claude Code UA, ChatGPT + unknown byte-identity; unit table over null / "" / Claude-User / claude-code / codex-mcp-client / ChatGPT-User / GPTBot / OAI-SearchBot / python-httpx
- [x] Full suites green: main 1172, widget 134, scripts 68; typecheck 4/4; `registry:check` + `schemas:check` clean
- [x] Whole-branch review + one external review round (7 comments, all addressed in 8fa0dbd)


workers/src/freetier.ts
```diff
@@ -346,10 +346,28 @@ async function hitPerIpLimiter(
...
+ * commerce-free because they reach ChatGPT's model (see
+ * `FREE_TIER_LIMIT_MESSAGE`); Anthropic hosts have no such policy, and the
+ * anonymous limit is the natural moment to say an account exists — the
+ * same conversion point Exa's keyless tier uses ("add your own API key to
+ * continue"). Bare domain, no deep link: deep links rot (the `/account/`
+ * path a previous version of this copy used was already stale when it was
+ * removed — see `PAYMENT_REQUIRED_REMEDY_FALLBACK` in `mcp.ts`).
+ */
+export const FREE_TIER_COMMERCE_UPSELL =
+  "Connecting a Tako account (tako.com) lifts these anonymous-access limits.";
```
robertabbott: `FREE_TIER_COMMERCE_UPSELL` makes three written records of the commerce ban false, and none is updated here. `docs/chatgpt-app-review.md` §1 is the reviewer-facing compliance record: its table says the four anonymous messages have "no mention of accounts or upgrades", and it claims the `freetier.test.ts` guard covers "every message the module can emit — including any added later" — but this constant is deliberately outside `ALL_FREE_TIER_MESSAGES` (freetier.test.ts:35). `workers/README.md:135` repeats the claim. Should this PR update all three to state the Anthropic-only gate, so nobody later cites a compliance doc that OpenAI could disprove from a transcript?

workers/src/mcp.ts
```diff
@@ -387,6 +387,28 @@ export function commerceCopyAllowedForUa(userAgent: string | null): boolean {
...
+ * (not `McpClientKind`) is the input because Claude Code deliberately
+ * buckets as "unknown" for widget routing yet has a completely different
+ * sign-in step (config-file API key) than claude.ai (connector Connect
+ * button). No URLs, no UI deep paths beyond the settings pane name —
+ * copy rots (see PAYMENT_REQUIRED_REMEDY_FALLBACK).
+ */
+export function anonymousSignInHint(userAgent: string | null): string | undefined {
+  if (userAgent === null || userAgent === "") return undefined;
+  if (userAgent.toLowerCase().includes("claude-code")) {
+    return "In Claude Code, re-add this Tako server with a Tako API key to enable it.";
```
robertabbott: Should the Claude Code hint name the OAuth flow rather than an API key? This worker is a full authorization server — `index.ts` routes `/register`, `/authorize`, `/token` and `.well-known/oauth-authorization-server` — and Claude Code authenticates an MCP server in place, without re-adding it. `_security.ts:160-163`, ten lines above the string this appends to, classifies Claude Code as an "OAuth-capable client [that] sign[s] in" and reserves the API-key remedy for config-file clients like Cursor. Is the re-add-with-a-key step the one you meant?

workers/src/mcp.ts
```diff
@@ -1429,8 +1471,11 @@ function registerTool(
 }
 
 /** Upstream statuses that clear on their own: worth one retry of the same
- *  call, as opposed to a 4xx that needs the request changed. */
-const RETRYABLE_STATUS = new Set([408, 429, 502, 503, 504]);
+ *  call, as opposed to a 4xx that needs the request changed. Includes 500
+ *  (transient handler crash or proxy error) since 5xx bodies are stripped from
+ *  the text channel, making the 500 indistinguishable from permanent failure
+ *  without the retry sentence — a 500 clears on retry exactly like a 502. */
+const RETRYABLE_STATUS = new Set([408, 429, 500, 502, 503, 504]);
```
robertabbott: Adding 500 reverses a deliberate exclusion, and the replacement comment asserts the opposite of the one it removed without a measurement behind it. 012a6f2 introduced this set and its message ends "Non-transient 5xx stays bare": 502/503/504 are infrastructure faults, while a Django 500 is usually a handler crash the same input reproduces. The 408 precedent quoted here was backed by an observed live recovery ("a cold-path answer call 408'd, then returned the figure on the very next attempt"). Is there a comparable observation for 500?

robertabbott (commented):
Reviewed all five changes. Findings are inline; the two that would change what ships are the compliance-record drift on `freetier.ts` and the 500 addition to `RETRYABLE_STATUS`.

workers/src/mcp.test.ts
```diff
@@ -1132,6 +1161,8 @@ describe("auth challenges (ChatGPT link-account flow)", () => {
       expect(
         (result._meta?.["tako/error"] as { kind?: string } | undefined)?.kind,
       ).toBe("unauthorized");
+      const text = (result.content?.[0] as { text?: string } | undefined)?.text;
+      expect(text).toBe(AUTH_INVALID_MESSAGE);
```
robertabbott: The free-tier 401 guard just below (line 1169, "does not claim the caller's session expired") asserts only `_meta` — never `content[0].text` — so it cannot fail for the thing it is named after. The PR body cites it as the unedited guard for change 4, but the regression it guards is now a text-channel one: hoist the `content:` override at mcp.ts:1281 out of the `tier === "authenticated"` gate and this test still passes green while anonymous callers are told to sign in about the shared free-tier key. Could it assert the text too?

workers/src/mcp.ts
```diff
@@ -1554,6 +1599,25 @@ export const PAYMENT_REQUIRED_MESSAGE =
...
+ * `authRequiredToolResult` does: connector hosts re-link, config-file
+ * clients re-key. The free tier NEVER gets this message — there the
+ * rejected credential is the shared free-tier key (see the 401 mapping in
+ * `registerTool`), and "sign in again" would mask a shared-key outage as
+ * a per-user failure.
+ */
+export const AUTH_INVALID_MESSAGE =
+  "This connection's Tako authorization is no longer valid, so the call could not run. " +
+  "Calls will keep failing until it is re-authorized: sign in with Tako again " +
+  "(reconnect the Tako connector), or update the API key if this connection uses one.";
```
robertabbott: `AUTH_INVALID_MESSAGE` names both remedies, but the challenge it now ships alongside on every client still says only "Your Tako session is no longer valid. Sign in with Tako again to continue." (mcp.ts:1288) — the session-specific wording this constant deliberately avoids. That `error_description` was written when the challenge went to ChatGPT alone; it now reaches raw-API-key connections too, where "sign in" is the remedy the text channel just told them not to use. Should the two agree?

workers/src/freetier.test.ts
```diff
@@ -1098,6 +1162,83 @@ describe("free tier end-to-end (worker.fetch with stub env)", () => {
...
+    );
+    const body = (await res.json()) as {
+      result: { content: Array<{ text: string }> };
+    };
+    const text = body.result.content[0]?.text ?? "";
+    expect(text).toContain("Tako API key");
+    expect(text).toContain("Claude Code");
+  });
+
+  it("anonymous gated-tool text on ChatGPT and unknown UAs is unchanged", async () => {
```
robertabbott: `anonymousSignInHint` has no direct unit test — every branch is reached only through `worker.fetch`, and this block pins ChatGPT and `python-httpx` but not `codex-mcp-client/…`. Codex is the branch that matters: it is ChatGPT-family, so a hint leaking there is the same policy failure the fail-closed design exists to prevent, and `detectMcpClient`'s ordering is the only thing stopping it. Would a table-driven unit test over null / "" / `Claude-User` / `claude-code` / `codex-mcp-client` / `ChatGPT-User` guard it better?

workers/src/mcp.ts
```diff
@@ -1937,13 +2014,15 @@ export async function handleMcpRequest(
         }`,
       );
     }
+    const signInHint = anonymousSignInHint(userAgent);
```
robertabbott: `signInHint` is computed and threaded into `createMcpServer` on the authenticated path, but `registerTool` reads `options.signInHint` at exactly one site (mcp.ts:1225), inside the `options.tier === "free"` dispatch gate — so on an authenticated connection it is wired through and never read. Intentional headroom, or should the call be gated on tier? As it stands the next reader sees an unused option and may "fix" it by appending the hint where the free-tier gate was not what fired.

workers/src/freetier.ts
25eliu: Updated all three records in 8fa0dbd: chatgpt-app-review.md §1 now scopes the four-message claim to ChatGPT-family/unrecognized clients and documents `FREE_TIER_COMMERCE_UPSELL` as a third client-gated string (parallel to the 402-remedy bullet, with the enforcing tests named); the README free-tier section states the Anthropic-only gate; and the `ALL_FREE_TIER_MESSAGES` comment now says BASE messages and explains why the upsell is deliberately outside the list. No compliance doc claims an absolute ban anymore.

workers/src/mcp.ts
25eliu: You're right — the Worker is the authorization server and Claude Code authenticates in place. Reworded in 8fa0dbd to name OAuth first: "In Claude Code, run /mcp and authenticate with Tako (or re-add the server with a Tako API key) to enable it." Kept the key clause as the fallback since it's still a valid remedy; docblock and the new unit test updated to match.

workers/src/mcp.ts
25eliu: No comparable observation — reverted in 8fa0dbd. You're right that the exclusion was deliberate and the audit's "most common upstream fault" claim had no measurement behind it; a handler crash the same input reproduces is the median 500, and inviting a retry there loops the model against a deterministic failure. `RETRYABLE_STATUS` is back to `{408, 429, 502, 503, 504}`, the comment now records the revert (and what evidence WOULD earn 500 a slot), and the non-transient-5xx test pins 500 again. The timeout change stays — it follows the observed-408 pattern (no response ≠ deterministic failure).

workers/src/mcp.test.ts
25eliu: Good catch — the guard couldn't fail for the thing it's named after. In 8fa0dbd it now pins the text channel too: asserts the content is not `AUTH_INVALID_MESSAGE` and matches no sign-in/session/re-authorize language, so hoisting the content override out of the tier gate turns it red.

workers/src/mcp.ts
25eliu: Aligned in 8fa0dbd: the error_description now reads "Your Tako authorization is no longer valid. Sign in with Tako again, or update the API key, to continue." — same remedy pair as `AUTH_INVALID_MESSAGE`, with a comment noting it ships to raw-API-key connections where sign-in-only was the wrong remedy.

workers/src/freetier.test.ts
25eliu: Added in 8fa0dbd: a table-driven `anonymousSignInHint` describe in mcp.test.ts covering claude-code (OAuth-first hint), the three claude.ai/Desktop UA shapes, and an undefined-row for ChatGPT, openai-mcp, `codex-mcp-client/0.148.0-alpha.9`, GPTBot, OAI-SearchBot, python-httpx, "", and null — codex pinned directly so a `detectMcpClient` ordering regression fails a unit test, not just e2e coverage.

workers/src/mcp.ts
25eliu: Gated in 8fa0dbd: `handleMcpRequest` computes `signInHint` only when `tier === "free"` (with a comment naming the dead-option trap you described), and the option's docblock now states the free-tier dispatch gate is its only reader. Authenticated servers never carry it.

25eliu:
Addressed all 7 comments from this review round in 8fa0dbd (one revert, one rewording, two message alignments, two test hardenings, three doc-record updates). @claude-address

robertabbott (approved):

PR #249, opened 2026-08-19
fix(mcp): recovery paths and retry semantics for model-facing errors

25eliu:
## Summary

Fixes the four highest-severity findings from the 2026-08-19 model-facing-message audit. Every change is additive and rides inside the existing HTTP-200 tool-result envelopes — no HTTP status changes, no `_meta` removals, ChatGPT-anonymous surface byte-identical (test-enforced).

**⚠️ STACKS ON #247 — merge that first** (this branch includes its commit; after #247 merges this PR shrinks to its own four commits).

1. **HTTP 500 is now retryable** (audit F4): added to `RETRYABLE_STATUS`, so the most common upstream fault carries the "transient — retry the same call once" sentence instead of reading as permanent.
2. **Timeouts carry the transient-retry sentence** (audit F2): `DjangoTimeoutError` has no status so the old status-gated sentence never fired for the likeliest transient failure (130s search/answer ceilings). New `retryable` predicate treats timeout-or-transient-status as one rule; a tool's own `modelGuidance` still wins.
3. **Authenticated 401s get a recovery path on every client** (audit F1): previously only ChatGPT-family got the `mcp/www_authenticate` re-link challenge; claude.ai/Claude Code users with a rotated or revoked key got a dead-end `Django returned 401…`. Now every authenticated client gets curated text (`AUTH_INVALID_MESSAGE`: reconnect the connector / update the API key) plus the challenge (hosts that don't read it ignore `_meta`). Free-tier 401s untouched — there the rejected credential is the shared key, and "sign in again" would mask an outage as a per-user failure (guard test unedited).
4. **Anonymous gated-tool guidance names the concrete step on Claude clients** (audit F6): `anonymousSignInHint` appends "Settings → Connectors → connect Tako" on claude.ai and "re-add with a Tako API key" on Claude Code. Fails closed: ChatGPT, codex, and unrecognized UAs get byte-identical text to today (an unknown UA could be OpenAI's review crawler).

## Test plan

- [x] New: 500 carries retry sentence; 501 (non-transient 5xx) still doesn't
- [x] New: direct `DjangoTimeoutError` result carries "timed out" + retry sentence
- [x] Rewritten: claude-client authenticated 401 → `AUTH_INVALID_MESSAGE`, no "Django", `invalid_token` challenge, `tako/error.kind` preserved; ChatGPT/codex 401 tests extended to assert the same text
- [x] Unedited guard: free-tier 401 gets no challenge and no session language
- [x] New e2e: claude.ai UA hint, Claude Code UA hint, ChatGPT + unknown UA byte-identity (hardcoded literal, both gate layers)
- [x] Full suites green: main 1170, widget 134, scripts 68; typecheck 4/4; `registry:check` + `schemas:check` clean
- [x] Whole-branch review (per-task spec+quality reviews plus final cross-task pass): clean, no blocking findings

### DIFF
--- a/workers/src/freetier.test.ts
+++ b/workers/src/freetier.test.ts
@@ -5,6 +5,7 @@ import type { Env, RateLimit } from "./env.js";
 import {
   checkFreeTierRateLimit,
   FREE_TIER_BATCH_MESSAGE,
+  FREE_TIER_COMMERCE_UPSELL,
   FREE_TIER_CREDITS_MESSAGE,
   FREE_TIER_GLOBAL_LIMIT_MESSAGE,
   FREE_TIER_LIMIT_MESSAGE,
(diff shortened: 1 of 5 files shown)


25eliu:
Folded into #247 at the author's request — same commits now on that branch.

PR #253, opened 2026-08-20
fix(mcp): keyless connections stop advertising tako_contents they cannot reach

25eliu:
## What

Keyless (anonymous free-tier) connections were handed a false contract: search/answer cards arrived marked `exportable: true` with a per-row USD rate card, and the `tako_contents` description said to gate on exactly that flag — but the tool isn't on the anonymous surface (hidden on Claude/Cursor/etc., auth-walled on ChatGPT). An agent following the documented contract walked into a wall. Three fixes:

### 1. Tier-aware card export markings

`slimCard` now takes `connectionCanExport` (derived from `isToolOnSurface("tako_contents", …)` in `_surface.ts`, so card markings can never drift from the actual tool surface). On a connection that cannot reach `tako_contents`:

- `exportable: true` → `false` — the flag's documented meaning is "can be fetched with tako_contents", and here it cannot;
- `export_pricing` is dropped (a rate card for an unreachable tool is noise);
- a **connection-scoped** `values_hint` is stamped: `full rows require an authenticated Tako connection; headline value is in description; for specific figures call tako_answer with node_ids [...] pinned and strict:true` — the same agent-executable routing the license-gated hint uses, worded honestly (the card *is* exportable with an account);
- inline preview rows are **kept** — they're the free tier's data.

**Deliberate carve-out:** anonymous ChatGPT-family connections keep `exportable: true`. `tako_contents` is listed-but-gated there on purpose (`CHATGPT_ANONYMOUS_DISCOVERABLE_TOOL_NAMES`) — the model calling it is what fires the link-account sign-in challenge. Re-marking cards would suppress those calls and kill the funnel. Documented on `connectionCanExport`.

The text renderer follows suit: `rowsPointer` in `_render_markdown.ts` no longer prints "(full export via tako_contents)" on a card whose `exportable` is false.

### 2. Agent-executable auth-gate fallback

`authRequiredToolResult` gains a `toolName` and appends what an autonomous agent can do **right now** — sign-in and API-key are human remedies an agent can't execute mid-run:

- `tako_contents` → headline is in the card `description`; specific figures via `tako_answer` (free-tier) with the metric node pinned, `strict:true`; web text beyond the returned snippets has no anonymous fallback — use the snippets and cite the URL;
- `tako_visualize` → present the data as a table, or cite a card's `embed_url`.

This **composes with** (does not replace) the client-specific sign-in step from `623ac6a` (`anonymousSignInHint`): `base + recoveryHint + agentFallback`. The recoveryHint stays client-gated (fails closed for unrecognized UAs — could be OpenAI's review crawler); the agent fallback is not client-gated because it names only free-tier tools and carries no commerce copy. The `_meta["mcp/www_authenticate"]` challenge is untouched. Both mcp.ts gates (dispatch + pre-dispatch) pass both parts.

### 3. "Web URLs always work" removed everywhere

Reworded to "never export-gated" (keeping the useful web-fallback routing) in the `tako_contents` description, the `web_results.url` schema description ("Always fetchable via tako_contents"), and `llms-full.txt` — the description reaches anonymous ChatGPT where the whole tool is auth-gated, and the schema text reaches anonymous models on all clients.

## Review

Deep multi-pass review ran before this PR: critical-pass checklist, 4 specialist subagents (security, testing, maintainability, API-contract), and an adversarial subagent. Findings and dispositions are in the section below. One reviewer-found inconsistency (the `rowsPointer` export hint) was fixed in-commit.

## Test plan

- [x] New unit tests: `slimCard` connection-gate behavior (flag flip, pricing strip, preview retention, license-vs-tier hint wording, default passthrough) — `_search_results.test.ts`
- [x] New unit tests: `authRequiredToolResult` per-tool fallbacks, composition order with recoveryHint, challenge metadata intact — `_security.test.ts`
- [x] New unit tests: `connectionCanExport` three-way matrix incl. the ChatGPT funnel case — `_surface.test.ts`
- [x] Updated end-to-end guard: ChatGPT/unknown-UA gated-tool text carries no client sign-in step but does carry the agent fallback — `freetier.test.ts`
- [x] New renderer test: no export pointer on non-exportable cards with preview rows — `_render_markdown.test.ts`
- [x] Full suites green: 1191 + 134 + 68; typecheck (4 projects); `registry:check` ok

## Review findings and dispositions

Multi-pass review (critical-pass checklist + security, testing, maintainability, API-contract specialists + an adversarial pass; Codex cross-model pass unavailable — local CLI needs upgrade). All CRITICAL findings fixed in-commit:

- **API-contract:** `tako_search`'s own description and `llms.txt` still said web URLs "always work" — reworded (+ registry regenerated); descriptions claiming non-exportable cards "return no rows" updated for the connection-gated case, which keeps preview rows.
- **Testing:** the tier flip is now proven through both handlers (free-claude flips, free-ChatGPT keeps, authenticated untouched), the hint composition is asserted through both mcp.ts gates end-to-end, and the renderer's export pointer has both positive and negative cases.
- **Maintainability:** the auth-gate fallback now interpolates the canonical `PINNED_FROM_CARD` fragment instead of hand-restating the pin recipe outside the drift guards; the tier default lives once, inside `connectionCanExport(ctx)`.
- **Adversarial:** hint wording no longer tells anonymous agents web text has "no fallback" (snippets in `web_results` are one); `tierValuesHint` says "full rows" so it can't contradict the preview rows shipped alongside.

Dispositions without code change:
- **Headless Codex CLI** shares the `codex` UA bucket with the desktop app, so it keeps `exportable: true` without a link-account UI — same accepted cost (and same no-per-request-signal reason) as `isWidgetClient`'s codex membership, now documented on `connectionCanExport`. Softened by the auth-gate agent fallback.
- **`export_pricing` stays on anonymous ChatGPT cards** — deliberate: the funnel keeps the full exportable experience there; the field is wire data, not commerce copy, and predates this PR.
- **`connectionCanExport` naming** (true on ChatGPT-anon where export 403s): kept — the docstring, the funnel test name, and the accepted-cost comment all pin the nuance.


robertabbott (commented):
Multi-pass review: PR-claim verification, blast-radius on `slimCard`/`exportable` consumers, three specialist passes (security/gate, test coverage, contract drift), failure-mode and mental traces of the four connection shapes, plus git archaeology on `_security.ts` and `_search_results.ts`.

The security pass came back clean and I confirmed it independently: `toolName` is compared against literals only and the pre-dispatch path already constrains it to `REGISTRY_TOOL_NAMES` before the call; the `mcp/www_authenticate` challenge is byte-identical in both gates; `connectionCanExport` can only narrow (`cardExportable && canExport`), no authorization path reads the flipped flag, and `NO_OPTIONAL_TOOLS` is safe because `tako_contents` is in none of the three per-client sets. Test coverage on the new branches is genuinely thorough — the ChatGPT funnel case, the explicit-`true` vs omitted-opts distinction, and the renderer's positive/negative pointer cases all earn their place.

The inline findings are all one class: the pre-PR contract still stated on surfaces this PR did not sweep. Three of them reach an anonymous model.

Three more are outside the diff, so they could not be inline comments:

- `_search_results.ts:887` and `:894` — the zero-card `guidance` emitted by `tako_search` (a free-tier tool) still instructs "Answer from the web_results (tako_contents on the most relevant url fetches its full page text)", on connections where that tool is not listed. This is the same claim the PR softened in three other places, and it is the one that reaches a keyless model as tool output rather than as a descriptor.
- `agent.json:85` — "Only cards that returned a `content` attribute are exportable — a card without one has no exportable data". The flip falsifies the reverse direction (content present, preview rows kept, flag false). Hand-written and outside `registry:check`'s scan, so nothing will catch it.
- `commands/data.toml:23` — "Need the full text behind a web result rather than its excerpt? Call `tako_contents` on the url", with no auth condition, in the shipped Claude Code command.

Not blocking: no data-loss, security, or incident risk, and the deliberate carve-outs (ChatGPT funnel, codex UA bucket) are correctly reasoned and documented where a reader will find them.

PR #260, opened 2026-08-26
feat: path-selected MCP surfaces; tako_answer and inline rows become opt-in

jed326:
## Findings

Four independent problems drove this change.

1. **The server keyed ~60 behaviors on the client `User-Agent`** — tool list, annotations, widget vs PNG, error copy, sign-in hints, the `securitySchemes` injection — and none of the six open-source competitor servers audited (Exa, Parallel, Firecrawl, Context7, Tavily, Perplexity) varies any of these by client. The anonymous listing also differed from the authenticated one, and the docs explained two auth methods 17 times. Developer interviews (2026-08-24) called the result confusing.

2. **Pricing copy was false since 2026-08-21.** tako#29572 removed the free row allowance, so every "free 20-row preview" claim in tool descriptions, README, and `llms-full.txt` had billed every delivered row for four days. `include_contents` defaulted to `true`, so a default 10-card search paid ~$0.012 in rows on top of a ~$0.007 search fee — on the shared anonymous account too.

3. **OpenAI rejected app v1.0.0** with "annotations … explicitly set to true or false (not null) for every tool": no tool set `idempotentHint`. Parity between the submission and the live surface was already CI-enforced, so the surface was not the cause — the missing fourth hint was.

4. **Demoting `tako_answer` broke the discovery flow, and a hand grep missed it.** `tako_available_data`'s lookup path emitted `next_call: {tool: "tako_answer"}` under a descriptor that says **"run it verbatim"**, while `tako_answer` moved behind `?tools=answer`. On both surfaces the model got the SDK's bare "tool not found" at the end of the flagship discovery→retrieval path. Five more sites carried the same defect. Found in review, after a by-hand sweep had already declared the class clean.

## What changed

- **Two path-selected surfaces, zero UA reads.** `POST /mcp` is the generic surface every client shares; `POST /mcp/chatgpt` is the OAuth-only surface submitted to OpenAI (widget `_meta`, top-level `securitySchemes`, Apps-review `openWorldHint` overrides, split agent pair, `tako_visualize` default-on). `detectMcpClient`, `McpClientKind`, `isWidgetClient`, `isChatGptFamilyClient`, `annotationClientFamily`, `commerceCopyAllowedForUa`, `anonymousSignInHint`, `descriptionByClient`, and the two anonymous/widget name sets are deleted. The one remaining `user-agent` read is the observability log line. `surfaceForPath` ignores a trailing slash, matching `canonicalizeResource`.
- **Listing is auth-invariant.** Anonymous `/mcp` lists the same three default tools (`tako_search`, `tako_available_data`, `tako_contents`); only execution is gated. `FREE_TIER_TOOL_NAMES` shrinks to search + available_data; an anonymous `tako_contents` call, or `tako_search` with `include_contents: true`, returns the `auth_required` tool result with one generic sign-in sentence — and unmetered, because `isMeteredJsonRpcBody` reads each tool's own `anonymousInputRejects`.
- **`tako_answer` is opt-in** (`?tools=answer`, documented as not recommended). Every path that routed to it is rewritten, `NextCall.tool` narrows to the single literal `"tako_search"`, and no listed tool names an unregistered one.
- **`include_contents` defaults to `false`** on search and answer, matching the raw v3 API; descriptors say when to set it and that rows bill per 1k. Every free-rows / "10 requests/min" claim is gone from code, docs, **and the comments that sit above the constants they explain** — which is where the drift regenerates.
- **`idempotentHint` set explicitly on all 12 tools**; `annotations_complete.test.ts` pins all four hints as booleans on both surfaces, and the submission parity check compares all four.
- **OAuth** accepts `${origin}/mcp/chatgpt` as a token audience. All four validation points route through one `isServerResource`, so widening cannot reach one caller and miss another. Existing tokens work on both paths; no re-grant, and no backend change — Django never sees the path.
- **`createMcpServer`'s `surface` is required.** It was optional and defaulted to `"generic"`, which made `commerceCopyAllowed` fail **open** where the deleted code failed closed. The compiler enforces it now.

## Review round

Review found two more reachable bugs and four coverage gaps. Both bugs were on the anonymous path.

- **`?tools=constructor` returned a bare 500 to an unauthenticated caller.** `OPTIONAL_TOOL_ALIASES` is an object literal, so that token and `?tools=__proto__` resolved inherited `Object.prototype` values, slipped past the `undefined` check, and threw `TypeError: toolNames is not iterable` — against the parser's own docstring promise that parsing never throws. On the anonymous path `handleMcpRequest` calls the parser *before* its outer `try`, and `index.ts` has no catch, so the throw reached the runtime. `noUncheckedIndexedAccess` types the lookup as possibly-undefined, which is exactly why the existing check looked sufficient to `tsc`. Only those two tokens survive `toLowerCase()`; the guard is on the lookup, not on a list of the two names.
- **Anonymous `tako_available_data` with `include_contents: true` escaped metering.** The unmetered condition was inlined as "`include_contents` is true", which exempted that input for *every* free tool. `tako_available_data` ignores the key (its `z.object` strips it) and declares no gate, so the call got four `djangoGet` round-trips on the shared key with no per-IP hit — by adding a key the tool never reads. It now reads the verdict off `TOOL_REGISTRY`.

**The comment explaining why it had to be inlined was wrong, and worth reading before the next author trusts a comment about this module.** It named a latent cycle `_render_markdown → _search_results → _chart_widget → _render_markdown`. That chain does not exist: `_chart_widget` imports only `env`, `embed_proxy`, and `types`, and `workers/src` is a clean DAG with zero cycles even counting type-only edges. Importing the registry *does* break — but it **creates** a cycle rather than exposing one: `freetier → _registry → _chart_widget → embed_proxy → freetier`, closed because `_chart_widget` took `EMBED_PROXY_PREFIX` from `embed_proxy`, which imports `freeTierRateLimitKey`. That constant now lives in `env.ts`, which imports nothing, so the cycle is gone and the registry read works. The previous body listed this as a ticket to file; there is nothing to file.

The failure mode is why it needed a test rather than a comment: `HTTP_URL_REGEX` arrives `undefined` at `_search_results`, zod stores it unchecked, and the **first parse** throws `Cannot set properties of undefined (setting 'lastIndex')` from inside a handler. `tsc` stays clean and every schema-only test passes, so the only signal is a few handler tests failing with a zod-internal stack that names neither module.

Also corrected: `tako_answer`'s descriptor said "START HERE", "the only tool whose single response can finish the job", and that `tako_search` "does not carry values" — while README, `llms.txt` and `_optional.ts` all say *not recommended*, and `tako_search` takes the same `include_contents` flag. `lhm.plugin.json` advertised opt-in `tako_answer` against a bare `/mcp`, so LobeHub installs were promised a tool the server never registers. `smithery.yaml` advertised four opt-in tools the same way; `agent.json` led with `tako_answer`, claimed `tako_visualize` is "on by default for ChatGPT and Claude" (true only under the deleted classifier), and pointed at `https://mcp.tako.com`, which `surfaceForPath` resolves to `null`. Five sites still described the UA classifier as the commerce-copy gate, including the OpenAI-facing review doc. README sent ChatGPT connector users to `/mcp?tools=agent`, which hands them the single-call `tako_agent` that cannot reset the host's ~60s per-call timeout. `workers/README` listed `_DRF_USER` (`/api/v1/answer/`) among throttles anonymous traffic reaches.

## Five guards, because conventions already failed here twice

Each was mutation-checked by reintroducing the bug it catches.

- **`phantom_tool.test.ts`** — no tool name in a listed tool's description, field `describe()`, or schema enum may be off that tool's surface. Allowed sets derive from `isToolOnSurface` and `OPTIONAL_TOOL_ALIASES`. This found five of the six `tako_answer` sites. Now also scans `SERVER_INSTRUCTIONS` and its free-tier variant against the **cross-surface intersection** of the default listings: that string ships on every `initialize` into the host system prompt, it is not a `ToolModule` so the registry walk never saw it, and its only guard was a hand-written `not.toContain` in `mcp.test.ts` — the pattern this file's own docblock bans, now deleted. The intersection is tighter than either surface alone, so it also catches `tako_visualize`, which the hand-written check could not.
- **`module_cycles.test.ts`** (new) — `workers/src` stays acyclic on value imports; type-only edges are erased and not counted. Failure names the constant to move rather than a four-hop path.
- **`gen-registry.ts` guard 2b** — every `LOBEHUB_TOOL_ALLOWLIST` entry must be default-on for the surface `cloudEndpoint` resolves to. `registry:check` structurally could not catch the `tako_answer` drift: it diffs generated JSON against committed JSON, and both sides agreed.
- **`gen-registry.ts` guard 2c** — an enumerated listing naming an opt-in tool must also name its `?tools=` alias. **Keyed on tool names**, so it covers `smithery.yaml` fully and `agent.json` barely, since that file describes capabilities in prose. It would *not* have caught what actually went stale there. Review then found the residual in the one file whose tool list this PR trimmed: `lhm.plugin.json`'s `description` went on saying *"Answer returns grounded prose"* next to `cloudEndpoint`, and `emitLobehub` carries that field through from the committed file, so regeneration cannot correct it. Clause dropped; the constraint now sits in a comment on the `...committed` spread, because the only derivable signal in a listing is a tool *name* and "Answer" is a capability word — matching those needs the hand-written vocabulary this repo bans.
- **`row_pricing_drift.test.ts`** — scans `src/**/*.ts` including comments for copy calling a delivered row free. Now joins comment continuations before matching, via a syntax rule (a leading `*` or `//`), because the two survivors that motivated the guard were *wrapped* comments and a per-line check saw neither half. The `tako#29572` exemption now scopes to the sentence carrying the claim, not a ±2-line window that exempted any nearby claim wholesale.

## Verification

- `npm run typecheck && npm test && npm run registry:check`: **1,196 + 134 + 162 tests green**, four tsconfigs, 12 tools, no registry drift.
- Rebased onto `ba2b8ce` (release 0.22.2). One conflict, in `agent.json`: main bumped `version`, this branch changed `description` and `url`. Both kept; the version chain reads 0.22.2 across all manifests.
- `npm run smoke` against a local `wrangler dev` (staging Django behind it): both surfaces pass — anonymous `/mcp/chatgpt` 401s with a `www-authenticate` challenge; authenticated `/mcp/chatgpt` lists 4 tools with `securitySchemes` and boolean `idempotentHint`; anonymous `/mcp` lists the auth-invariant 3 and refuses `tako_contents` / `include_contents: true`.
- Grep gates: zero hits for the deleted identifiers in `workers/src` + `workers/scripts`. `ctx.client` survives in exactly one comment, explicitly labelled deleted.

## Not done here

- **claude.ai loses the interactive widget** on `/mcp` (gets the inline PNG). Accepted: restoring it is a fast-follow gated on anthropics/claude-ai-mcp#753 and #40. `bakeImage: ctx.surface !== "chatgpt"` is unreachable today and **kept deliberately** — it is exactly what that fast-follow needs, and the comments say so rather than describing the deleted client policy.
- **`tako_answer` callers get an opaque "tool not found"** rather than "reconnect with `?tools=answer`". In-band recovery would cost a body clone on every authenticated request; hosts re-list per session, so the model-facing case self-heals. Not addressed.
- **The `/mcp/chatgpt` 401 advertises a PRM whose `resource` is `${origin}/mcp`.** Only a live ChatGPT connect settles whether a strict RFC 9728 client balks; the staging check below is that test. out of scope: no path-scoped well-known documents.
- **The row-pricing guard stays scoped to `workers/src`.** Widening it to the docs produced a false positive that cannot be removed without hand-tuning a threshold: `llms.txt` has 41 characters between "rows" and "free" in correct copy where *free* modifies a **call**, and the guard's gap is 40. No distance separates that from a real violation, because what "free" modifies is not a distance. Doc coverage needs a different shape than proximity matching, so it is not attempted here.
- **Two production-impossible test fixtures remain** (`mcp.test.ts` passes `tier: "free"` to a chatgpt-only function; the chatgpt surface is always authenticated). `app_info.description` has no parity guard. `serverInstructionsForTier` keys on tier, not surface — the new guard makes that safe rather than fixing it.
- **TakoData/docs and TakoData/tako PRs** (MCP docs page, product page, monorepo `CLAUDE.md` guidance) follow this release.
- README directional language ("see below", "the badge above") is left alone: 3 instances this branch touched against ~17 pre-existing, so fixing only the new ones makes the file less consistent.

## Post-Deploy

- Run `SMOKE_BASE_URL=https://mcp.tako.com TAKO_SMOKE_API_TOKEN=... npm run smoke` and confirm the `/mcp/chatgpt` 401, authenticated listing, and anonymous refusal lines all pass. `<production-only>`
- Verify a live ChatGPT developer-mode OAuth connect against `https://mcp.staging.tako.com/mcp/chatgpt`, including the `resource=` round-trip on authorize and token. This is the check that settles the PRM `resource` question above. `<staging-only>`
- Confirm `POST https://mcp.tako.com/mcp?tools=constructor` with an `initialize` body returns a JSON-RPC response, not a 500. `<production-only>`
- Resubmit the ChatGPT app with MCP URL `https://mcp.tako.com/mcp/chatgpt`, auth OAuth-only, and the four annotation justifications per tool from `chatgpt-app-submission.json`. The file carries no URL field and no `app_info` guard — see `docs/chatgpt-app-review.md`. `<production-only>`
- File the anthropics/claude-ai-mcp#753 repro (extra content block starving the widget of tool-result events) using the generic surface's PNG path as the baseline.
- Re-run the eval querysets against the rewritten descriptions and watch for empty-result retry loops and search→contents call chains.

## Lines changed

74 files, +2,919 / −2,767. Logic (worker source, scripts, registry, submission): +1,207 / −1,331. Tests: +1,423 / −1,167. Docs: +289 / −269.

https://claude.ai/code/session_01WRMJ7oxAZu9wFbHqBa168T


robertabbott (commented):
Two distribution manifests this PR doesn't touch now advertise tools `/mcp` no longer registers. `registry/smithery.yaml` lists `tako_answer` (48), `tako_visualize` (52), `tako_agent` (54) and `get_credit_balance` (58) — all in `OPTIONAL_TOOL_ALIASES` — against `server.url: https://mcp.tako.com/mcp` (27), with an "Answer a Data Question" example. `agent.json` sells `tako_answer` as the headline capability (3) with `answer-question` as its first skill (33-34) against `"url": "https://mcp.tako.com"`, and its `visualize-data` skill (100) still says `tako_visualize` is "on by default for ChatGPT and Claude" — true only under the deleted UA classifier; it's now path-keyed to `/mcp/chatgpt`. Nothing generates or checks either file, so the drift is silent. Were they meant to be in the sweep?

Everything else inline. Checked clean: version chain agrees on 0.22.1 across all 11 manifests; `chatgpt-app-submission.json` matches `isToolOnSurface(_, "chatgpt", ∅)` exactly with all four hints; `NextCall.tool` narrowing is sound at all four producers; `surfaceForPath` is fail-closed on `//mcp`, `/`, and percent-encodings; `/mcp/chatgpt` fail-closes anonymous; the widened token audience grants nothing cross-surface.

robertabbott (approved):

PR #263, opened 2026-08-26
feat!: make ?tools= an allowlist that replaces the defaults, and delete four tools

jed326:
`?tools=` could only ever ADD to the listing. Three things were unreachable or wrong at once because of it: a caller could not drop a default tool, `/mcp/chatgpt` advertised opt-ins its host cannot request (OpenAI snapshots the listing at submission, and no ChatGPT connector URL carries a user-controlled query string), and four tools had no job left after two live spikes.

So `?tools=` is now an allowlist of tool NAMES that REPLACES the default listing, `/mcp/chatgpt` is a fixed five-tool surface, and `docs/TOOLS.md` is generated from the registry so a human reads exactly the text the model reads.

Design: `docs/superpowers/specs/2026-08-25-tool-surface-redesign-design.md` (D1, D2, D3, D5 and the rename; D4 is PR (b)).

## Breaking changes

- **`?tools=agent` now lists ONLY `tako_agent`.** Before, it listed the three defaults plus the agent. Callers who want both must name both: `?tools=search,available_data,contents,agent`. Tokens are tool names with the `tako_` prefix optional; unknown tokens are dropped, and a param naming nothing recognizable serves the defaults, so a typo still cannot break a connection.
- **`tako_graph_search`, `tako_graph_node`, `tako_agent_start` and `tako_agent_wait` are deleted.**
- `get_credit_balance` is now **`tako_credit_balance`**.
- The group aliases (`graph`, `credits`, `answer`, `visualize`, `agent` as a group) no longer exist as tokens.
- **Three legacy `?tools=` values collapse to a single tool.** `agent`, `answer` and `visualize` were group aliases *and* are valid new tokens, so a connector already on `?tools=agent` goes from *defaults + the agent pair* to `tako_agent` alone and loses `tako_search`. Measured: 1 tool served where 4–6 were before. `?tools=credits` and `?tools=graph` are not tool names, so they fall through to the defaults — which now contain what they used to enable, so those two land clean. Not gated on a pre-Release log check: these were developer-mode URLs and the risk is accepted rather than measured. `handleMcpRequest` logs `[mcp] tools param=` if anyone wants to look.
- **The ChatGPT app surface no longer serves the Answer Agent in any form.** `/mcp/chatgpt?tools=agent` was the documented path (base `README.md:311`); `?tools=` is ignored on that surface now, so the URL serves the fixed five. `/mcp?tools=search,available_data,agent` registers `tako_agent` for a developer-mode connector, but ChatGPT's ~60 s per-call ceiling cannot hold a 30–90 s run — unsupported, not a replacement. Spec §D2 considered keeping the split pair for developer mode and rejected it: the agent returns to ChatGPT as reviewed app functionality, not a hidden opt-in.

## Surfaces after this PR

| | `/mcp` default | `/mcp` opt-in | `/mcp/chatgpt` (fixed) |
|---|---|---|---|
| | `tako_search`, `tako_available_data`, `tako_contents`, `tako_graph_related`, `tako_credit_balance` | `tako_answer`, `tako_agent`, `tako_visualize` | `tako_search`, `tako_available_data`, `tako_contents`, `tako_visualize`, `tako_graph_related` |

`tako_graph_related` and `tako_credit_balance` become default because both are free or trivial and both were unreachable in practice behind a group alias no client documented.

## What the spikes found

28 live cases against production, recorded in the spec's Appendix A and B:

- `tako_available_data` is **confidently wrong in 7 of 28** cases — "US core PCE" resolves to a company named Core, "S&P 500" to a Vanguard ETF, "2028 presidential election" to HKEX ticker `2028`, "Polymarket" to `Blockratize, Inc.`. Wrong-but-confident, not empty.
- The three graph primitives are correct but unslimmed: an NVIDIA overview is **82,741 characters**, Anthropic 83,487. No agent budget absorbs that as a routine discovery step.
- Answering "what can 3 primitives find that 2 tools cannot?" produced two things: a wider candidate list, and candidate metadata (aliases, subtype, label). Both are reachable from `tako_available_data` + `tako_graph_related`, so `tako_graph_search` and `tako_graph_node` earn nothing and are deleted here. Their behaviour fixes are PR (c).

## New guards

- **`chatgpt-app-snapshot.json`** pins the sha256 of every reviewed tool's description and input schema. `assertChatgptSubmissionParity` guards the tool set and the four annotation hints; it cannot see the two fields OpenAI also froze at submission. Editing a listed description now fails `registry:check`, and accepting the change is explicit: `npm run registry:gen -- --accept-chatgpt-snapshot`. The schema digest is key-sorted, so a zod upgrade that reorders keys cannot report a description change on all five tools at once.
- **`docs/TOOLS.md`** is generated by the same `registry:gen` run and CI-checked for drift. It renders each tool's verbatim description, its parameter table, its `fixedInputs`, per-surface annotations, the published JSON Schema, and both server-instruction variants. Its banner deliberately carries **no version**: release-please does not track the file, so an embedded version would fail `registry:check` on release-please's own PR.
- **`fixedInputs` values are asserted against the body the handler actually sends**, for every row that names a wire path. The checkable rows are computed from the declaration rather than listed, and a completeness assertion stops a new row silently skipping.
- **`ci_paths.test.ts`** enforces `workers-ci.yml`'s own rule — every file a guard READS must be a trigger path — against the generator's exported read-set. The rule had failed three times while it lived in a comment.
- **`phantom_tool.test.ts`** is re-framed for allowlist semantics and now covers the server-instructions string across resolved `?tools=` sets, in both directions: no phantom name survives a partial allowlist, and no legitimate sentence is over-filtered.

## Review rounds

Three passes. The first two are described in the commit messages; this section records what the reviewers found that the code did not already say.

**Round 1 — two defects that would each have fired on a later PR.** `docs/TOOLS.md` embedded the release version, so the next `release: X.Y.Z` PR would have regenerated nothing and failed `registry:check` on the bot's own PR. And `gen-registry.ts` had begun importing its own output: a `FREE_TIER_TOOL_NAMES` import reached `freetier.ts`, which imports `TOOL_REGISTRY` from the barrel the generator writes — so deleting a tool file kills `registry:gen` with `ERR_MODULE_NOT_FOUND` before `main` runs. This PR deletes four tool files, so it is exactly the operation that trips it.

**Round 2 — 21 findings, 18 fixed.** The two that mattered: `?tools=` read only the first query param (`searchParams.get` drops repeats, so `?tools=agent&tools=visualize` resolved to `tako_agent` alone), and `tako_graph_related` shipped unslimmed.

**Round 3 — 10 findings, all addressed.** Four were guards that agreed with a comment instead of with the code:

- `docs/TOOLS.md` published four Worker-side constants under *"Fixed request inputs (the caller cannot change these)"*. Round 2 renamed the fields to fix that, and the rename did nothing — the generator emits that heading for every non-empty `fixedInputs`, so the false claim just got a longer label. `FixedInput.scope` decides the section now.
- The `2c` disclosure guard checked adequacy **file-wide**, so one correct URL vouched for every other `?tools=` string beside it. Every value is checked individually now; the six explainers write the bare token as a token, so nothing is left to exempt.
- `ci_paths.test.ts` could not guard its own input. It reads `workers-ci.yml`, which by the rule it enforces makes the workflow a guard input, and `.github/workflows` was in neither `paths:` list — so deleting a trigger path ran no `typecheck-and-test` and the test never objected. Fourth recurrence.
- `tako_available_data`'s note claimed the top 4 candidates are drilled. `RENDER_FULL_N` is 1: one drill plus three `limit=1` probes. That note and `tako_agent`'s poll row are derived from their constants now.

Three guards added in round 2 did not bite, each proven by mutation before being fixed: deleting the `resolveToolSet` argument in `mcp.ts` left all 1,199 tests green; dropping one extra sentence from partial allowlists passed all 30 phantom tests; and the `N of M` marker plus the drilled `relation` branch of both hooks — the `limit: 100` payload the slimming exists for — had no coverage at all.

`graphErrorMessage`'s `toolName` is required now; the `` `tako_graph_${op}` `` fallback synthesized `tako_graph_search`, a tool no surface registers since the file was deleted. Its 400 handler is a `switch` with a `never` default, which deletes the arm that went dead when `"node"` left `GraphOp` and fails the build if a member returns.

## Measured: `limit` is inert in overview mode

Answering a reviewer's question about the response ceiling turned up a contract detail neither the spec nor the tool description stated. Against production, `node_id` alone:

| | `limit=3` | `limit=100` |
|---|---|---|
| bytes | 82,785 | 82,785 |
| groups | 16 | 16 |
| max items per group | 10 | 10 |

Byte-identical. The server caps each group at 10 regardless of `limit`, which bites only on a drilled relation. So the ceiling **is** bounded — the unbounded dimension is group count, not items per group. The `limit` description says so now.

Real NVIDIA payload through both slimming hooks: **181,298 → 22,446 chars, 8.1x**, confirming the fixture's 8.4x against live data.

That description change moved `tako_graph_related`'s reviewed input schema, so `chatgpt-app-snapshot.json` was re-accepted deliberately. Safe: this PR added that tool to the submission and it has **never been submitted**, so OpenAI holds no copy to drift from — the correction rides along with its first submission.

## Accepted risks

- **`next_call` can name an unregistered tool** on a hand-written `?tools=available_data` URL. Both default listings carry `tako_search` and `tako_available_data`, so only deliberate narrowing reaches it, against the warning in `docs/TOOLS.md`.
- A caller who writes `?tools=agent` alone gets a description naming `tako_search` with nothing to call. That is the caller's own choice, and forcing the defaults into every allowlist would remove the point of the feature. Stated in `docs/TOOLS.md`.

## Verified

`npm run typecheck` (4 configs), `npm test` (**1,534 tests across 52 files in three configs**), `npm run registry:check`, `npm run schemas:check`, and `wrangler deploy --dry-run` for both envs all pass.

Every guard in this PR was checked by breaking it on purpose rather than trusting a green run. Round 3 alone: reverting `readToolsParam`, drifting a `fixedInputs` constant, retuning `RENDER_FULL_N` and `AGENT_POLL_BUDGET_MS`, deleting a CI trigger path, restoring `"node"` to `GraphOp`, injecting a bare `?tools=agent` beside an adequate URL, and unslimming the drilled relation — each produced the intended failure, then was reverted.

`npm run smoke` ran against a local `wrangler dev` on this branch: `tools/list` returns the 8 allowlisted tools, `/mcp/chatgpt` returns the fixed five with `securitySchemes` and `idempotentHint` on all, anonymous `/mcp/chatgpt` gets 401 with the `www-authenticate` challenge, and `tako_credit_balance` answers under its new name. Two limits: `wrangler dev` has no free-tier bindings, so the smoke's anonymous block skips — those assertions are covered in `mcp.test.ts`, which runs in CI — and `/embed-html/` fails its CDN-rewrite assert against a local target because `PUBLIC_CDN_URL` names the staging distribution.

## Not in this PR

- The `tako_search` / `tako_search_advanced` split and the D4 defect list (`sources: "tako"`, `tako_contents.url`, `tako_answer.ts`'s `?? true`, `card_json`, the `max_rows` claim) — PR (b).
- The `tako_available_data` and `tako_graph_related` behaviour fixes from Appendix B — PR (c).
- The `TakoData/docs` re-sync (PR #417 there) and the KG tickets Appendix B lists.
- `fixedInputs.note` is unguarded prose. The values are checked; the sentences beside them are not, and two of them have now been wrong.
- The backend's OpenAPI description for `graph/related`'s `limit` has the same overview-mode gap this PR fixed in the tool description. Worth correcting upstream.
- `AGENTS.md` says release-please tracks "seven tracked files"; `release-please-config.json` lists ten. Pre-existing, left alone.

Out of scope, still open: TAKO-3877 stays open.

## Lines changed

- Logic: 1,243 insertions, 890 deletions (28 files)
- Tests: 1,291 insertions, 710 deletions (23 files)
- Docs, listings and generated artifacts: 1,772 insertions, 225 deletions (14 files) — `docs/TOOLS.md` alone is 1,621 generated lines
- Total: 65 files, 4,306 insertions, 1,825 deletions

## Post-Deploy

- <production-only> Run `npm run smoke` in `workers/` against `https://mcp.tako.com` with a real `TAKO_SMOKE_API_TOKEN`, and confirm `tools/list → 8 tools` on the allowlist URL, the fixed five on `/mcp/chatgpt`, and the anonymous block's sign-in refusals for `tako_contents`, `tako_graph_related` and `tako_credit_balance`.
- <staging-only> Connect a ChatGPT developer-mode connector to `https://staging-mcp.tako.com/mcp/chatgpt` and confirm the five-tool listing and that the OAuth `resource=` round-trips.
- At ChatGPT resubmission — which waits for PRs (b) and (c) — run `npm run registry:check` and confirm it passes without `--accept-chatgpt-snapshot`. The snapshot on this branch already matches what the resubmission should carry, so a drift error at that point means a later PR changed a reviewed tool and the submission text needs re-reading before it goes out.

https://claude.ai/code/session_01QyDNDEgj1o3pRzGnGvxvKR


robertabbott (commented):
Two findings land on workflow files this PR doesn't touch, so they can't be inline.

**`.github/workflows/regenerate.yml:68`** — the OpenAPI-sync bot now regenerates a file it doesn't stage. `registry:gen` writes `docs/TOOLS.md` (`gen-registry.ts:1040`), but the commit step stages only `schemas.ts`, `_registry.ts` and `registry/server.json`. The `Verify` step runs `registry:check` against the uncommitted working tree, so it passes; the pushed commit then carries a stale `docs/TOOLS.md`. `predeploy:staging` / `predeploy:production` also run `registry:check`, so the failure surfaces as a blocked deploy rather than a red sync PR. Should `docs/TOOLS.md` join the `git add`?

**`.github/workflows/workers-ci.yml`, the `paths:` lists** — the workflow's own rule ("every file a guard READS must be a trigger path") is broken by two new inputs. `registry:check` reads `chatgpt-app-snapshot.json` (`gen-registry.ts:971`) and `docs/TOOLS.md` (`:1018`), and neither is a trigger path. A PR that edits only the snapshot digests — which is exactly how you'd silence the ChatGPT drift guard — runs no CI at all. The comment block enumerating "what reads what" wasn't updated either. Same gap the comment says already shipped a broken skill.

Checked and clean, for the record: `?tools=graph` and `?tools=credits` degrade gracefully (unknown tokens → `null` → the defaults), so only the single-tool aliases narrow; `phantom_tool.test.ts` is genuinely derived from `resolveToolSet`; `assertChatgptSnapshot` throws before any write, so the snapshot guard is not self-updating; `get_credit_balance` has no live survivors outside the migration notes.

noahjax (commented):
Reviewed the allowlist semantics, the new generator guards, and the test coverage that replaced `_optional.ts`. The bootstrap fix — moving `FREE_TIER_TOOL_NAMES` out of `freetier.ts` so `gen-registry.ts` stops transitively importing the barrel it writes — is the best thing in the diff; it would have bitten on exactly this PR, since this is the one that deletes tool files.

I verified all six `fixedInputs` declarations by hand against their body builders (`tako_answer.ts:228`, `tako_visualize.ts:448-450`, `tako_available_data.ts:253,308`, `tako_agent.ts:193`, plus `tako_contents`); every constant is accurate today. Only `tako_search` has a drift guard, so that hand-check is the whole assurance for the other five.

Three things from the design read that aren't inline comments:

**1. `?tools=agent|answer|visualize` degrade badly; `graph|credits` degrade well.** The two old group aliases that aren't tool names fall through to the defaults, and the new defaults happen to contain what they used to enable — a nice landing. The other three are simultaneously old aliases and valid new tokens, so an existing connector on `?tools=agent` goes from *defaults + agent* to *agent only* and silently loses `tako_search`. `handleMcpRequest` already logs the raw param — worth reading live values off `wrangler tail` before the Release rather than after. If nobody is on those three URLs this costs nothing.

**2. `tako_graph_related` promoted to default while unslimmed.** Confirmed the exposure: no `slimStructured` and no `renderText`, so the response is stringified into `content.text` *and* reported as `structuredContent` — the host pays for the 82,741 characters twice. `limit` bounds a drilled relation; overview mode (`node_id` alone, which the description offers) is unbounded. The only brake is the Post-Deploy sentence about release ordering. Holding the promotion for PR (c) — one line in `_surface.ts` plus a regen — would delete a cross-PR dependency that currently lives only in prose. Your call; the risk is honestly labelled and staging-only.

**3. `chatgpt-app-submission.json` now describes intent, not the submission.** `tako_graph_related` was added to it, so after this PR the file is what we intend to submit next. Only `chatgpt-app-snapshot.json`'s `note` draws that distinction, and it draws it about the snapshot. One line in the submission file saying the same thing would help whoever resubmits.

robertabbott (commented):
Re-reviewed `d284497..f46445a` only. All 12 of my previous findings are addressed, and several were fixed harder than asked — `fixed_inputs_drift.test.ts` invokes the real body builders and carries both a per-case and a registry-completeness non-empty guard, `tako_credit_balance.test.ts` drives the real `logWireGuardFailure` rethrow rather than mocking it, `slimStructured` covers all four keys of `graphRelatedOutputShape`, and `readToolsParam`'s `getAll` join is pinned through a real `URL`. The `_surface.ts:36` answer on `.claude-plugin/plugin.json` pinning a query-less URL settles that one — that constraint wasn't visible from the diff and it's the right call.

One correction to something I'd have flagged: the `phantom_tool.test.ts` "unfiltered" case does guard against over-filtering, because `resolveToolSet("generic", null)` returns a Set rather than `null`, so the filter path really runs. The narrower gap is in the comment below.

Nine follow-ups, all on the new code. Four are issues; the rest are guard-strength.

noahjax (approved):
Round 2 — approving. All six of my findings are fixed, and I checked the code rather than the replies.

- **`readToolsParam`** — extracted, `getAll("tools").join(",") || null`, both call sites, 6 tests including the repeated-param case. Capping the logged copy at 200 was the right catch: `getAll` joining every param is what turned Robert's unbounded-echo note into a ~16KB ceiling on an unauthenticated path.
- **`agent.json` / `smithery.yaml`** — all now adequate, and you found two entries I missed (`agent.json:100`, `:116`). Leaving `smithery:62` alone is correct, and the new guard agrees with that.
- **`minimumToolsFor`** — derived from each description rather than hand-listed, which is the part that makes it survive a reword.
- **`fixed_inputs_drift.test.ts`** — computing the checkable rows instead of listing them, plus the completeness assertion so a new checkable row can't silently skip, is the right shape. That it immediately turned up `tako_visualize`'s three chart-URL rows is the guard earning its keep on the first run.

**The accepted risk I raised is gone rather than deferred.** `tako_graph_related` now ships `renderText` + `slimStructured` here, not in (c). Dropping `description`/`aliases` from related items while both stay `.optional()` on `graphNodeSchema` is what lets the advertised `outputSchema` hold unchanged, and the ≥5x-cut test pins the reduction rather than trusting the measurement. The cross-PR release-ordering dependency that was my main objection no longer exists.

Two other things in this round are worth naming because neither was asked for. `serverInstructionsForTier(tier, registered)` closes a hole wider than any tool description — I verified the default-listing output stays byte-identical to the old constants, so existing integrations see nothing, and `phantom_tool.test.ts` now guards over-filtering as well as under-filtering. And `ci_paths.test.ts` converts a rule that had failed three times while it lived in a comment into a test whose read-set comes from the generator's own path constants.

One non-blocking question below.

PR #264, opened 2026-08-26
feat(resources): answer resources/list with real documentation instead of an empty array

frejazhang2025:
An advertised capability that returns nothing reads as broken, not simple.

The `initialize` handshake declares `"resources": {}` on every connection, so a client asks `resources/list` — and only the `chatgpt` surface registers the chart widget, so on the `generic` surface, which every non-ChatGPT client lands on, the answer was `[]`. An external agent-readiness audit scored that as a failure with the exact remedy *"if you don't intend to expose resources, omit the capability — the check returns na with no penalty for tool-only servers."*

Omitting was never an option here: `chatgpt` gets its widget through that same declaration. So the fix is to make the claim true.

## What registers now

Two documents, on every instance, before the tool loop:

| URI | Content |
|---|---|
| `tako://guide/tools` | The default toolset and which need each tool serves; that `include_contents: true` on `tako_search` beats a follow-up `tako_contents` call; that `tako_available_data` is free, so an exact measure name is worth one call |
| `tako://guide/coverage` | The domains the licensed graph covers, and the query shapes it does **not** serve, so the model stops asking |

**Documentation, not data.** A resource is fetched blind — the client never says what it wants — so anything whose value depends on the question belongs in a tool call, where the model states its need. Only guidance a client needs *before* its first call belongs here.

**Text is inlined, not fetched from tako.com.** A `resources/read` that depends on a second network hop turns a documentation lookup into a failure mode, and Workers bill for the subrequest.

## Removed: the hand-wired empty-list fallback

It existed for a real reason — the SDK only wires the `resources/*` request handlers on the *first* `registerResource` call, so a widget-less instance answered JSON-RPC `-32601`, and the block traded that hard error for a polite `[]`. Reasonable at the time.

`registerDocResources` now runs unconditionally, so the SDK has wired all three handlers by that point and the list it answers with is non-empty — which the empty fallback never was. Keeping the block would leave dead code implying a case that can no longer happen. Its four now-unused type imports went with it.

## The phantom-tool test earned its keep on the first run

The guide was drafted naming `tako_answer`. That tool has since moved behind `?tools=answer` and is **absent from the default surface** — `generic` and `chatgpt` both register `tako_search`, `tako_available_data` and `tako_contents`; `chatgpt` adds `tako_visualize`. A guide naming a tool the server does not register sends the model after a ghost, the same failure this repo's sibling already guards for prompt surfaces.

Two checks now, so a future failure message points at the real cause:

1. every `` `tako_*` `` in the guide is registered on the default surface
2. every one exists in `TOOL_REGISTRY` at all

The guide was rewritten around the real toolset.

## Two existing tests asserted the bug

Both expected `resources/list` to equal `[]` — they encoded the audit finding. Rewritten against `DOC_RESOURCES` so the expectation cannot drift from what registers. The UA-independence test (`a claude UA gets the same list`) additionally asserts no `ui://` resource appears, so it still proves the thing it was written to prove — that the User-Agent buys nothing extra — rather than just that a list is empty.

## What I did not do

- **No resource templates.** `resources/templates/list` answers with an empty list, which is correct: there is no parameterised resource to template. The capability is wired by the SDK either way, and the test asserts it answers rather than throwing.
- **No live data as a resource.** See the reasoning above — that is what tools are for.
- **No `llms.txt` mirror.** Serving tako.com's site index here would duplicate a document that is already one fetch away and would go stale independently.

## Self-review found the drift bug again (commit `51becfc`)

Tako's domain list existed in **three** places and the three already contradicted each other:

| Source | Said |
|---|---|
| `serverInfo.description` | finance, economics, demographics, prediction markets, "and more" |
| `SHARED_INSTRUCTION_PARAGRAPHS` | 13 domains, **including weather** |
| the coverage resource | 10 groupings, **no weather**, plus "US government spending" neither of the others had |

An agent reads the instructions and the resource in the same handshake, so one was telling it Tako has weather data and the other that it does not — the failure this resource exists to prevent. `vocabulary.ts` is now the single source: `DOMAINS` carries the name an agent matches on plus the detail only the long-form resource renders, so neither surface restates the other.

**Worth knowing:** a first draft of that list "cleaned up" the nouns, collapsing `finance` and `company KPIs` into `company financials`. An existing test caught it, because those nouns are what a model matches a question against — renaming one is a discoverability change, not an editorial one. The names are now the original vocabulary verbatim; `US government spending` is the only addition.

Five more findings in the same commit:

- **The documents are rendered per connection**, not module constants. A static document describes somebody else's connection: `tako_visualize` registers on chatgpt only, and `tako_contents` answers sign-in instructions on the free tier. The repo already varies prose by tier (`FREE_TIER_SERVER_INSTRUCTIONS`) for that reason. The chatgpt guide now documents `tako_visualize`, the generic guide does not promise it, and only the anonymous guide describes the anonymous gate.
- **The freshness sentence** had a second near-verbatim copy in the guide; now one constant.
- **A URI collision throws** instead of being silently skipped — the documents run first against a fixed unique list, so a collision means something else claimed a `tako://guide/*` URI.
- **The phantom-tool test covers parameter names**, not just tool names. The guide tells the model to set `include_contents: true`; a renamed parameter left that instruction silently wrong.
- **`DOC_MIME_TYPE` gains `charset=utf-8`**, matching every other markdown surface we serve.

## Verification

`resources/list` non-empty on both surfaces and on the anonymous free tier; the `ui://` widget still registers alongside the docs on `chatgpt`; `resources/read` returns each document and errors on an unknown URI; `resources/templates/list` answers rather than throwing.

1212 + 134 + 165 tests pass across the three suites. `npm run typecheck` clean. Prettier clean on both new files (`mcp.ts` and `index.test.ts` carry pre-existing formatting drift on `main`, so they were left as-is rather than reformatted into a large unrelated diff).

## Lines changed

- logic: 379
- tests: 294
- docs: 0


jed326:
Closing in favour of #283, which takes the opposite approach to the same audit finding: instead of filling the resources capability, it stops declaring one on the generic surface.

Two things changed my mind while rebasing this onto the tool-surface rewrite.

**The premise here does not hold.** This PR says omitting the capability was never an option because "chatgpt gets its widget through that same declaration." The SDK declares `resources` lazily and per server instance — `setResourceRequestHandlers()` at `sdk/server/mcp.js:332` is its only caller and fires on the first `registerResource` call — and `createMcpServer` builds a fresh server per request. So the two surfaces are independent: deleting the generic declaration leaves ChatGPT's widget untouched. #283 asserts that as a test pair so it stays true.

**Nothing would have read the guides.** MCP resources are application-controlled: the host decides whether they enter context. Claude Desktop and claude.ai put them in a user-facing attachment picker, Claude Code needs an explicit `@server:resource` mention, and ChatGPT consumes `ui://` widgets rather than document resources. The model never fetches them itself. The payoff you argued for — long-form guidance somewhere that does not cost every request the way `instructions` does — only lands if something reads it.

Against that, the cost was two more model-visible prose surfaces restating tool routing, which is a fifth copy after tool descriptions, `instructions`, `next_call` hints and `docs/TOOLS.md`, landing right after #263–#279 consolidated exactly that.

None of that is a knock on the diagnosis, which was right and is what #283 is built on: the handshake advertised `resources` on every connection and served nothing, and the hand-wired empty-list fallback was the thing keeping that true. The reasoning in this body about resources being documentation rather than data, and about inlining text rather than fetching it from tako.com, is the part I would want back if we ever do expose a resource here.

For the record, the guide content had also drifted past a rebase: it documented `include_contents: true` on `tako_search` (cut in the D4 split, which inverts its "don't chain `tako_contents`" advice), named `tako_answer` after the answer fold deleted it, omitted `tako_graph_related` from the default listing, and varied text by tier against the rule `instructions.ts` now states outright.

Review on #283 would be welcome — you have the most context on this capability of anyone.

PR #272, opened 2026-08-27
feat!: cut the anonymous tier to tako_search alone, and stop varying prose by tier

jed326:
## The finding

The free-tier gate keyed on tool NAME, not cost, and the costs were inverted.

Graph endpoints carry no credit throttle. What bounds them is a per-USER rate limit (`graph_explore_user_burst`, 180/minute, `app/config/settings/base.py:517` in the monorepo), and every anonymous request authenticates as the one `FREE_TIER_API_KEY` user. One anonymous `tako_available_data` call fans out into up to ~9 graph requests (one `graph/search`, a 4-page coverage drill, 3 limit=1 candidate probes; two searches plus 2 pair probes on the lookup path). So roughly **20 anonymous coverage checks a minute exhausted the graph bucket for every anonymous caller worldwide** — while `tako_graph_related`, which makes ONE graph request, was refused. Meanwhile the per-IP limiter metered `tako_available_data` as if it spent credits. Graph calls spend none.

## What this changes

**`FREE_TIER_TOOL_NAMES` = `{tako_search}`.** The anonymous tier is a taste of search: it spends credits on the one metered tool and makes zero graph requests. The listing stays auth-invariant — four tools listed, one runs; the rest answer sign-in instructions plus the `www_authenticate` challenge at dispatch. The metered set now equals the spending set, which is what `freetier.ts`'s per-IP comment always claimed.

**`initialize` instructions are tier-invariant.** `serverInstructionsForTier` and `FREE_TIER_SERVER_INSTRUCTIONS` are gone; one string for every tier, still filtered by the resolved toolset. The MCP spec makes `instructions` an optional hint the host loads once at `initialize`, and nothing guarantees a mid-conversation sign-in re-initializes the session or rebuilds the system prompt. So the anonymous variant outlived the state it described — a model holding "`tako_contents` needs a signed-in connection" kept avoiding the tool sign-in had just unlocked — and any tier-varying prefix text guaranteed a prompt-cache miss at the sign-in boundary. The sign-in signal now lives only in the dispatch-time `authRequiredToolResult`, which is true at the moment the model reads it and needs no client refresh.

**A description audit derives its exemption from the tier boundary, not from a name.** `FREE_TIER_TOOL_NAMES` is what makes a sign-in claim true or false: a tool outside the set never runs anonymously, so "requires a signed-in connection" holds whenever a host reads it, and only a tool inside the set can wrongly tell a model to skip it. `_surface.test.ts` therefore audits the free set and skips the rest — `tako_contents` needs no name literal. A companion test pins the audited set as non-empty, since an empty free set would pass the loop vacuously. The pattern itself lives in `tools/__test_helpers.ts`: three suites had hand-rolled it into two spellings, and it now also covers "log in", "requires an account", "credentials", and "API key". It stays a hand-written pattern because the property is semantic — no type encodes "this sentence describes who may call the tool" — so the reasoning rides with the constant.

**`PAYMENT_REQUIRED_MESSAGE` names a capability, not a tool.** It used to end "`tako_available_data` is free and still works". `?tools=` can leave that tool unregistered, so the message could point a model at something the connection cannot call — and after this cut, "free" reads as credit-free in one clause and anonymous-executable in the next. Its test now asserts the message contains no registry tool name at all.

**Docs that a model reads.** `llms.txt`, `llms-full.txt`, `docs/gemini/GEMINI.md`, and `AGENTS.md` each still said `tako_available_data` runs anonymously. `AGENTS.md` is the doc the surface-split design names definitive, so the definitive doc contradicted the code.

**`smoke.ts`** drops five stale `tako_credit_balance` references (tool deleted in `e384125`; the staging smoke has been red on `requiredTools` since), and gains the anonymous `tako_available_data` sign-in assertion.

**A pinned ChatGPT description did change, but this is still not a resubmission trigger.** Review found that `tako_search`'s description told the model to run `tako_available_data` "(free)" — the one description an anonymous caller is guaranteed to read, naming the tool this PR just removed from their tier. Fixing it trips the `chatgpt-app-snapshot` guard, since `tako_search` is a submitted tool; the snapshot is accepted in `22fce51`. Do not resubmit for it: `/mcp/chatgpt` is OAuth-only, so every connection there is authenticated and "(free)" was already true in the credit sense. The edit keeps one description true across all surfaces and changes nothing for ChatGPT users. No schema changes.

## What I did not do

- No per-call graph budget (`anonymousInputRejects`) for `tako_available_data` — moot now that no anonymous call reaches graph; the hook stays dormant for the next priced input.
- Plugin artifacts keep prescribing `tako_available_data`: `commands/coverage.toml` and the three bundled skills. Their audience is signed-in, where "free" is accurate in the credit sense, and an anonymous `/coverage` returning a sign-in refusal is the intended prompt. Rewording 12 sites (each mirrored into README's embedded copies for `skills.test.ts` parity) buys no behavior change.
- `llms.txt` and `AGENTS.md` still list `tako_credit_balance` among the default tools. That staleness predates this branch (the tool died in `e384125`) and is untouched here.
- The `spec D1` / `D4` / `D6` labels in `mcp.ts`, `mcp.test.ts`, and `freetier.test.ts` cite documents under the gitignored `docs/superpowers/`, so they resolve for nobody but their author. Each comment states its constraint in full, so the labels are provenance only; `freetier.ts` no longer cites a path there, but the labels stay.
- `docs/superpowers/specs/2026-08-25-*` still describe the two-tool anonymous set. Those files are untracked local notes, not repo content.

## Review round (22fce51)

Review caught that the set cut removed `tako_available_data` from the anonymous tier and left every sentence GRANTING access to it. The audit above could not see any of them: it matched only availability-*denying* vocabulary, never granting. `free` reads as "you can call this".

- **`TIER_CLAIM` now covers the granting half** (`\bfree\b`, `no charge`, `at no cost`, `unlocks?`). Verified load-bearing: it matches all three strings below and clears all three rewrites.
- **`tako_search`'s description** said run `tako_available_data` (free) — the one description an anonymous caller is guaranteed to read, and inside the exact set the audit loops over.
- **The `initialize` sentence** said `tako_available_data` is free. An anonymous `?tools=available_data` connection registers one tool, ships that one sentence, and refuses every call it invites.
- **`buildZeroResultGuidance` now branches on tier.** All four branches opened on `tako_available_data`, three also routing to `tako_contents`, inside a `tako_search` result — so an anonymous zero-result caller got numbered steps in which every step answers a refusal. A tool RESULT may vary by tier where a description may not: it lands after the host's cached prefix and is true when the model reads it. This widens that rule from the refusal itself (`authRequiredToolResult`) to a successful result. The `tier` parameter is required, not defaulted, so the compiler found all 14 call sites. (The `tako_contents` half was already broken before this PR — that tool has never been free-tier — so it is a pre-existing fix carried along, not a regression here.)
- **`PAYMENT_REQUIRED_MESSAGE` drops its surviving-move clause.** Two versions shipped, naming `tako_available_data` then "graph exploration"; both are unreachable on `?tools=search`. The hazard is reachability, so no spelling is safe unfiltered.
- Restored `tako_available_data` to the authenticated `securitySchemes` loop (it covers the one tool whose membership flipped); corrected the `freetier.ts` claim that Django's per-user throttles bound every anonymous request (a refused call reaches neither Django nor the per-IP bucket, leaving only the per-colo limit); and corrected two test comments to claim only what they prove.

## Verification

- `npm run typecheck`: clean across all four tsconfigs.
- `npm test`: 42 + 3 + 8 files, **1646 tests**, green.
- `npm run registry:check`, `npm run schemas:check`: clean.
- `docs/TOOLS.md` regenerated (`registry:gen`): `tako_available_data` now renders "Runs anonymously: no", one instructions block instead of two.

Lines changed: 515 insertions / 364 deletions across 28 files.


workers/src/tools/_surface.ts
```diff
@@ -36,14 +36,29 @@ export const GENERIC_DEFAULT_TOOL_NAMES: ReadonlySet<string> = new Set([
...
+ * limit in Django (`graph_explore_user_burst`, 180/minute across EVERY
+ * anonymous caller worldwide, since they all authenticate as one key). One
+ * `tako_available_data` call fans out into up to ~9 graph requests
+ * (`graph/search` + a 4-page coverage drill + candidate probes), so roughly
+ * twenty anonymous calls a minute exhausted the bucket for everyone — while
+ * `tako_graph_related`, which makes ONE graph request, was refused. With
+ * this set at `tako_search` alone, anonymous traffic makes zero graph
+ * requests and the tier's only cost is the credit-priced search it meters.
+ * The anonymous tier is a taste of search that converts to sign-in, not a
+ * discovery surface: a refused call returns the sign-in result, which is the
```
robertabbott: The PR body says the zero-result guidance "already exists tier-neutrally in `_search_results.ts`, so no copy change there" — but all four of its branches name tools an anonymous caller can no longer run. `_search_results.ts:961` opens the all-empty recovery with "call tako_available_data (free) with the entity to learn the exact metric NAMES"; `:936`, `:942` and `:955` do the same, and three of them also route to `tako_contents`. That text is emitted by `tako_search` — now the only anonymous tool — on its failure path, so an anonymous caller whose search returns nothing gets a recovery protocol in which every step refuses. A tool *result* is dispatch-time, which is where this PR argues the tier signal belongs. Should `buildZeroResultGuidance` branch on tier?

workers/src/instructions.ts
```diff
@@ -153,7 +170,7 @@ const SHARED_INSTRUCTION_PARAGRAPHS = [
  */
 type ToolSentence = { readonly tools: readonly string[]; readonly text: string };
 
-const AUTHENTICATED_TOOL_SENTENCES: readonly ToolSentence[] = [
+const TOOL_SENTENCES: readonly ToolSentence[] = [
```
robertabbott: `TOOL_SENTENCES` still serves "`tako_available_data` is free, and answers what data Tako has…" (line 180) to every anonymous connection, where that tool now always answers `authRequiredToolResult`. The module comment added just above forbids exactly this — text "true on one tier and false on the other". "Free" was written when it meant credit-free *and* anonymous-runnable; this PR split those, and `mcp.ts:1241` was reworded for that very ambiguity while this line was not. Sharpest case: an anonymous `?tools=available_data` connection registers one tool, ships this sentence, and refuses every call. Should the sentence say credit-free, or drop?

workers/src/tools/__test_helpers.ts
```diff
@@ -85,3 +85,21 @@ export function requestFrom(
...
+ * is the tier boundary, so only a tool inside it can wrongly tell a model it
+ * needs sign-in. Callers pair this pattern with that set rather than exempting
+ * tools by name.
+ *
+ * It lives here because three suites audit for the same thing
+ * (`_surface.test.ts`, `instructions.test.ts`, `mcp.test.ts`) and three
+ * hand-rolled copies drifted into two different spellings within one PR.
+ */
+export const TIER_CLAIM =
+  /anonymous|sign(ed|ing)?[ -]in|log(ged|ging)? ?in|Tako account|requires? (an )?account|credentials|API key/i;
```
robertabbott: `TIER_CLAIM` matches only availability-*denying* vocabulary (sign-in, Tako account, credentials), never availability-*granting* vocabulary — "free", "no charge", "unlocks". That's the half the audit needs: it loops over `FREE_TIER_TOOL_NAMES`, and `tako_search.ts:55` — the single tool in that set — tells the model to "run `tako_available_data` (free) instead of guessing here", now a guaranteed refusal anonymously. The new `_surface.test.ts` audit passes green over it, and `instructions.test.ts:17` certifies the same word clean in `SERVER_INSTRUCTIONS`. Should the pattern cover "free" and its synonyms?

workers/src/tools/_surface.test.ts
```diff
@@ -90,3 +104,38 @@ describe("toolAnnotationsForSurface", () => {
     expect(toolAnnotationsForSurface(tool, "chatgpt").openWorldHint).toBe(false);
   });
 });
+
+describe("tool descriptions are tier-invariant", () => {
```
robertabbott: Would this audit fail if the PR were reverted? `SHARED_INSTRUCTION_PARAGRAPHS`, `TOOL_SENTENCES` and every tool description are byte-identical to `5d2f532` — the tier claims lived in the deleted `FREE_TIER_TOOL_SENTENCES`, which is gone rather than asserted against. So the block that headlines the PR is a pure forward guard over unchanged strings; the assertions that would catch a revert are elsewhere (`FREE_TIER_TOOL_NAMES` equality, `isMeteredJsonRpcBody`, the `tako_available_data` auth-challenge case). Worth saying so in the comment, so a later reader doesn't read green here as evidence the cut held?

README.md
```diff
@@ -63,11 +63,11 @@ claude plugin marketplace add TakoData/tako-mcp
 claude plugin install tako@tako
 ```
 
-That's it — `tako_search` and `tako_available_data` work right away on the anonymous free tier. To unlock the full toolset and your own account limits, authenticate once with OAuth: run `/mcp` inside Claude Code, select **tako**, and choose **Authenticate**. A browser opens to sign you in with your Tako account and a per-host API key is minted for you automatically (visible and revocable at [tako.com/console/api-keys](https://tako.com/console/api-keys)). The same OAuth flow powers the plugin on Claude.ai — the plugin's Tako connector connects with a click, no token pasting.
+That's it — `tako_search` works right away on the anonymous free tier. To unlock the full toolset and your own account limits, authenticate once with OAuth: run `/mcp` inside Claude Code, select **tako**, and choose **Authenticate**. A browser opens to sign you in with your Tako account and a per-host API key is minted for you automatically (visible and revocable at [tako.com/console/api-keys](https://tako.com/console/api-keys)). The same OAuth flow powers the plugin on Claude.ai — the plugin's Tako connector connects with a click, no token pasting.
```
robertabbott: This line sells the plugin as working "right away on the anonymous free tier", but the same plugin bundles `commands/coverage.toml`, whose entire prompt is "Call `tako_available_data`. It is free and fast" — one call, now always refused anonymously. The three bundled skills prescribe the same tool as the FREE coverage check and as the mandatory recovery step between two priced searches (`skills/tako-financial-research/SKILL.md:24,32,33,62` and the two siblings). Should the plugin artifacts carry an anonymous fallback, or is the sign-in refusal the intended prompt there too?

workers/src/mcp.ts
```diff
@@ -1233,7 +1238,7 @@ export function djangoErrorToToolResult(err: DjangoError): {
 export const PAYMENT_REQUIRED_MESSAGE =
   "This Tako account is out of credits, so the call could not run. " +
   "Priced calls will keep failing until the account has credits again; " +
-  "`tako_available_data` is free and still works.";
+  "calls that spend no credits, such as graph exploration, still work.";
```
robertabbott: "calls that spend no credits, such as graph exploration, still work" is unreachable on an authenticated `?tools=search` connection, which registers `tako_search` alone. The comment above argues a capability is safer than a tool name because `?tools=` can leave a tool unregistered — but the hazard was reachability, not the spelling, and the model's only way to act on "graph exploration" is a tool it does not have. The new assertion enforces only the absence of registry names, so it can't see this. Should the clause be gated on the resolved toolset the way the instructions sentences are, or dropped?

workers/src/freetier.ts
```diff
@@ -23,11 +23,11 @@
  *   `app/backend/api/throttling/policy.py` in the monorepo rather than from
  *   here: a copy of that table went stale the moment `tako_answer` moved
  *   behind `?tools=answer` (it named answer, which no anonymous connection
- *   can reach, and omitted `tako_available_data`, which every one can).
+ *   can reach), and again when `tako_available_data` left the free set.
```
robertabbott: This paragraph still says the genuinely global ceiling is "Django's Redis-backed per-user throttles … which every anonymous request lands on" (lines 19-22), now false for the class of request this PR de-metered. An anonymous `tools/call tako_available_data` is answered at `mcp.ts:1444` or `mcp.ts:828` and never reaches Django, while `isMeteredJsonRpcBody` skips the per-IP bucket — leaving only the per-colo bucket the same paragraph calls "not a true global ceiling". Should the header say a refused tool call has no global bound, and is the per-colo limit sized for that?

workers/src/instructions.test.ts
```diff
@@ -1,18 +1,19 @@
...
+import { TIER_CLAIM } from "./tools/__test_helpers.js";
 
 describe("instructions module", () => {
-  it("serves the authenticated text for authenticated and the anonymous text for free", () => {
-    expect(serverInstructionsForTier("authenticated")).toBe(SERVER_INSTRUCTIONS);
-    expect(serverInstructionsForTier("free")).toBe(FREE_TIER_SERVER_INSTRUCTIONS);
+  it("serves one string with no tier parameter", () => {
+    // The host loads `instructions` once at `initialize`; a per-tier variant
+    // outlives a mid-conversation sign-in. The signature is the guard.
+    expect(serverInstructionsFor.length).toBe(0);
```
robertabbott: `expect(serverInstructionsFor.length).toBe(0)` can't hold the invariant the comment claims ("the signature is the guard"). `Function.length` stops counting at the first defaulted parameter, so a reintroduced `serverInstructionsFor(registered = null, tier: Tier = "authenticated")` still reports 0 — as does an options-object signature, or a tier read off module state. It catches only a *required* leading parameter. Since `mcp.ts:236` is the sole caller, would asserting the call site's argument list hold more than an arity number?

workers/src/tools/_security.test.ts
```diff
@@ -32,7 +32,7 @@ describe("securitySchemesForTool", () => {
   });
 
   it("never advertises noauth on AUTHENTICATED connections (the caller is already linked)", () => {
-    for (const name of ["tako_search", "tako_answer", "tako_available_data"]) {
+    for (const name of ["tako_search", "tako_answer"]) {
```
robertabbott: Dropping `tako_available_data` from this loop removes coverage on the one tool whose free-set membership actually flipped in this PR. `securitySchemesForTool` returns `[oauth2]` for every name when `tier !== "free"`, so the assertion was still true and still meaningful — and a future "advertise noauth for tools that used to be free" regression is exactly what it would catch. Is there a reason beyond tidying?

robertabbott (commented):
Three of these are model-facing correctness on the anonymous surface: after the cut, the copy an anonymous connection reads still routes it to `tako_available_data` as "free" — in the `initialize` instructions, in `tako_search`'s own description, and in `tako_search`'s zero-result recovery guidance. The new `TIER_CLAIM` audit can't see any of them because it matches only availability-denying words. The rest are suggestions.

Verified clean along the way: the dispatch gate is genuinely double-layered and no `?tools=` value lets a non-free tool reach Django, and `docs/TOOLS.md` matches what `gen-registry.ts` now emits.

workers/src/tools/__test_helpers.ts
jed326: Right, and this was the root cause of the other two. I audited for prose that DENIES availability and never for prose that GRANTS it.

`TIER_CLAIM` now covers the granting half (`\bfree\b`, `no charge`, `at no cost`, `unlocks?`). I checked it is load-bearing rather than decorative: it matches all three original strings and clears all three rewrites, and `tako_contents`'s legitimate sentence still matches (it is exempt structurally, not by pattern).

workers/src/instructions.ts
jed326: Fixed. Dropped the claim rather than rewording it to "spends no credits" — the credit axis is not what a model needs from that sentence; what it needs is what the tool does. It now reads "`tako_available_data` answers what data Tako has on an entity or a metric, including a measure's exact name."

Your `?tools=available_data` case is the one that made the call easy: one registered tool, one sentence, every call refused.

workers/src/tools/_surface.ts
jed326: You are right and the PR body was wrong — I checked that text for denying claims and never for granting ones.

`buildZeroResultGuidance` now takes a required `tier` and branches. Your framing decided it: a tool result is appended AFTER the host's cached prefix and is true when the model reads it, so branching costs no prompt cache. Worth naming that this widens the rule from the refusal itself (`authRequiredToolResult`) to a SUCCESSFUL result, which is new territory — I made the parameter required rather than defaulted so the compiler found all 14 call sites instead of one silently defaulting.

The anonymous all-empty branch opens on the shape rules (the one lever that works) instead of on the coverage check. New tests cover all six source/web/pin shapes on both tiers, and assert the authenticated protocol is unchanged — a blanket strip would have cost paying callers their recovery path.

One correction to the bundling: the `tako_contents` half was already broken for anonymous callers before this PR, since that tool has never been in `FREE_TIER_TOOL_NAMES`. I fixed it in the same branch because it was free to do there, but it is a pre-existing bug rather than a regression this change caused.

workers/src/mcp.ts
jed326: Dropped the clause entirely. You are right that the hazard was reachability and not the spelling — I had swapped a tool name for a capability and called it fixed, which was treating the symptom.

Gating it on the resolved toolset is the only way to make it reliably true, and that is not worth the plumbing for one optional sentence, so the message is now cause plus retry semantics only. The test asserts the absence of both prior phrasings.

workers/src/freetier.ts
jed326: Fixed. The paragraph now scopes the Django claim to requests that REACH Django, then names the gap explicitly: a `tools/call` outside `FREE_TIER_TOOL_NAMES` is answered at the dispatch gate without touching Django, and `isMeteredJsonRpcBody` skips the per-IP bucket because it spends nothing, so the per-colo bucket is its only bound.

On your second question — I have not resized anything. The refusals are cheap (no upstream call, no credit), so I recorded it as a sizing note at the constant rather than treating it as an alarm. Say the word if you want the per-colo number revisited before merge; that class did grow to every default tool but `tako_search`.

workers/src/instructions.test.ts
jed326: Correct, and I have stopped claiming otherwise rather than pretending the number holds the invariant. `Function.length` stops at the first defaulted parameter, so it catches only a required leading one.

Asserting the call site's argument list from a unit test is not really reachable, so the comment now says what actually holds this: the single call site at `mcp.ts:236` passing only a resolved toolset, plus the two same-instructions-per-tier assertions that compare served STRINGS rather than an arity. The check stays as a cheap tripwire, labelled as one.

workers/src/tools/_surface.test.ts
jed326: Fair, and the comment now says so — read green there as "no free tool oversells itself", not as evidence the tier boundary held. It names the three assertions that do fail on a revert of the cut.

One update since you wrote this: the premise is now partly overtaken. Fixing the `free` claims changed `tako_search`'s description and the instructions sentence, so the audited strings are no longer byte-identical to 5d2f532. It still would not catch reverting the SET, which is your real point.

workers/src/tools/_security.test.ts
jed326: No reason beyond tidying, and you are right that it was the wrong thing to tidy. Restored, with a comment saying why that name specifically earns its place: its free-set membership is the one that flipped, so "advertise noauth for tools that used to be free" is exactly the future regression this catches.

README.md
jed326: Deliberately leaving the plugin artifacts alone, and the sign-in refusal is the intended prompt there.

Reasoning: the plugin's audience is overwhelmingly signed-in, and for them `free` is accurate in the credit sense — `coverage.toml`'s description already says the precise thing ("free, no credits spent"). Anonymous `/coverage` returning a sign-in refusal is a coherent answer rather than a broken one, and we are not optimizing that path. The alternative is rewording 12 sites across three skills and two commands, each mirrored into README's embedded copies for `skills.test.ts` parity, for zero behavior change.

Happy to be overruled if you read the anonymous plugin experience as more load-bearing than I do.

robertabbott (approved):

